PR #26, opened 2015-01-30
added a xml doc for routing diagnostics for easy usage of debugging

alan-stokes:
adds a provance report (currently only has routing diagnostics). The report is in XML so that it can be used in big data stuff more easily. (theres a possible debate of putting it as a ontology, but i cant be bothered to do that jsut yet)


alan-stokes:
grand, deleting branch

PR #71, opened 2015-05-12
Jamie_capocaccia2015

neworderofjamie:
As well as the spynnaker-side of the application name hashing, this fixes/improves a couple of things:

* Details of what is stuck in the beginning of the DMA buffer does not belong in the synapse row implementation
* Removed application magic numbers and replaced with md5 hash - seems to work, although some trickery might be required for mac
* Intrinsic plasticity implementations require time
* Fixing of bugs relating to delta synapses, but actually removing them to extra models
* Seperated the concept of leaky integrate neurons from those that integrate and fire


alan-stokes:
Hi Jamie. 

So having completed my own implemtnation fot his, id thought id try merging your stuff into my brnach as a way of doing the pull request. As your stuff (we think) may be complemtntory to my own changes.

Anyhow. As i find quanderies, ill post as comments (so apoplogies if theres a lot of comments in the end). 

1. can you explain why youve split the leaky_Intergrate_and_fire component into 2 classes ( leaky_Intergrate_and_fire and leaky_intergrate), I cant see the reasoning for this split, as after i pulled all your pull requests, it still seems that this devision is only between leaky_Intergrate_and_fire and leaky_intergrate. If this is the case, i cant see the reason to complicate the matters more by the speration, than by keeping them as one entity.

neworderofjamie:
Well, lots of neuron types are based around leaky integrators, but not all
of them simply fire a spike after a fixed threshold. I was working on some
weird neuron model (which I subsequently abandoned) that sends payload
packets and there are other neuron models that leaky integrate and burst
and, those like Evie's where the threshold is not fixed.

alan-stokes:
to point 1. fair enough. would like to see the evie's model updated to reflect this, otherewise it does seem to hang on its own. 

The note i've figured is that your application name hash is on a entire model perspective. which means i could say im using components x,y,z in my name and not bother doing so within the code. Id be more cofttable using the md5 thing to work on components and use that instead of a arbitary magic number. for clarity of my explination. please refer to pull request #72 (though this uses magic numebrs still)

alan-stokes:
2. the abstract_delta_population_vertex looks like a fully fledged neural model. Its internal name is IFCurrentDeltaPopulation. This would imply that its not abstract, and strangely enough, it inhirrits from itself. see line 7 and 18 in abstract_delta_population_vertex. 

having done a significant rewrite with components instead of vertices and such, im inclined to blieve that the name would be IFCurrentDeltaPopulation which inhirrts from:

 DeltaComponent, IntegrateAndFireComponent, CurrentComponent, and AbstractPopulationVertex

please advise if this is what you were intending.

alan-stokes:
to add to point2: it seems you have a if_curr_delta as well (which doesnt use the abstract_delta_population_vertex at all). would i be correct in stating the abstract_delta_population_vertex is some rediment of a bodge thats been dropped and not cleaned?

alan-stokes:
point 3: the python docs on lines 1,2,3,24,25,26 of the if_curr_delta seem to discuss a if_curr_exp. i persume a copy and paste error.

neworderofjamie:
Fixed the doc string!

neworderofjamie:
Whoops - I got confused at some point and forgot to commit that - If you
pull my branch of extra models now, it should make more sense..

alan-stokes:
point 4: methods on lines 121 and 126 have no python docs for the if curr delta.
point 5: get_n_synapse_type_bits  method for the if_curr delta on line 132 does not match the spec in the synmatic manager on line 282. I note this is more a error of the synaptic manager's definition as a abstract static, as it currently is only defined as abstract. so i recommend going to line 282 and adding the tags in this order (otherwise pycharm says errors)

    @staticmethod
    @abstractmethod

alan-stokes:
points 2 and 3: k, will do.

alan-stokes:
point 6: i dont agree with the yuck comment in your abstract_delta_population_vertex. If the delta has no synapse params, then it would be just as good to pass the function. I also note that theres no python docs for these function either. 

Due to the differences in distinctions in our brnaches, these fucntions have moved to the dleta_component in models/components/shaping_components/delta_component in my brnach, but the aim is equiv.

neworderofjamie:
Well, the yuck (if it's the one I'm thinking) is because that that function
is actually required to switch regions as well as write stuff in it :)

alan-stokes:
too say, we could push that call inside the abstract dataspecable population vertex functionality. please place a issue for it, and once these merges are done, we can tackle it.

neworderofjamie:
Raised!

alan-stokes:
point 7: just more a comment on style than anything (to reduce workload for the pull requests in future) we prefer to use  AbstractModelComponent.__init__(self) instead of super(AbstractModelComponent, self).__init__(). this is mainly for calrity sake and consistency, as the first way is done thoughout the python  code.

neworderofjamie:
Sure - feel free to change any styling details you see fit..

alan-stokes:
point 8: i note for the intergrate_and_fire and leaky_integrate models, no python docs were done for the properties and setters, top file, under class name etc. I recommend using a pep8 code checker, (id personally recommend pycharm and set all its code style controls up to max). Im filling them in as I go with the merge, but please try to fill these out in future, as my ide throws lots of wollies about these things now a days (most of our code makes my ide scream now. lol).

alan-stokes:
point 9. can you please please python doc your work. all your python classes in the extra models module are missing all the documentation.

alan-stokes:
point 10. ALL your c code in new models needs to have proper doxygen comments. please refer to examples such as 

https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/neural_modelling/src/neuron/models/neuron_model.h

for what is expected for documentation.

neworderofjamie:
Yawn...I think all my classes just derive off other stuff, do I still need
to copy and paste the same pydoc definitions over?

alan-stokes:
yep. to steal the concept from testing world. 100% covereage is preferred. We're starting to do it in the code of spynnaker, but it would be nice if the devleopers supported this process. esp when they are the ones who are expects in the field and thus the best qualified to comment what each function actually does.

alan-stokes:
point 11. 80 char per line. Ive come across one in vogels_2011_time_dependency on line 50. please check your code and try to follow the pep8 compliance in both c and python.

neworderofjamie:
Sure - can we just merge this for now though? I don't have time to do this
stuff right now and I'm more worried about this stuff getting lost in some
abandoned branch

alan-stokes:
question. 1 how does the recurrent_time_dependency map to the 4 different types of recurrent_time_dependency in the c code? as making a component hd5 for this seems obpague.

alan-stokes:
the stuff is being merged into my component tracker branch (which is how im doing your pull request) and by doing so, im also adapting mine to go component md5. this brnach is then going to be reviewed by rowley when i review his upgraded spinnman. not going to disappear. 

I dont have time myself to go though the code docuemnting it, yet im making the time now as i go through bits of this merge which i need to adjust. But i could rightly bounch the entire pull request due to lack of documentation. Please get around to it when you can and inform me, so i can update the branch, or if you want i can commit what ive got now, and you can start working off component tracker itself for the documentation.

alan-stokes:
qu 2. which python models do the synapse strcture weight state accumlator.impl and accumulator_only.impl and accumulator_window.impl map to?

neworderofjamie:
Tbh, just delete everything apart from the dual_fsm and the
accumulator_only structure - The rest are basically just past experiments
Simon and I were doing which there's no point in continuing to maintain.
There's no builds setup for them and, as you noticed, no Python..

rowleya:
I am happy to keep things in C without python code if it could be useful in the future.  One thought is to make an “experimental” folder for this sort of code.

neworderofjamie:
That was kinda why that work originally lived in its own repository...

alan-stokes:
ive placed them into their corrapsonding experiemntal folders. Not sure i like the apporach, but will leave it be for now. 

I persume then that recurrent_time_dependency links to timing_recurrent_duel_fsm_impl
i currently havent reached a new python class for a synpase_structure_state_accum_only_impl, but i do hope there is one. 
 
qu3.  theres a new file called plasticy_helpers, which is only used by timing_dependancy's can it be pushed down into the abstract_timing_dependency or reated as a new inhirtted class? such as *uses_look_up_table* i also note that write_exp_lut is ambigious, write_exp_look_up_table would be more clear.

alan-stokes:
jamie. you said to remove all but the dual_fsm in the recurrent_time_dependency's, but theres a if_curr_exp_stdp_recurrent_pre_stochastic_multiplicative model in the make files. I also cannot see how you distinquish between synapse_structure_weight_state_accumulator_only_impl and synapse-structure_weight_impl in the python side. please explain how this distinction is created.

neworderofjamie:
Python has no distinction on those synapse types, they're all handled
by PlasticWeightControlSynapseRowIo with different numbers of header words.

neworderofjamie:
Python has no distinction on those synapse types, they're all handled
by PlasticWeightControlSynapseRowIo with different numbers of header words.

alan-stokes:
Hi Jamie

So having had a discussion with rowley and yourself. A few things it seems came out of the wash which were not apparent before. Could you please verify if the conclusion is valid from the logic below:

1. synapse_strucutre is inexplicitly linked to the timing dependency, as in a each timing dependency requires a specific synapse_structure.
2. Currently these are defined in the make files for each neural model, and are linked during code "joining", which gives the impression its a component.
3. We will be going down the breaking up of the c code into more components and having them tied together in the near future, therefore removing requirements in make files that are not components is desirable.

Now if this is understood correctly, then the miscommunciation that we've had the last few days (renemebr that if i've had this miscommunciation, its a bet other end users could easily do so too) is due to the fact that the synpase structure being defined in the makefile gave it the impression of a component that was interchangable, whilst it is indeed not so.

Thus I respect the reasoning to remove the requirement of it requiring a component md5 as it isnt a "component", but could we then remove it from the makefile requirements by having the synapse_dynamics (ill use synapse_dynamics_stdp_impl for example) on line 98, 226, 241, 245 call the timing dependency which in turn calls its corrapsonding syanpse_strucutre and have each timing dependency import said synapse_structure impl?

This then makes it obvious that its directly linked to a timing dependance and thus end users, and myself, would be intrequed to think it was a dial that could be tweaked. 

Do you agree? or is there something that mysef and Rowley are missing in this regard?

neworderofjamie:
I don't really understand what the overall purpose of your 'component'
change is, so it's very difficult for me to comment. I don't understand how
the desire to formalise 'components' has any impact on the concept of
synapse structure and, while I agree that including header files using
Makefiles is somewhat unpleasant, it is a useful way of providing
compile-time polymorphism in C. As such, I don't understand why that is
suddenly only 'allowed' for 'components'.

Perhaps if I explain the purpose of the synapse structure it'll help you
understand their purpose - which is general and decoupled from STDP (there
are other learning mechanisms I'm trying to work on that have no concept of
timing or weight dependences, but still need to store data in plastic
synapses). Basically the function of the synapse structure is to convert
the packed data (plastic_synapse_t) stored in the plastic region of a
synaptic row into an unpacked format (update_state_t) using
synapse_structure_get_update_state which the learning rule then works with.
In the case of standard STDPish rules, this update_state_t is simply a
weight_state_t, as defined by the weight dependence, but it can also
contain traces etc used by the timing dependence. Once an individual
synapse is processed, the update_state_t is then repacked
using synapse_structure_get_final_state, the weight is extracted from this
using synapse_structure_get_final_weight and inserted into the ring-buffer
and the result of synapse_structure_get_final_synaptic_word is written back
to the plastic region.

rowleya:
I think that the point that Alan is making is that if you have a particular timing rule, it won’t compile unless you have a particular synaptic structure.  Therefore, the dynamic C header inclusion is unnecessary in this case, and we can simply refactor the timing rules to include the appropriate synapse structure header directly.  It doesn’t mean that this isn’t a “component” that can be used multiple times in the same code, but more that it isn’t an interchangeable piece of code that you can just plug in with any other piece in isolation.

I will note at this point that I was responsible for splitting the synapse structure out of the STDP during the refactoring of the code, as I saw commonalities in the various rules that could be extracted this way.  I then took this to be an exchangeable piece without realizing that it really wasn’t!  I am therefore quite happy to look at further refactoring this to take away this problem, although it will probably make sense to first get these changes pulled in.

In summary, don’t worry about the MD5 for the synapse structures, but equally, for now, leave them as a dynamically included header until we can get around to refactoring this problem away.

PR #95, opened 2015-06-26
Reload without reinjection

alan-stokes:
pushes reload without reinjection into master.


rowleya:
This is currently awaiting missing files in SpiNNFrontEndCommon - during a merge with master it was found that a file had been changed in master but in this branch it has been removed from this project and was not available in SpiNNFrontEndCommon.  Once this file is restored, the merge from master can be done.

rowleya:
Also appears to depend on changes in PACMAN that I can't find (no matching branch and not in the pull requests)

spynnaker/pyNN/models/spike_source/spike_source_array_partitioned_vertex.py
```diff
@@ -1,15 +1,52 @@
...
 
+# spinn front end common imports
+from spinn_front_end_common.interface.buffer_management.buffer_models.\
+    sends_buffers_from_host_partitioned_vertex_pre_buffered_impl \
+    import SendsBuffersFromHostPartitionedVertexPreBufferedImpl
+
 
 class SpikeSourceArrayPartitionedVertex(
-        PartitionedVertex,
+        PartitionedVertex, RequiresRoutingInfoPartitionedVertex,
```
rowleya: RequiresRoutingInfoPartitionedVertex doesn't exist in PACMAN, and there is no branch.  Should this actually be in sPyNNaker, and is then just missing?

spynnaker/pyNN/utilities/database/spynnaker_data_base_interface.py
```diff
@@ -0,0 +1,179 @@
+"""
+SpynnakerDataBaseInterface
+"""
+# front end common imports
+from spinn_front_end_common.utilities.database.\
+    front_end_common_data_base_interface import FrontEndCommonDataBaseInterface
```
rowleya: This doesn't appear to exist!

rowleya:
There are still some missing files.  Everything else appears OK.

PR #97, opened 2015-06-29
Modifications to sPyNNaker to support enhanced external I/O…

AlexRast:
Includes the ability to instantiate a LivePacketGatherer directly; gets rid of "spikes_per_second" and "ring_buffer_sigma" parameters for vertices that don't need them. New EIEIO_parameters object to manage I/O parameters more consistently. Database now outputs population type information for checking, and checks against type for LivePacketGatherer. Various other small fixes.

This pull is part of an integrated suite including changes to:

PACMAN
SpiNNFrontEndCommon
SpiNNMachine
SpiNNMan
sPyNNaker
sPyNNakerExternalDevicesPlugin
Visualiser

which should be approved as a bloc.

When merging be sure that all the other projects are also merged or else crashing will probably result.

### DIFF
--- a/spynnaker/pyNN/__init__.py
+++ b/spynnaker/pyNN/__init__.py
@@ -19,6 +19,7 @@ from spynnaker.pyNN.spinnaker import executable_finder
 from spynnaker.pyNN import exceptions
 from spynnaker.pyNN.utilities.conf import config
 from spynnaker.pyNN.utilities.database.socket_address import SocketAddress
+from spinnman.messages.eieio.eieio_type import EIEIOType
 
 # neural models
 from spynnaker.pyNN.models.neural_models.if_cond_exp \
@@ -52,6 +53,12 @@ from spynnaker.pyNN.models.spike_source.spike_source_array \
 from spynnaker.pyNN.models.spike_source.spike_source_from_file \
     import SpikeSourceFromFile
 
+#utility parameters
+from spinn_front_end_common.utility_models.live_packet_gather_parameters \
+    import LivePacketGatherParameters
+from spinn_front_end_common.utility_models.reverse_ip_tag_parameters \
+    import ReverseIPTagParameters
+
 # connections
 from spynnaker.pyNN.models.neural_projections.connectors.all_to_all_connector\
     import AllToAllConnector
@@ -72,6 +79,8 @@ from spynnaker.pyNN.models.neural_projections.connectors.\
     DistanceDependentProbabilityConnector
 from spynnaker.pyNN.models.neural_projections.connectors.\
     fixed_number_post_connector import FixedNumberPostConnector
+from spynnaker.pyNN.models.neural_projections.connectors.from_file_connector \
+    import FromFileConnector
 from spynnaker.pyNN.models.neural_projections.connectors.small_world_connector\
     import SmallWorldConnector
 
@@ -115,7 +124,6 @@ _spinnaker = None
 
 # List of binary search paths
 _binary_search_paths = []
-
 
 def register_binary_search_path(search_path):
     """
@@ -306,7 +314,7 @@ def Projection(presynaptic_population, postsynaptic_population,
 
 def NativeRNG(seed_value):
     """
-    fixes the rnadom number generators seed
+    fixes the random number generators' seed
     :param seed_value:
     :return:
     """
--- a/spynnaker/pyNN/models/abstract_models/abstract_population_data_spec.py
+++ b/spynnaker/pyNN/models/abstract_models/abstract_population_data_spec.py
@@ -222,8 +222,8 @@ class AbstractPopulationDataSpec(
 
     def generate_data_spec(
             self, subvertex, placement, subgraph, graph, routing_info,
-            hostname, graph_mapper, report_folder, ip_tags, reverse_ip_tags,
-            write_text_specs, application_run_time_folder):
+            hostname, graph_mapper, report_folder, write_text_specs, 
+            application_run_time_folder):
         """
         Model-specific construction of the data blocks necessary to
         build a group of IF_curr_exp neurons resident on a single core.
--- a/spynnaker/pyNN/models/abstract_models/abstract_synaptic_manager.py
+++ b/spynnaker/pyNN/models/abstract_models/abstract_synaptic_manager.py
@@ -445,10 +445,8 @@ class AbstractSynapticManager(AbstractProvidesIncomingEdgeConstraints):
         for subedge in in_sub_edges:
             sublist = subedge.get_synapse_sublist(graph_mapper)
             sublist.sum_n_connections(total_items)
-            edge = graph_mapper.get_partitionable_edge_from_partitioned_edge(
-                subedge)
 
-            if edge.synapse_dynamics is None:
+            if stdp_max_weight is None:
 
                 # If there's no STDP maximum weight, sum the initial weights
                 sublist.max_weights(absolute_max_weights)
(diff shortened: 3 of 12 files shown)


rowleya:
I have had a look through these changes.  Unfortunately, there are a lot of changes contained within the branches that we discussed but decided not to go ahead with.  These include:

1) Differentiating between vertices that take certain parameters e.g. spikes_per_second and ring_buffer_sigma.  Whilst I agree that this is a bit odd at present, we realised that going down this route might mean having a number of differentiations, making adding new vertices in the future more complex (which interface do you pick?).

2) Taking the iptags and reverseiptags and turning them into vertices which take these parameters.  These were passed in at data spec.  This was done in a similar way as is done with other things that come from the automatic configuration via pacman.  Differentiating the tags in this way seems odd.

Beyond these changes, from what I can see, this branch (across the projects) just updates the database to generate a class for the LivePacketGather object, and then uses it in the visualiser.  As discussed, this comes with its own issues; for example, it isn't difficult for me to create an edge to 2 LivePacketGather objects, and so this will again fail.  I don't see this as an improvement, just a difference...

AlexRast:
At the risk of reopening old arguments, here's my thinking.

On 21/07/15 10:53, Andrew Rowley wrote:
> I have had a look through these changes. Unfortunately, there are a lot of
> changes contained within the branches that we discussed but decided not to go
> ahead with. These include:
>
> 1) Differentiating between vertices that take certain parameters e.g.
> spikes_per_second and ring_buffer_sigma. Whilst I agree that this is a bit odd
> at present, we realised that going down this route might mean having a number of
> differentiations, making adding new vertices in the future more complex (which
> interface do you pick?).

My opinion here is that if various vertices of necesssity need different 
interfaces because the nature of what they need is fundamentally different, then 
this sort of diversity is forced upon us. spikes_per_second and ring_buffer 
sigma were "point solutions" designed to overcome the problem of weight scaling 
with various neuron models; this interface is inherently specific to those sorts 
of vertices that implement a neuron model with both synaptic weights and a 
ring-buffer internal implementation of synaptic currents/conductances. It was 
always a fairly rough-and-ready solution; in fact maybe the whole way this is 
implemented in terms of interfaces needs to be revisited.

Notwithstanding the point is, that we have such vertices as live packet 
gatherers, robot controllers, spike sources and other such things which operate 
in fundamentally different ways from ordinary neuron models is in my opinion a 
strong indication that a one-size-fits-all interface is naïve. It's appropriate 
to have a parent top-level vertex class that has the minimal interface required 
for *anything* that might end up as a vertex, but at some level, the interfaces 
are going to diverge due to different requirements. Otherwise you end up with a 
"wide interface" requiring every possible parameter that *any* vertex might 
need, and this actually makes adding new vertices in the future possibly even 
more complex because then every time you create a vertex that needs some new 
parameter, that parameter would have (unnecessarily) to be added to the 
interface for every vertex: more code changes, in more places. The subclassing 
model answers the question "which interface do you pick?" with "you choose the 
minimal interface that meets your needs while adding any necessary parameters in 
a subclass for your new vertex".

> 2) Taking the iptags and reverseiptags and turning them into vertices which take
> these parameters. These were passed in at data spec. This was done in a similar
> way as is done with other things that come from the automatic configuration via
> pacman. Differentiating the tags in this way seems odd.

Here I'm using the same methodology as was as I understand it the design model 
for the entire toolchain: that you use classes to contain any type-specific 
behaviour so that a minimum of if-tests are required; the class automatically 
instantiates the necessary functionality. In this case, this means that only 
vertices of a type that actually have tags attempt to instantiate them - and 
likewise their interfaces have these additional parameters whereas vertex types 
that don't need them, don't need to take them, never attempt to instantiate 
tags, and in general are prevented from doing anything strange with tags by 
simply not handling them at all. In addition vertices with tags now conveniently 
hold their own tags themselves, again minimising the possibility that something 
will go wrong with interacting tags in a global dict. The tags themselves are 
still allocated at data spec time. Again this also avoids the "wide interface" 
problem in DataGeneratorInterface. The basic interface should not be taking 
parameters that only apply to a subset of vertex types - these parameters should 
be contained within the vertices themselves and instantiated through them when 
required - otherwise you run into the possibility of endless extension of 
DataGeneratorInterface to cope with yet more hitherto unforeseen parameters.

> Beyond these changes, from what I can see, this branch (across the projects)
> just updates the database to generate a class for the LivePacketGather object,
> and then uses it in the visualiser. As discussed, this comes with its own
> issues; for example, it isn't difficult for me to create an edge to 2
> LivePacketGather objects, and so this will again fail. I don't see this as an
> improvement, just a difference...

The branch was initially designed to get around the limitation expressed in bug 
report sPyNNaker 47 "reverse ip multicast soruce / spike injector instaniation 
with prefix". Also I believe this allows testing of the fix in bug report 
sPyNNaker 46 "reverse iptag multicast source has a bug with is vs =="

The branch allows you to instantiate reverse iptag multicast sources directly 
(as well as live packet gatherers) and thus gain access to the full interface 
rather than a simplified subset.

The modifications to the database are somewhat peripheral to the main changes 
and were really designed to circumvent problems in getting spike_io to work 
because of naming conventions. At the time these were made I wasn't specifically 
thinking about multiple gatherer objects for the same vertex but any necessary 
changes would be easy to implement because the underlying DB queries for each 
vertex' output details returns all live packet gatherers associated with that 
vertex. However what the database was doing before, looking for a specific name 
for the live packet gather vertex, is intrinsically fragile and was, I suspect, 
again a temporary expedient to get things working.

rowleya:
I have no issue with you raising these points, but it is worth separating out the issues that you are trying to solve in to separate branches so that they can be discussed separately.  As it stands, these “fixes” are all lumped together, so we have to decide if we want them all or none of them.

AlexRast:
A problem here is that there is a limited amount of separation that could really 
be done. The changes that differentiate between vertices based on parameter 
interfaces are necessary because it is precisely these additional parameters 
(spikes_per_second, ring_buffer_sigma) that was preventing the 
ReverseIPTagMulticastSource from being instantiated independently. This in turn 
requires the tags to be owned by the typed vertices because by definition such 
vertices MUST have tags - this is the basis of their differentiation.

About the only thing that could be separated is the database modifications, but 
these are a small and ancillary part of the update and in any case it would seem 
a little retrogressive and circuitous to then embed within the rest of the 
update naming conventions that are only the result of a design quirk in the 
current implementation of the database interface - only to then change them 
again once the database changes were incorporated.

There are always going to be updates that require fairly widespread 
modifications because the nature of the problem or additional support required 
has exposed limitations at the design level that need to be rectified before the 
requested support can be provided. In this case I think that what has happened 
is an attempt to layer 2 independent support components - support for external 
devices and support for variable weight scaling - over an existing codebase 
turned out to involve incompatible modifications until some changes were made to 
the underlying existing codebase.

Veering a bit off-topic, I think perhaps a vertex that requires 
spikes_per_second and ring_buffer_sigma ought to be subtyped as a 
RingBufferedVertex. Many of the vertices we have neither have nor need ring 
buffers, there's no reason to expect even other neuron types in future would 
necessarily require this structure, and it should be reserved for those vertices 
that actually use it. For later implementation but I think worth considering.

neworderofjamie:
To avoid adding further levels of abstraction, couldn't the ring-buffer
specific things be moved up into whatever the base class is for vertices
that represent sub-populations of neurons?

rowleya:
The ReverseIPTagMulticastSource (in SpiNNFrontEndCommon) doesn’t use these parameters at all; it is the SpikeInjector in SpyNNakerExternalDevicesPlugin that introduces these parameters.  This allows it to be used via the PyNN interface.  That said, these are parameters passed in from the config files; I am wondering if they can actually be read in from the config file directly into the vertex to avoid the issue altogether...

In any case, can you tell me why you want to instantiate ReverseIPTagMulticastSource in isolation from PyNN?  i.e. what is the use case that this is addressing?

PR #101, opened 2015-07-02
Optimise getspikes

alan-stokes:
This again gives a pretty massive speedup

NEEDS FIXING FIRST!


alan-stokes:
Jamie, this doesnt work properly. I get broken lines of synfire chains. So please fix and then respond to this pull request when done. I alas didnt test it beforehand (was distracted)

neworderofjamie:
What example are you running?

alan-stokes:
standard synfire script from pynn examples.

neworderofjamie:
Good spot - No idea what you were doing with the branches so fixed it in optimise_getspikes_fix branch!

rowleya:
This has actually been merged in via another request

PR #102, opened 2015-07-02
Implemented part of the PyNN 0.7 "procedural"/"low-level" API

apdavison:
see http://neuralensemble.org/trac/PyNN/wiki/LowLevelAPI

This is not ready to be merged, as the "record_X()" functions still need to be implemented, but I would appreciate a quick review of what's there so far, as this is my first attempt to work with the sPyNNaker code.

### DIFF
--- a/spynnaker/pyNN/__init__.py
+++ b/spynnaker/pyNN/__init__.py
@@ -320,3 +320,55 @@ def get_current_time():
     """
     global _spinnaker
     return _spinnaker.get_current_time()
+
+
+# ==============================================================================
+#   Low-level API for creating, connecting and recording from individual neurons
+# ==============================================================================
+
+def create(cellclass, cellparams=None, n=1):
+    """
+    Create n cells all of the same type.
+
+    If n > 1, return a list of cell ids/references.
+    If n==1, return just the single id.
+    """
+    return Population(n, cellclass, cellparams)
+
+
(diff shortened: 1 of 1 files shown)


spynnaker/pyNN/__init__.py
```diff
@@ -320,3 +320,55 @@ def get_current_time():
...
+        """
+        Connect a source of spikes to a synaptic target.
+
+        source and target can both be individual cells or lists of cells, in
+        which case all possible connections are made with probability p, using
+        either the random number generator supplied, or the default rng
+        otherwise. Weights should be in nA or µS.
+        """
+        connector = FixedProbabilityConnector(p_connect=p, weights=weight, delays=delay)
+        return Projection(source, target, connector, target=synapse_type, rng=rng)
```
rowleya: The lines above here are indented by 8 rather than 4

rowleya:
This looks good - just a minor indentation issue

alan-stokes:
just a note, having not impleimented errors to functions not yet supported should not be a limiting factor for mergability. A lot of the standard pynn 0.75 interface has these for functionality which doesnt exist yet. such as population views.

rowleya:
I have just made some changes to this, and pushed them back to our repository (I don't think I can write to yours).  I will merge in the changes from the local branch (sorry if this makes it more complicated for you to get them!).

apdavison:
great!

PR #110, opened 2015-07-21
Fix live spikes

rowleya:
Note that this needs the spinn_common changes in the same branch name to work (and you need to make and make install spinn_common)


alan-stokes:
I have no idea where my comment went. But ill retype it here anyhow.

So had chat with jamie over the devide and am happy to merge.
Having looked at the code, I have 3 reservations which are minor, and therefore have been pushed in anyhows. But listing them here for future knowledge.

1. The live pakcet gatherer c code holds just 256 entries for the 2 types of pakcets it can receive (mc, mc pay). But this seems small given it deals wtih all the boards live packets. A back of the envolpe calc says we could push this up to 640 each and still have some DTCM left over. Maybe we should consider this?
2. The way the ciruclar queue changes of the makefile look, it seems (and this could easily be me misunderstanding) that now every model has a circular queue linked in? I'm not sure how smart the compiler is in getting rid of none used code, but given we are execution size limited as well, is there a way to not force everything to have this code?

3. The live packet gatherer now checks if it has successfully added a packet to its queue, but does nothing if it didnt. I think as this is the first place where packet loss can be detected reliablly, we should actually store this loss and report it to the end user via provanence data. I went and built this in a seperate brnach called # new_prov_live_packet_gatherer and will be pull requesting these changes for review shortly. 

anyhow, good work rowley. 
alan

PR #112, opened 2015-07-21
Speed up data transfer 2

rowleya:


neural_modelling/src/neuron/spike_processing.c
```diff
@@ -163,13 +169,37 @@ void _dma_complete_callback(uint unused, uint tag) {
...
+                    current_buffer->row);
+
+                // Print out the row for debugging
+                for (uint32_t i = 0;
+                        i < (current_buffer->n_bytes_transferred >> 2); i++) {
+                    log_error("%u: 0x%.8x", i, current_buffer->row[i]);
+                }
+
+                // Don't go into real RTE here to allow proper debugging
+                vic[VIC_DISABLE] =
```
alan-stokes: No idea what this bit of code gains you over rt_error(RTE_SWERR)

why are you trying to fake a error during normal runtime behaviour?

rowleya: This was an attempt to avoid the cpu stopping on an error, thus allowing you to read DTCM.  Without this, an RTE stops the core I think, so you can't issue scp commands to that core.  The issue I was trying to debug turned out to be a write issue though, so I don't mind this being replaced if it is preferred.

Sent from my android device.

-----Original Message-----

PR #117, opened 2015-07-29
added support for infinite running and tested with a modified version…

alan-stokes:
sorts out infinite runs and thus alows the push bot pull request to be compelted


neworderofjamie:
Looks good

PR #139, opened 2015-08-26
spike source possion now works with rate 0

alan-stokes:
as title states


rowleya:
Unfortunately, this relies on the Database changes in another branch, and so breaks the SpynnakerLiveSpikeConnection (it can't read the database).  This will have to wait for another release.

PR #145, opened 2015-10-02
Split input from model

rowleya:
This branch separates the core neuron model from the input (current or conductance), the threshold (currently only static in this module) and the synapse type (e.g. exponential), as well as doing some refactoring to tidy up the class heirarchy in python, and make model writing easier than ever!  Note that spikes_per_second and ring_buffer_sigma have been removed from the interface, and are now automatically read from the config file if not provided by the model itself.


neural_modelling/src/neuron/plasticity/common/maths.h
```diff
@@ -11,14 +11,22 @@
...
+    const uint32_t num_words = (num_entries / 2)
+                               + (((num_entries & 1) != 0) ? 1 : 0);
+
+    // Copy entries to LUT
+    spin1_memcpy(lut, start_address, sizeof(int16_t) * num_entries);
+
+    // Return address after words
+    return start_address + num_words;
+}
+
```
neworderofjamie: What was the purpose of this? It wasn't inline because it's not performance critical and is called in multiple places so will potentially bloat the code size!

rowleya: Yes, I did see this as potentially an issue.  My only real reason for this was to make the Makefile a bit nicer, but if it is going to be contentious, I don’t mind moving it back again, although the code bloat from those few lines is probably not going to be much!

One thought is to move this in to a library somewhere instead (e.g. spinn_common).

PR #148, opened 2015-10-19
Multi-Run implementation

Scaatis:
The main changes are to spinnaker.py and the c_main functions of the various models. However, some changes to buffers were also necessary to allow rewinding and resending of input data.


alan-stokes:
Hi felix, 

so it seems github cant support my line comments on this pull request due to the number of changes that are acutally involved. So ill need to give file names and line numebrs here to link to the code.

spinnaker.py lines 944 to 960: _tell_cores_to_exit:
Even thouhg I can see what you want to do, and actually im happy with the underlying logic. Im a bit worried about the relaiability of this code. Theres no checks to see if after the load of pacekts haev been sent, if they all have been reiceived. SDP is unreliable, and therefore say you had 400 cores to put into exit, it could be that only 200 recieve the exit command. To make this more reliable, we need to send the blocks, then scan the machine for cores that are still in the sync state in case they hadnt arrived, and if so, repeat till they have been received. Due to hwo different the branch that im merging into is, ill do it here.

alan-stokes:
spinnaker.py lines 273 to 288. Having looked here, and seen commended out lines, Im rather worried about you saying this thing ran correctly. If you havent fixed the new time allocation, then how has the cores reran for a specificed time period? Its also worth noting that the data.append only appends 24 bits of a 32 bit number on lines 282 to 284.

Finally, this code doesnt support the detection of any loss of these packets. Having had a chat with Luis and Steve Temple here at manchester, they came up with a noval way of doing this, which si that when the runtime is received, it can move from sync-1 to sync0 which then allows us to scan for cores not in sync0 and therefore allow us to see which cores have failed to receive the packet and resend it. Yet again, due to the differences in the tool chains, ill fill this out myself.

Scaatis:
Regarding `_tell_cores_to_exit`:
I'm actually very unhappy with this code. How does sending the STOP code differ from sending the SYNC1 or SYNC0 code? Why do I just call `Transceiver.send_signal` for SYNC, but I build my own SDP packet here? Why is this function not in `FrontEndCommonInterfaceFunctions`.

Also, though not directly related to this line, the C code where this signal is received is not working. I'm fairly certain that the SCP packet is not constructed properly

regarding the runtime sending:
That's your code : )
I removed this when I realized that my way of receiving SDP packets did not work. Obviously, setting a different simulation time did not work. However, for our test case, it was sufficient to only run for the same amount of time each run until someone with a clearer idea of SCP reviewed that section.

alan-stokes:
Hi Felix, 

the changes in this brnach actually mean that the _tell_cores_to_exit acutally fits better as a callable function from the equiv of FrontEndCommonInterfaceFunctions (thats what I'm currently doing as I type. lol)

On the send runtime code (which [assuming it is my code] was done without any testing, nor was complete when I gave it to you ;-) ) but to anwser your question. We'll just need to see what was making it fail, it should easily be fixable.

On why the stop code differs from the signals, the signals are used to support barriarization's of distrubuted cores to, as close to as possible, support a machine scoped syncronised start. Therefore a signal just pushes you back into the code's natural flow. Your stop would need to actually kill a binary natural flow, and this is not currently done via signals and sark. It does open up the question of wondering if it could be done..... Ill throw it at the low level software guys at my next opporuntinty.

alan-stokes:
Having had a chat with Steve Temple, this functionality is coming in the next release of the low level software, but in the current release that this tool chain works off, its not implimented, and therefore we need to do the sdp check, resend as needed cycle.

alan-stokes:
Hi Felix, 

For the getSpikes gsyn, v functionality from the population. You seem to be adding parameters to the PyNN population interface, this is a serious no no. What I think is the more cleaner way to handle this is to call the spinnaker.py to get the runtime value and have the population contain its last_read_state as you hint with the self._last_spike_read variable. Its also worth noting that the functionality define by PyNN for get-spikes in a multi-run mode is that it returns all the spikes from all the current runs (assuming no reset has been called in the meantime) and so you also need to read the spikes_cache_file before saving to it. I'm guessing this will need you to filter it on your application level. Yet again, ill modify it as the pull request is rather complicated.

alan-stokes:
having looked at this more cloesly, it seems that 2 things have been missed (one i think was myself when i started):

1. the pops and projections are not what spynnaker canc heck for the changed param (not sure how this branch has ever run correctly). So im pushing this into the vertices and edges so that the changes can be detected [think this one was my slip up].
2. if any vertex is set to record, they must all have their get_spikes called before the next run. Therefore all these temp files, last spikes_time etc needs to be put within spinnakwer.py, not the pops and projections themselves. 

Yet again, will do so myself.

alan-stokes:
having dug a bit deeper, 
1. it looks like there was logic to this. pops have data which is only ever stored there, and we now store the pops and projections in spinnaker.py (looks like whoever did this bit was acutally intelligent, and i refuse to accept it was me [dont do intelligent on a good day])

alan-stokes:
ok. so having regiggled all the getSpikes, getV and getGsyn and adjusted the logic flow to hopefully wrok with both multiple runs and a real reset (this mode of operation makes reset nearly free, so plugged it in as a freebi)

Ive finally got the rest of this pull request and am rather happy with the generation of a function for init'ing recording (one day we can jsut set up reocrding wtihout all this boiler plate) as seen in the possion and  c_main. 

given the changes from front end common, we can remove the event_wait calls in delay, c_main, possion as these have been encapulated in the single simulation call.

Due to the functionality being concerned with is resume and pause, this makes reinitilsation of the c_main not required anymore (reset kills the binaireis and reloads them and applciation data) and therefore im going to remove that bit. 

Im aslo going to do the same for the sdp packet receive and use sergio's fixes to support port calls.

alan-stokes:
Hi felix,

Incase you've not seen all these pull reuqests I've set off now:
SpiNNakerManchester/sPyNNakerExternalDevicesPlugin#19
SpiNNakerManchester/sPyNNaker#167
SpiNNakerManchester/SpiNNMan#43
#29
SpiNNakerManchester/SpiNNakerGraphFrontEnd#10
SpiNNakerManchester/PyNNExamples#19
SpiNNakerManchester/PACMAN#27

These contain a PyNN matching implemntation of your Reset which has been molded as a pause/resume as well as a reset. It updates the runtime and exits the cores cleanly, and even supports you changing the SpikeTimes of a SpikeSourceArray without going back though mapper (whilst also being tied into the enw way the tools execute the functions).

The only issue to date is that if you change the application graph (apart from your spikes times of the spike soruce array) it will raise an error stating this isnt supported. I did try to get it to support this, but bug after bug after bug has got me to the point where I need to close this for the time being, so an exception has been placed instead. Ive raised the issue as an issue for future notice (SpiNNakerManchester/sPyNNaker#168)

Your welcome to try to debug it if you want, I just dont have time right now.

The only cavieats from your point of view is that the spike source array's spike times have to be shifted to match the new runtime, so using the psueduo code example below:

runtime=1000
total_runtime = 0
ssa = p.pop(spike_source_array, spike_times = [0])
p.run(runtime)
total_runtime += runtime
sa.tset(spikes_times=[1001])
p.run(runtime)
total_runtime += runtime
time = 1
sa.tset(spike_times=[total_runtime + time])
p.run(runtime)

You need to adjust your injected spike times to be in line with the new simualtion time frame. So the tset array needs to be adjusted by the runtime you've currently got to, as the bottom lines try to indicate. When your receiving spikes, you'll need to window off the spikes to be the ones your only interested in (as get_spikes must return all the runs versions of spikes between setup and either end or reset) (This needs to be updated in the PyNN interface though).

Given all this, I'm going to close your pull requests and thank you for your contributions, and welcoem your feedback on these pull requests and ask that if its not too much trouble for you, to mvoe over to the master once these changes have been pushed in so that your not going to be left hanging as new releases come by.

Alan

PR #166, opened 2015-11-09
Buffer out 20150724

sergiodavies:
New classes and functions to support output buffering functionality


rowleya:
Something as simple as running a simulation for 10 seconds whilst recording spikes (no buffering), v (1MB buffer for 4MB v) and gsyn (2MB buffer for 8MB gsyn) requires the ability to read at about 1MB/s to keep up (~8Mb/s).  This speed is not possible unless the multiple-packets-in-flight is enabled; without this the maximum is about 0.5MB/s (4Mb/s), so the buffer can never keep up.  Unfortunately, the multiple-packets-in-flight protocol is going to have to be disabled until multiple-boards can be used with it.

I have added an option to the buffering to disable it by default, but allow users who know what they are doing to enable it if they want it.  I have also added options for the size of the buffering regions.  This has been added to the default config file as well (all disabled of course), to let users know that these options exist.  Note that spike source array uses buffering for recording in any case, as you never know how big the buffer should be in this case.  This appears to function well.

PR #167, opened 2015-11-10
Work flow multi run

alan-stokes:
changes to get the graph front end to compile down with the new flow system.

Currently the ehat demo fails at runtime, not too fussed. There is a change to the generate dsg to return a lsit of file paths. but thats fiar enough to remove the assumptionm of one dsg file per vertex.


alan-stokes:
finally ready to be merged into master again via new brnach down stream

PR #187, opened 2016-01-26
Router compressor

alan-stokes:
allows bigger simulations to operate as router entries will be compressed. This also includes vertex_key_alloc so that keys are partitionable_vertex|partitioned_vertex|atom instead of malloc based.

should allow 100% microcortical column to operate correctly.


rowleya:
Replaced by adjustment for routing path branch

PR #190, opened 2016-02-10
Add default value to SpikeSourceArray spike_times

astoeckel:
SpikeSourceArray did not specify a default value for spike_times
in its constructor. This caused sPyNNaker to crash with the
weird exception

    File "sPyNNaker/spynnaker/pyNN/models/pynn_population.py", line 79, in __init__
      self._vertex = cellclass(**cellparams)
    TypeError: __init__() takes at least 5 arguments (5 given)

if an empty parameter dictionary was passed to the Population
constructor (which supposedly should not happen):

    sim.Population(10, sim.SpikeSourceArray, {})

This fix simply adds the missing default value to the spike_times
parameter. Note, that the parameter order in the
SpikeSourceArray constructor was altered.

### DIFF
--- a/spynnaker/pyNN/models/spike_source/spike_source_array.py
+++ b/spynnaker/pyNN/models/spike_source/spike_source_array.py
@@ -46,7 +46,7 @@ class SpikeSourceArray(
 
     def __init__(
             self, n_neurons, machine_time_step, timescale_factor,
-            spike_times=[], port=None, tag=None, ip_address=None,
+            spike_times=None, port=None, tag=None, ip_address=None,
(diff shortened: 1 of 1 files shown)


alan-stokes:
Hi astoeckel.

Just had a glance at this code, and found a bug. 

By setting a default parameter to a list, you force new instances to use that list instance. Its a annoying feature of python that such a defintion is the same as a static definition.

The better solution is to define that default as None, adn check inside the init for none and then create the list.

Im about to mvoe off to other stuff, so id hope you yourself fix this pull request.
regards.

astoeckel:
Thanks Alan, I fixed the issue.

PR #194, opened 2016-02-16
Portal hackery

alan-stokes:
- [x] build front end common function which generartes 105 virutal machine and runs partitionging on it
- [x] from partitioned_graph, deduce size of machine in temrs of boards.
- [x] Plug in a fake jonathon interface to start with, that returns a ip_address, bmp_address, size in x and y, ethernet connected chip ids etc.
- [x] Run standard operations till p.end and encapsulate chrash protection.
- [x] at end / crash end, call new function to deallocate machine
- [x] working in brnach
- [x] update with jonathon interface
- [x] working in branch and portal
- [x] review complete.


alan-stokes:
to make this work. needs changes to cfg file, as portal interface will make.

PR #196, opened 2016-02-26
Profiler update

mmikaitis:
Resurrected Jame's profiler. Allows inserting entrance and exit calls in low level C code that record timestamps and produce statistics from them using numpy. I have used this extensively for UG project report as well as Jamie has mentioned that this was used in the publications, hence the pull request.
- [x] Refactor into SpiNNFrontEndCommon to allow general use of profiler
- [x] Wait for merge of SpiNNakerManchester/SpiNNFrontEndCommon#145


neworderofjamie:
This was some fine code archaeology @mmikaitis! @rowleya, as the master of the Makefiles, it would be better if it was integrated, as it is in Nengo SpiNNaker, so that a separate set of  executables with profiling enabled is always built and these are swapped in when ``profile`` is set on the ``abstract_population_vertex`` (see https://github.com/project-rig/nengo_spinnaker/blob/master/spinnaker_components/Makefile)

neworderofjamie:
This was some fine code archaeology @mmikaitis! @rowleya, as the master of the Makefiles, it would be better if it was integrated, as it is in Nengo SpiNNaker so that a separate set of  executables is always built and these are swapped in for a vertex when ``profile`` is set on (https://github.com/project-rig/nengo_spinnaker/blob/master/spinnaker_components/Makefile)

mmikaitis:
Hi @rowleya  and @neworderofjamie . What is missing for this to be appropriate commit into trunk?

neworderofjamie:
No idea! I think @rowleya is rather busy, but I've assigned it to him in an attempt to bring it to his notice :)

rowleya:
Sorry, I have been busy preparing for a release of the software for the HBP project.  I will hopefully have some time to look at this later.  It is likely to be in conflict with a number of changes recently made at this point though, so it might take a bit of merging to get it working again...

mmikaitis:
Perfectly understandable! Thanks. And yes I can see the conflicts, I will try to rebase it onto the latest master version.

mmikaitis:
I have now rebased this onto current master and checked that it works. Some problems were with region address calculation, the way to calculate it has changed as seen from the last commit.

rowleya:
Thanks - the regions now have absolute addresses as you spotted.  I will try to look at this soon.

mmikaitis:
Another thing that I have noticed is that with newest master version we cannot use more than 1000 profiling samples anymore and the simulation fails after repeatedly showing "2016-04-02 13:54:07 INFO: Error getting state - retrying...". Is wonder if this is the consequence of the new regions in SDRAM and whether they use a lot of memory for something.

rowleya:
This normally indicates that too much traffic is flying around the machine to talk to it, or else that something has crashed (although I think this would be displayed as well).  Is this when used on a specific network?

mmikaitis:
I have used stdp_example.py to detect this error but it also happens with the other ones that I have.

rowleya:
Closing as now moved to SpiNNakerManchester/SpiNNFrontEndCommon#54

alan-stokes:
this is now working. There is a belief that we could tie this into the recording region and auto pause and resume. But the complexity of figuring out how often this is called per profile method is rather complex and therefore left as future work if it ever becomes criticle to do so.

Christian-B:
Depends on:
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/145

PR #197, opened 2016-03-08
Adjustment for routing path generation

rowleya:
Being opened in place of router_compressor branch


AlexRast:
Just a question - in brief, what was the reason to change edge constraints to 
being associated with partitions?

rowleya:
Although edges do exist (mainly because it is easier to make an edge-partition in bits), once the partition is created, the rest of the tool chain works on edge-partitions rather than individual edges.  Note that the edge-partition is different from a partitioned edge or a partitioned vertex!  It is a set of edges which all have the same pre-vertex.

AlexRast:
OK, got it. So basically what's happening is that the toolchain is grouping 
edges with common termini in order to simplify the routing job.

PR #204, opened 2016-04-01
Tester

alan-stokes:
tester for cfg file paramters


neworderofjamie:
I can see why they clearly don't belong where they were, but if you're
going to delete the STDP examples, can you add them to whatever the
examples project is called?

alan-stokes:
i point you to the PyNN examples module, where there are already 4 STDP examples.

neworderofjamie:
But are there THOSE ones?

alan-stokes:
sersiously jamie. 
1. mwh_synfire = synfire_if_curr_exp.
2. stdp_curve = stdp_curve.
3. stdp_triplet doesnt work, as we dont have a IF_curr_exp_stdp_mad_pfister_triplet_additive.aplx 4. svn_synfire is synfire_if_curr_exp, but with a extra SSA that does nothing. 
5. test makes some sort of poisson to if curr, but as its not got a usful name, im mapping it to be equiv of a va or prunnel example (which lives there).

we didnt even knew they existed till i was making this tester script, and it found them. From the point of the STDP exampels, they are either exactly the same or dont work in the first place.

neworderofjamie:
exactly, stdp_triplet is missing - It's cool and it was used in my workshop
presentation which I imagine is still given to people

alan-stokes:
stdp_triplet doesnt work, and therefore doesnt even get to go into PyNN examples on that ground. Fix it to work by all means, and then you can put it into the PyNN example folder yourself. But the current state of affairs is that it wouldnt work, and therefore giving it to end users will just expose us to unnessary grumbles in the another fine product from the nonesense factory release. 

if its being used in your presentation, why does it not work anymore, and why did you put it in a example folder in spynnaker in the first place?

PR #205, opened 2016-04-02
Poisson live rate

rowleya:
- [x] change Poisson data structures to allow rate to switch between fast and slow
~~- [ ] Add "play back mode" for rate changes and make this SpikeSourceVariablePoisson~~
~~- [ ] Move existing live functionality into SpikeSourceLiveRatePoisson~~
- [x] Add a "live rate update" connection


alan-stokes:
ive got a few misigivings about this impemntation. These are as follows:
1. this is entirely designed around the idea that you can have a syncronised external device that decides to make the change in possion rate. This i feel is only applicable in a very small number of cases, and is never guanrteed to be executed at the right simulation time step, due to interupt levels on host. Id have preferred a system which supproted both of the following modes:
1.1 a set of rate changes, with runtime time points at PyNN level, allowing users to ensure stuff happens at a given time period at compile time (play back mode).
1.2 a live appoach (such as whats already impliemnted) but which takes a runtime point which is added to the message. this allows you to do two things. 1. bank it for future useage using the strucutre supported in 1.1 (such as we do with the SpikeInjector) and 2. detect when the host is behind shecdule and thus the rate change is done later than expected (such as now if the time step was in the past) and record it for provenance gatherering. Otherwise this will undougtfully cause issues in reproducability. 
2. Youve overloaded the SSP to always be one that supports both static and changable at the PyNN level. Id have prefered a seperate version for live (such as SpikeSourcePoissonLiveRateChange ][and maybe even a SpikeSourcePoissonPlayBackrateChange]) which took these extra params whilst SSP kept to the orginial params avilable to the PyNN SSP. This makes it much more explicit for the end user to decide to use these different functionalities, instead of having them all merged into one object for PyNN users (explicit is better in this case i feel). 
3. Youve merged this live rate adaption functionality into the spynnakerLivespikesConnection, and I feel this is such a distinct piece of functionality, that it should have its own connecion type in its own right. Its not liveSpikes at that point, but a LiveRateConnection and the code itself is almost disjointed in the SpYnnakerLiveSpikesConnection anyhow. 
4.  I dont see any test scripts to be able to verify if the live mode works correctly, and how to teach end users on how to use this new functionality. 

This all said and done, and the little amount of time avilable for functionality to be put into the AnotherFineProductFromTheNonsenseFactory release means that I'm going to not do this pull request at this time and we can discuss it at a later date and see if we can rectify all the issues stated.

sorry old boy.

apdavison:
I'd like to second Alan's 2nd point. Rather than extend `SpikeSourcePoisson` it would be better to create a new model, e.g. `SpikeSourceVariablePoisson` or possibly two new models, one with a pre-defined array of (time, rate) changes and one which allows 'live' rate changes from an external device.

rowleya:
I never intended for this to make it into the release, just tidying up branches, and didn't want this to get too far behind with the future planned changes to the poisson source.  That said, it does have issues particularly with regard to the partitioning (in that it can't be partitioned when in live mode).  I would like to keep the poisson generation itself common but understand the other issues.

alan-stokes:
no worries, I was working on the logic that if it was mergable, then it would be better to have it in and then work on the partitioning issue as as a general issue (which you state, shouldnt be partitioned if in live due to need for a reverse_iptag). Im just trying to tidy as much functionality as I can before i froze the functionality (which ive basically done now, and am fixing doc issues and install issues [such as setup.py missing directories]).

alan-stokes:
either way, listing the misgivings means I wont forget them. oh and Rowley. YOUR ON HOLIDAY!

rowleya:
Ok, last comment; one of the main differences in this from master is that master separates slow and fast sources into two lists, whereas this keeps them as a single list with a boolean indicating fast or slow; thus a source can change between being fast and slow.  This will probably be the source of most conflicts with future changes.

dkfellows:
Merge is nasty; needs someone who knows what they're doing

rowleya:
This has been merged and updated, and could now be merged in.  The main changes are:
1. The Poisson source vertex is the same whether live or not.  The Poisson source supports the reception of multicast packets with payload which control the rate.  Note that these are expected to come down an edge in a "control" partition.
1. To allow live changes, you need to call ```p.external_devices.add_poisson_live_rate_control(poisson_pop)```.  This adds a ReverseIPTagMulticastSource with an edge (in the "control" partition) to the Poisson source.  Users can then send this multicast packets with payloads to be sent on to the Poisson source.
1. A new ```p.external_devices.SpynnakerLivePoissonConnection``` has been created to facilitate the changes to the rate.  This takes the label of the Poisson source to control, and works similarly to the SpynnakerLiveSpikesConnection, except that it translates the rates into S1615 payload values and sends these along with the keys.

I have also modified sPyNNaker8 and have an example script, but would like to know if it is worth doing a pull request for i.e. if this pull request is to be accepted before I also add a pull request on those modules...

alan-stokes:
Will get to it once ive fixed my pr's and done chrsitians pynn 0.9.
Hopefully ill get to it before share key becomes criticle

neural_modelling/src/spike_source/poisson/spike_source_poisson.c
```diff
@@ -352,28 +342,22 @@ void resume_callback() {
...
-    uint32_t seed_size = sizeof(mars_kiss64_seed_t) / sizeof(uint32_t);
 
     // Copy the current seed
-    memcpy(&address[PARAMETER_SEED_START_POSITION], spike_source_seed,
-    		seed_size);
-
-    // Get number of spike sources
-    num_spike_sources = address[PARAMETER_SEED_START_POSITION + seed_size];
-    log_info("\t spike sources = %u", num_spike_sources);
+    parameters.spike_source_seed = spike_source_seed;
```
dkfellows: I'm not sure that this is correct; it might be better off being deleted if the seed is being stored directly in the `parameters`.

rowleya:
Fixed the merge...

alan-stokes:
python doc the public methods of spynnaker_poisson_control_connection.py please

rowleya:
I added ```@overrides``` instead as this copies the docstring from the superclass

alan-stokes:
fair enough. still reading through the other classes.

alan-stokes:
Not sure about this one. constants CONTROL_PARTITION_ID. its only used by the poisson related functionality. Shouldn't it be called POISSON_LIVE_CONTROL_PARTITION_ID

im not sure if we can use it for other stuff, or if there is other potential for using it for say basic neuraon models, or if we did, that we'd want to use the same partition id.......... i think id like explicit naming until we find a situation where we can generalise.

alan-stokes:
after that change, im happy to rubber.

rowleya:
Done.

PR #206, opened 2016-04-08
started possion heuristics

alan-stokes:
supprots splitting atoms via limited resoruces of the core for packets.
is being generised for multicast packets (both payload and none payload) fixed_route packets, sdram, cpu, dtcm.

- [x] started interface
- [ ] updated models to express their requirements 
- [ ] updated resource tracker
- [ ] updated partitioner
- [ ] tested with simple test to verify nothing else goes boom.
- [ ] tested with column to verify that it works in the hard case.
- [ ] awaiting review


alan-stokes:
covered by pull requests already merged into master.

PR #222, opened 2016-07-01
Fixes to make STDP work correctly

rowleya:


alan-stokes:
dont know what state this branch is in, nor what it depends upon, so ignoring.

PR #223, opened 2016-07-01
Fix interface

rowleya:
Fixes related to SpiNNFrontEndCommon interface changes
- [x] Changes to match up with SpiNNakerManchester/SpiNNFrontEndCommon#53


alan-stokes:
so this is now nearly complete. The only issue is that something inst quite lined up. No example runs correctly, but goes up and down the stack happily enough. This includes all the changes being considered from all the other modules.

I know its somet to do with the pynn related stuff (master pop / synaptic matrix) as the GFE examples, including conways work flawlessly. I've done quite a lot of digging, and explicitfying the code to see if that would empty the issues, and even though I found two errors, fixing them made a improvement and then went back to this strange behaviour we have now.

My next plan is to go back to master, build a partitionable version of Conways, verify that works, and try bringing it over to this new form in the hope this exposes whatever the error is.

PR #232, opened 2016-09-14
Recording fixes

alan-stokes:
fixes to the recording interface so that it no longer uses dsg regions. it now uses memory addresses directly and makes end users support recording data regions directly.

This branch still ensures the state region exists, this pull request is a temporary one, but may become the final if i give up pulling the state stuff out.

is tied to 
      1. FEC pull request 
      2. GFE pull request


rowleya:
This has been obsoleted by the recording_no_state_region branch

PR #235, opened 2016-09-21
Removal of synapse benchmark flag

alan-stokes:
removes this compile flag, as its not by default done, and it makes the provenance data item invalid.


rowleya:
I would be interested to know what the issue is with the provenance data item, as the idea here is to avoid doing the calculation unless you need to...

alan-stokes:
the provenance item is returned all the time, and so if its not compiled in, returns 0 when it definitely happened. Gary and Jamie pointed this out to me just before I was planning to leave (as Gary was looking at it for unknown reasons). Anyhow, to keep sure that the provenance item is valid, it would need to know that the binaries are compiled with this flag, and therefore omit the prov item when not set. I thought it would be easier to just always generate the data item, as it seems to just be adding to a counter, there's no if's etc where this flag is hiding code. 

If you think we should do the other work around............. ermmm............. well i guess we'd need to think how to detect this compile state somehow, because off my head I cant think of a easy way of doing so.

neworderofjamie:
To be clear we were just confused as to why the provenance item was always
0 - I think the saving in calculation is marginal having it always enabled.

rowleya:
OK, I am happy to accept that.  

The only other issue is that this seems to include changes related to the command sending i.e. that aren't to do with this request.  Can these changes be removed so this change is only for the provenance data item?

alan-stokes:
lol, teach me for doing stuff for people quickly. i did not switch branches. was in middle of my router fixes when i did this. hopefully there's only one commit covering them. 

could just define that pulll request being dependant on command sender changes.

PR #239, opened 2016-09-26
Spread neuron spikes

rowleya:
Added the fix for neuron spikes to enforce them to be spread out.  Note that this is slightly different that that done in the other similar modules as it doesn't use the SARK busy-wait procedure.  This is because the SARK procedure doesn't take account of interrupts in the wait time, which matters more in the neuron model than in the other models.


alan-stokes:
so it'll do the job, but i note it uses a busy wait. I wonder if it would be possible to somehow use a timer or a sleep...... just from a energy saving perspective. Anyhow, its more a note for the future when we focus on energy savings, rather than a real comment. rubber stamping.

PR #245, opened 2016-09-26
Updated to make pyNN-spiNNaker a script

### LINKED ISSUE #242: setup.py doesn't get executed on recent pip versions
It seems that setup.py doesn't actually get executed on some recent pip versions, meaning that pyNN-spiNNaker doesn't get installed correctly.  This should be replaced with a script that can be executed after sPyNNaker has been installed.

rowleya:
Call it as python -m spynnaker.setup_pyNN

Fixes #242 

Note that the instructions would need to be updated once this is deployed!


apdavison:
@rowleya What do you think about adding a "spiNNaker" module directly within the main PyNN distribution. This would do `from spynnaker.pyNN import *`, and fail gracefully with a nice warning message if `spynnaker` wasn't available.

This seems cleaner than patching an existing PyNN installation.

rowleya:
That sounds like a nice idea :).  One thing that is currently changing in our code is that we will soon have a sPyNNaker0.8 and a sPyNNaker0.7.  At this point, there would need to be different imports for the different versions.

Is it possible to make changes to PyNN 0.7 still?

apdavison:
Yes, PyNN 0.7 has its own Git branch, and I still release new versions from time to time, as bugs are fixed.

alan-stokes:
At the moment, the two modules are named spynnaker7 and spynnaker8 (putting
the 0. in caused issues in python paths).

rowleya:
Closing as now in separate repositories.  We still want this to go into PyNN, but probably only after we have made the next release of the software.  We can remove these links once PyNN has them directly, and we have updated our instructions.

PR #254, opened 2016-11-01
Dynamic ports

alan-stokes:
supports running multiple scripts from different folders at the same time. Basically if buffer manager stuff.


alan-stokes:
closed due to further one with more changes.

PR #266, opened 2017-01-09
Router compressor on chip

alan-stokes:
added on chip router compressor of mundy's into the main tool chain.

It includes fixes to the c code so that:
1. it works using sdram all the way though,
2.  supports different functionality in how much and when to do compression. 
3. No more RTE's but using error codes via user 0 register

includes python changes:
1. which pushes more control logic into SpiNNMan from FEC.
2. fixes pacman executor to handle optimal algorithms properly. 
3. supports coresubsets and core subset infos which are the info objects from the cores. instead of doing some translation between them 
4. supports getting the first none_monitor core from the chip.

the python polls the cores till they finish, as its a in deterministic timeframe
There are a few integration tests, including a speed test between the python and c.

 depends upon these branches;
PACMAN, FrontEndCommon, SpiNNMachine, SpinnMan, GFE, SPY

- [x] done by alan
- [ ] needs reviewing from rowley.


rowleya:
Note - tests are failing due to reliance on a different branch of SpiNNFrontEndCommon

PR #267, opened 2017-01-23
Neuron params during run updates

alan-stokes:
pull request that allows pynn users to change parameters in between runs. This includes Poisson rates, neuron parameters etc.

It required a change to sdram calcs, and some extra demands on neuron interfaces, and was requested by basab and alloc.

modules associated with this functionality are:
1. DSG
2. FEC
3. SPY
4. SPY extra models

- [x] works for normal models
- [x] works for poisson
- [x] doced
- [x] tests for normal for i_offset
- [x] tests for poisson with rate and duration
- [x] tested with basab code
- [x] reviewed by alan
- [x] reviewed by rowley


rowleya:
Note this is failing travis due to dependence on SpiNNFrontEndCommon

PR #283, opened 2017-03-07
fix to setting up machine_time_step

Christian-B:
Allowed it to come from configs

Added some checks.


apdavison:
The default time step in PyNN is 0.1 ms. This is part of the API.

It is possible to support a value `"auto"` or `None`, but the default should stay as 0.1 ms

Christian-B:
Changed back as requested by @apdavison

PR #293, opened 2017-03-23
Proposed change to include structure for C code

dkfellows:
This commit just handles the changes for input types (now selected by defining `INPUT_TYPE_CLASS` to be either `Current` or `Conductance`) but the principle should extend to our other magical include tricks too.

PR for now to get comments on approach


rowleya:
Having discussed with Donal, we are looking at alternatives

PR #301, opened 2017-04-06
New style include structure

dkfellows:
These are my suggestions for improving the building of the C code. I'm pretty sure this isn't the last word on this matter, and the downstream code (external devices, extra models) will need a bit of work to fit in.


rowleya:
We used to be able to go into the individual build folder and type "make".  With this version that doesn't seem possible - it just ends up telling me that "build/additional_input.h" is up to date.  Thus, there seems to be a problem with the default target in the build.  Can this be resolved?

PR #314, opened 2017-05-24
Add KWStyle support for C style enforcement

dkfellows:
This adds KWStyle as a C style enforcement mechanism.


dkfellows:
Alas, this hangs during the build. _Do not commit!_

PR #315, opened 2017-05-24
Enforcement of basic C style rules

dkfellows:
This branch uses the tool `vera++` to add enforcement of basic C code style rules, similar to how we do with `flake8` for the Python code.


dkfellows:
Done in a different way.

PR #317, opened 2017-05-25
supports dma recording

alan-stokes:
fixes for recording interface and simulation registration of dma's. looks like its working. Have not been able to run sudoku due to visualiser issues. will check tomorrow.

depends upon FEC https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/137

- [ ] merged FEC


alan-stokes:
having asked Andy to switch over and run soduko script (his vis works) we can confirm that these changes have no or little effect on the timer overruns from the script. So something else is slowing down the c code there.

PR #323, opened 2017-06-07
fixed issue with pynn 0-.8 and initise

alan-stokes:
fixed via checking if the parameter is a iterable. 
related to issue https://github.com/SpiNNakerManchester/sPyNNaker8/issues/13#issuecomment-306529777

### DIFF
--- a/spynnaker/pyNN/utilities/utility_calls.py
+++ b/spynnaker/pyNN/utilities/utility_calls.py
@@ -35,7 +35,7 @@ def convert_param_to_numpy(param, no_atoms):
 
     # Deal with random distributions by generating values
     if globals_variables.get_simulator().is_a_pynn_random(param):
-        
+
         # numpy reduces a single valued array to a single value, so enforce
         # that it is an array
         param_value = param.next(n=no_atoms)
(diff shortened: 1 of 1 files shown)


spynnaker/pyNN/utilities/utility_calls.py
```diff
@@ -40,7 +40,11 @@ def convert_param_to_numpy(param, no_atoms):
 
         # numpy reduces a single valued array to a single value, so enforce
         # that it is an array
-        return numpy.array([param.next(n=no_atoms)], dtype="float")
+        param_value = param.next(n=no_atoms)
+        if hasattr(param_value, '__iter__'):
+            return numpy.array(param_value, dtype="float")
+        else:
+            return numpy.array([param.next(n=no_atoms)], dtype="float")
```
rowleya: This doesn't look quite right; the "else" seems to call param.next again!  Guessing the issue is that pyNN 0.8 RandomDistribution.next returns an array every time, whereas pyNN 0.7 must return a single value when no_atoms = 1.  It might then be better to remove the initial "if no_atoms > 1" statement, otherwise this is becoming more cluttered for no reason.

alan-stokes: i ack that the param.next() in the else needs to disappear. that said, pynn 0.7 still needs to be detected, so you'd need something like this.

spynnaker/pyNN/utilities/utility_calls.py
```diff
@@ -40,7 +40,11 @@ def convert_param_to_numpy(param, no_atoms):
```
rowleya: I still believe that the no_atoms > 1 check is no longer needed, as even in PyNN 0.7, if the return value is an array, it will haveattr __iter__, thus it will be picked up in the next check.  A simple test would confirm this...

PR #326, opened 2017-06-14
Fix typo in ssp

alan-stokes:
typo fix


alan-stokes:
lol, must have been brnached off chip-power-monitor. oh well. could monkey patch it, or wait till that functionality is done. lol

rowleya:
The choice is either to remove them here and then this will be merged in, or else wait for the merge that actually includes these changes before merging this in.  I don't mind which, but I don't think it should go in as-is to the current master.

alan-stokes:
i concur. 6 of one 6 of a other, im happy to wait, it was only detected whilst i was trying to tie in buffer management for the chip monitor core. its not exactly fixing a bug, but more obaying our convention of not having magic numbers.

dkfellows:
Fixed on `master`.

PR #327, opened 2017-06-14
fixes get_weights

alan-stokes:
title. 

fixes the get weights by clearing the cache at a new run call. should work in both pynns.


spynnaker/pyNN/models/pynn_projection_common.py
```diff
@@ -267,6 +267,10 @@ def _get_synaptic_data(self, as_list, data_to_get):
         connection_holder.finish()
         return connection_holder
 
+    def _clear_cache(self):
+        post_vertex = self._projection_edge.post_vertex
+        post_vertex.clear_connection_cache()
```
rowleya: This could do with a check to ensure that the vertex actually has a method clear_connection_cache().  This could either be added to a new abstract class or else added to AbstractAcceptsIncomingSynapses.

spynnaker/pyNN/models/neuron/synaptic_manager.py
```diff
@@ -799,6 +799,9 @@ def write_data_spec(
 
         self._weight_scales[placement] = weight_scales
 
+    def clear_connection_cache(self):
+        self._retrieved_blocks = dict()
```
rowleya: Just a thought - if we know what each of the edge synapse types are, we could avoid re-reading the synaptic data when the connection doesn't use plasticity.  The "clear" here would then iterate through the retrieved blocks and only reset those where the connection isn't plastic.  The synapse info is known at the time of retrieval, so this information is available to be stored in the retrieved block dict.

alan-stokes: i imagine that structural plastic can change static synapses as well. peter would be best positioned to figure this out. if this is simple to implement, id say its a post optimisation over this fix. i wouldn't say its a blocker to this fix itself. 

i assume youd need to delete the entire block instead of just the synapse row, as then the caching mechanism is going to be much more complicated. This idea would also be safe in pynn 0.8 where the projection gives you access to all params, as the static is only weight and delay, which unless structual effects these, would be constant. 

worth noting that this may interact with post run set of a projection weight (which we will need to support at some point)

rowleya: No, this isn't a blocker, more just a thought (though if not here, it probably won't happen ;).  You are right that structural plasticity makes this more complicated, though it can equally detect this and clear if structural plasticity is in place.  Once this is merged, a ticket can be created (once the other change is done ;).

alan-stokes: peter says that static in structrual can change. and he also says his version with this merged in is still bust.......... which surprises me

rowleya: This doesn't fix any issues with plasticity itself.  We need a different test for that...

alan-stokes: hes impling that this change doesnt fix the get weights overall. i cant see how it would fail given its the blunt force right now. and testing a static syanpse is being reread when the data is identical in master....... im more inclined to believe static strucal is bust right now, but peters on the case to check

spynnaker/pyNN/models/pynn_projection_common.py
alan-stokes: fixed, and added the call to the interface as well. verified that only APV uses it, but better safe than sorry.

alan-stokes:
ready for merging me thinks

PR #329, opened 2017-06-15
Reduce warnings

Christian-B:
Only did warnings expected every single time.

### DIFF
--- a/spynnaker/pyNN/models/recording_common.py
+++ b/spynnaker/pyNN/models/recording_common.py
@@ -1,3 +1,4 @@
+from spinn_utilities import logger_utils
 from spinn_front_end_common.utilities import exceptions as fec_excceptions
 from spinn_front_end_common.utilities import globals_variables
 
@@ -97,7 +98,7 @@ class RecordingCommon(object):
                 "This population does not support the recording of gsyn exc")
         if not isinstance(
                 self._population._vertex.input_type, InputTypeConductance):
-            logger.warn(
+            logger_utils.warn_once(logger,
                 "You are trying to record the excitatory conductance from a "
                 "model which does not use conductance input.  You will "
                 "receive current measurements instead.")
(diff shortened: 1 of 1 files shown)


alan-stokes:
i disagree with this, by only sednign the message once, the end user will assume the other ones are fine.

Christian-B:
Only a very select set of warning messages currently have this implementation.

All other warnings especially does that depend on the current state remain repeated every time.

rowleya:
I would agree if the message were different each time - in this case, the message will be exactly the same giving no additional information.

alan-stokes:
but the 2 youve choosen are also on this view point. 
p1 = p.pop(IF_Cond_exp, blah)
p1.record_v() # or whatever the call is
# warn
p2 = p.pop(IZK_Cond_exp, blah)
p1.record_v() # or whatever the call is
# no warn

so as a end user, izk cond is fine. only if cond. its only after i fix the if cond would then izk appear to be broken. so the system is therefore not warning me of all my errors at the front. thats bad. 

becuase it apepars at specifc points in the code, thats the extra info.

alan-stokes:
print "im building my if cond")
p1 = p.pop(IF_Cond_exp, blah)
p1.record_v() # or whatever the call is
# warn
print ("im building my izk)
p2 = p.pop(IZK_Cond_exp, blah)
p2.record_v() # or whatever the call is
# no warn

what i see as output is now
im building my if cond
warning
im building my izk

so ive now lost the knwoeldge that izk is also not recording conducance

rowleya:
Yes, but the average user doesn't see where the message comes from - the message would be the same and printed twice.  You have purposefully added a print, but most users wouldn't.  Additionally, if they fix it for the first one, they will continue to get the warning for the second.

An alternative would be to change the message to include the model type.  This would then display once per model type.  We could also go down the C route and actually print "(this message will only be displayed once)"

alan-stokes:
could do both. that im happy with.

PR #330, opened 2017-06-16
when you say type use type(self). 

Christian-B:
Especially is self does not have _model_name!

### DIFF
--- a/spynnaker/pyNN/models/neuron/abstract_population_vertex.py
+++ b/spynnaker/pyNN/models/neuron/abstract_population_vertex.py
@@ -714,7 +714,7 @@ class AbstractPopulationVertex(
                 self._change_requires_neuron_parameters_reload = True
                 return
         raise InvalidParameterType(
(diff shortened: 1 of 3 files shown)


alan-stokes:
Id like to know what the output of type is instead of self._modelname. does it give you the overall odel name, or the class in which its executing?

Christian-B:
It gives the full type hierarchy similar to what you use in import statements.
This makes finding the broken code easy!

rowleya:
It does - type(self) is determined by the leaf in the inheritance hierarchy.  Also you need to actually *have* a self._model_name (which most code doesn't).  Note that some of the ideas for the PopulationView object might mean that you do want to look at an internal variable, but we can cross that bridge when we come to it...

alan-stokes:
6 of 1 1/2 doz of another. id argue that finding offending code is easy with the stack trace, but i doth not hath any real care either way.

PR #332, opened 2017-06-20
Recordable property required by pynn

Christian-B:


Christian-B:
population.celltype.recordable is something the pynn documentation says a users can call so that is why we want the property.
See http://neuralensemble.org/docs/PyNN/neurons.html
Section: Recording variables and retrieving recorded data
===

I played with several different options.

1. Use a chain of calls to super to build up the list.

2. Keep a list in the Recordable of the variables,
2a. Fill this list during init calls
2b. Fill this list during new calls

3. Have an Abstract static propery that is implemented several times over

4. Use hasattributes to look for getSpikes, getV ect  . IE Ducktyping

===

Rowleys idea should also work. I will have a play with that,

Clearly the method and property names have to be as defined by pynn otherwise this whole thing is not very useful.

Christian-B:
Did two different versions using mro().

Which Alan does not like as he does not know mro()

Clearly when we pick one implementation the others have to be removed.

spynnaker/pyNN/models/common/recordable.py
```diff
@@ -0,0 +1,120 @@
...
+class Recordable(object):
+    """ Indicates that spikes can be recorded from this object
+    """
+
+    __slots__ = ()
+
+    # Here we have to choose between using @abstractstatic or @staticmethod
+    # Using both either fails or cause @abstractstatic to be ignored
+    # However even abstractstatic only enforces that one of the
+    # abstract_x_recordable interfaces implements the method not all.
```
rowleya: This is definitely an issue - I don't think python provides any way around this (Java doesn't know how to do this either ;) )

spynnaker/pyNN/models/common/recordable.py
```diff
@@ -0,0 +1,120 @@
...
+        if not isinstance(aclass, type):
+            aclass = type(aclass)
+        variables = list()
+        for sub in aclass.mro():
+            if sub in Recordable.__subclasses__():
+                variables.append(sub.get_recordable_variable())
+        return variables
+
+    @staticmethod
+    def get_all_possible_recordable_variables1(aclass):
```
rowleya: I think this one is the most correct, as it uses issubclass.

Christian-B:
An alternative approach may be to push the list of recordable variables higher up the stack.
Then instead of using the existance of the various abstract_XYZ_recordable interfaces to do all data recording (except Spikes) with a more general approach.

rowleya:
I think this is how it was before these changes...

Christian-B:
Using a single class for all but spike data type is the better solution.

This stuff is dead but branch left temporarily for reference.

PR #342, opened 2017-06-30
Abstract slots

Christian-B:
Must be done at same time as:
https://github.com/SpiNNakerManchester/sPyNNaker7/pull/13
https://github.com/SpiNNakerManchester/sPyNNaker8/pull/28


Christian-B:
Avoids Two classes with the same name.

Christian-B:
Nice idea but breaks too much to be worth it.

PR #343, opened 2017-07-03
Mantas/neuromodulation

dkfellows:
This is the rescue of Mantas's neuromodulation code from total unmerge-ability. I suspect it ought to actually go in the extra models plugin if it is going to be retained at all.


mmikaitis:
Hi Donald, Andrew. I just spotted this by accident - github did not notify me of this change. So I plan to continue this work soon and this will be quite a useful featuture for learning rules, introducing what is called third-factor learning rules for SpiNNaker (E.g. Frémaux et al 2016). Should I move it to extra plugins, i.e. what is the protocol here for these kind of things? Many thanks.

andrewgait:
I'm sure other people will have their opinions, but the current plan is to move ExtraModels into sPyNNaker in the near future.  So my suggestion would be to keep your work in this repository.

rowleya:
I agree - leave this in sPyNNaker for now.

mmikaitis:
Hi. I have recovered my original branch neuromodulation. So this branch can be removed shortly.

alan-stokes:
hi mantas, whats the plan with this?

mmikaitis:
_This_ branch can be deleted. I have another branch for this work (Called just "neuromodulation") and the work itself is far from ready to be merged to master.

Thanks Alan

alan-stokes:
branch and pull request?

mmikaitis:
Yes, I will create a new one from the correct branch.

Cheers

PR #344, opened 2017-07-03
Neuron implementation work

andrewgait:
Work to allow end users to modify neuron behaviour (specifically input_type, where inhibitory input could depend on excitatory input) more easily.  This all came from an end user's original model of a spiking Elementary Motion Detector (sEMD); the changes mean they can implement this through sPyNNaker7/8NewModelTemplate rather than having to edit neuron.c in sPyNNaker.

Examples have been added on the same branch name to PyNN7/8Examples; repositories that depend on this one will need this pull request alongside in order to work.


rowleya:
Understood - the other option of moving it to extra models and then back again felt like more work :).

andrewgait:
I agree with that.  I think I was planning to do this slightly differently in that I was considering having a models/ and an extra_models/ directory in sPyNNaker/pyNN, but this way works too.  I thought there was also a question of pyNN vs. non-pyNN models that needed to be addressed, but perhaps I'm wrong...

rowleya:
I thought that the best way to handle pyNN vs. non-PyNN was to just do this at the python level.  I don't think the C code needs to worry about it, as that is just about building the binaries, which end up in the same place anyway.

In python, there are now extra_models and external_devices modules, which just import the appropriate bits.

rowleya:
I have now split this out again - sorry for the confusion!

andrewgait:
That sounds like a good plan to me I think - I'll have a look at this sometime in the next week, and also move out the sEMD implementation into extra_models while in there.

andrewgait:
(Build errors are "bare excepts" from new version of flake8, I think these have been done already in a different pull request... ?)

dkfellows:
Master had the fixes for those. Needs a merge anyway, as there are now some conflicts too (probably minor, but I've not looked).

oliverrhodes:
Just for the record: I've talked this branch through with @andrewgait , and don't see any issues merging it.

The comments I had were:

- input_type_set_inhibitroy_multiplicator_value function might not be necessary, as could be done via the input_type_convert_inhibitory_input_to_current function, which would save having to add this to input_type.h

- The synapse implementation is similar to the neuron implementation, however it doesn't capture the functionality from synapses.c in the same way the neuron implementation does from neuron.c. For example, the synapse implementation splits out the synapse_types_add_neuron_input, but leaves synapse_types_shape_input in the synapse_type. It might be better to have everything in one place of the other.

- The changes to synaptic_manager regarding weight_scale and input_type won't change functionality, however we can hopefully simplify things here when I've finished the precision based weight scaling, as this may remove the need to have input_type based scaling.

- It's good to see this doesn't really affect code size, with IF_curr_exp going from: 24960 (master) to 24980 (this branch) on my mac with version 5.4 of the compiler

andrewgait:
I'm still wondering whether this goes far enough towards what we want, given the recent discussions between @rowleya and @dr-david-lester - the neurons and synapses are still relatively separate in this implementation as it stands at the moment. I suppose it is possible to work out some way to join them together even further, but at the moment it's not at that stage; basically, it is split into a "neuron_impl" and a "synapse_impl". 

I would suggest too, however, that doing this would take us away to some extent from the future idea we had of building on this implementation as it stands in order to implement something like Jamie's model, whereby the synapses and neurons are on different cores (and the amount of cores dedicated to each depends on how many synapses per neuron there are, to some extent).

(Of course, we could still make it possible to implement models in either way... !)

dr-david-lester:
Andrew(s),

The separation required by the Jamie model is between accumulation of ring-buffer-updates and other stuff.

In particular, the current-shaping should occur in the neural update, which uses (or used) "current buffers" (not sure what they're called these days).

Once upon a time, the synapse rows DMA-ed into DTCM was added to ring-buffers.

Then at the correct clock-tick, those elements now at the front were transferred to current buffers (and the old ring-buffer elements were zero-ed for reuse).

If I recall correctly, we also shaped the currents at this point.

What I want to happen was the current shaping was moved into the neuron update phase.

Does that help?

Dave
________________________________

neural_modelling/src/neuron/implementations/neuron_impl_semd.h
```diff
@@ -0,0 +1,219 @@
...
+#include <recording.h>
+#include <debug.h>
+#include <string.h>
+
+#define V_RECORDING_INDEX 0
+#define GSYN_EXCITATORY_RECORDING_INDEX 1
+#define GSYN_INHIBITORY_RECORDING_INDEX 2
+
+typedef struct input_type_current_semd_t {
+
```
andrewgait: I think it's fine to do this here, but it seems a bit strange to do it here rather than actually in the input_type directory.  I understand, though, that it's a bit weird to add functions to the standard input_type that will never be used by it... so I'm happy with this the way it is - just thought I would mention this here in case any more thought is required.

neural_modelling/src/neuron/builds/IF_cond_exp/Makefile
```diff
@@ -3,6 +3,8 @@ BUILD_DIR = build/
 NEURON_MODEL = $(SOURCE_DIR)/neuron/models/neuron_model_lif_impl.c
 NEURON_MODEL_H = $(SOURCE_DIR)/neuron/models/neuron_model_lif_impl.h
 INPUT_TYPE_H = $(SOURCE_DIR)/neuron/input_types/input_type_conductance.h
+NEURON_IMPL_H = $(SOURCE_DIR)/neuron/implementations/neuron_impl_standard.h
+SYNAPSE_IMPL_H = $(SOURCE_DIR)/neuron/implementations/synapse_impl_standard.h
 THRESHOLD_TYPE_H = $(SOURCE_DIR)/neuron/threshold_types/threshold_type_static.h
```
andrewgait: synapse_impl.h / synapse_impl_standard.h no longer exist in the current version (after the add_input function was moved to the neuron implementation) so this and all subsequent references in further Makefiles - I'm not going to mark all of them! - can be removed.

rowleya: Fixed

neural_modelling/src/neuron/implementations/neuron_impl_semd.h
rowleya: The SEMD model is a little bit odd certainly, in that it almost, but doesn't quite, fit in to the standard neuron model.  As you say though, we can't really add a function each time someone wants something different from the interface, as this would be hard to keep up with!

The SEMD implementation ended up being custom at the C level, but actually standard at the Python level.  As I say, this is a bit odd but it does work!

sPyNNaker8NewModelTemplate now contains an example where this is not the case.  Hopefully that helps to show what a completely non-standard model looks like...

spynnaker/pyNN/models/neuron/synapse_types/synapse_type_exponential.py
```diff
@@ -63,8 +63,9 @@ def has_variable(self, variable):
     @overrides(AbstractSynapseType.get_values, additional_arguments={'ts'})
     def get_values(self, parameters, state_variables, vertex_slice, ts):
 
-        decay = lambda x: numpy.exp(-ts / x)  # noqa E731
-        init = lambda x: (x / ts) * (1.0 - numpy.exp(-ts / x))  # noqa E731
+        tsfloat = float(ts) / 1000.0
```
rowleya: It is probably worth checking if there is anywhere else in the changes that forgets to do this!

andrewgait: I found one more in dual_exponential; I think there might be another in NewModelTemplate which I'll get to at some point...

rowleya:
I believe that I have fixed external devices now as well - at least they worked in my tests.  I might want to plug in a pushbot to be sure, but I don't think there are any major issues now

PR #352, opened 2017-07-19
One to one matrix limit

rowleya:
An old branch that adds the ability to limit the amount of DTCM used by the direct matrix, which can otherwise grow large, if not impossibly large.


oliverrhodes:
Thanks Andrew, I'll test this out and merge it once I'm back.

Issue #260, opened 2016-11-29
Division by wrong value in fixed-to-float conversions


rowleya:
Some of the code for converting values from fixed-point S1615 to floating-point uses an incorrect value of 32767.0 instead of 32768.0 (namely gsyn and v).

Christian-B:
Duplicate of https://github.com/SpiNNakerManchester/DataSpecification/issues/10

rowleya:
This is a bit different - it should be checked though I think this is fixed

alan-stokes:
fixed in master

PR #362, opened 2017-08-01
Divide by correct value

andrewgait:
- should be 32768.0, not 32767.0

Fix of https://github.com/SpiNNakerManchester/sPyNNaker/issues/260


alan-stokes:
there's a possible better fix here. which is to use the data specification data type to have this conversion value. this makes the magic number in one place, instead of spread out around the tools.

PR #363, opened 2017-08-01
Config load

Christian-B:
Need to review ASAP or revert SpinFrontEndcommon


Christian-B:
After discussions not a good idea.

PR #366, opened 2017-08-03
Neuromodulation

mmikaitis:
Izhikevich's neuromodulated-STDP learning rule.

Tasks:
- [x] Replicate experiment results from Izhikevich paper.
- [x] Make this rule separate from classical STDP (Enable or disable by the user at PyNN level and choose appropriate models at Python and C levels).

(Not necessary for this pull request) Allow options for various weight dependence rules (Rename to eligibility trace dependence rules). This includes eligibility trace scaling and clamping, similarly to how weights are processed in current version STDP.

Merge with https://github.com/SpiNNakerManchester/PyNN8Examples/pull/24

Tested by https://github.com/SpiNNakerManchester/IntegrationTests/pull/8


mmikaitis:
- [ ] Replicate experiment results from Izhikevich paper.
- [ ] Make this rule separate from classical STDP (Enable or disable by the user at PyNN level).
- [ ] Allow options for various timing and weight dependece rules.

mmikaitis:
The major change for plasticity when neuromodulation rule is enabled, is that all weight dependence rules are now modifying not weights but eligibility traces (Per synapse variable). So weight dependence rules become eligibility trace dependence rules and weight is modified with a single fixed rule somewhere else (Most likely called from timing dependence).

neural_modelling/src/neuron/plasticity/stdp/synapse_structure/synapse_structure_weight_impl.h
```diff
@@ -4,8 +4,11 @@
 //---------------------------------------
 // Structures
 //---------------------------------------
-// Plastic synapse types are just weights;
-typedef weight_t plastic_synapse_t;
+// Plastic synapse types have weights and eligibility traces
+typedef struct Plastic_synapses {
+    int16_t eligibility_trace;
+    weight_t weight;
+} plastic_synapse_t;
```
neworderofjamie: The performance of accessing these will probably be massively improved if you switch from structs to ``int16_t`` in the same way I did in this change https://github.com/SpiNNakerManchester/sPyNNaker/commit/97071e4d2b99a1ad17ba913c7de039a9967d1d43

neural_modelling/src/neuron/plasticity/synapse_dynamics.h
```diff
@@ -12,6 +12,9 @@ bool synapse_dynamics_process_plastic_synapses(
     address_t plastic_region_address, address_t fixed_region_address,
     weight_t *ring_buffers, uint32_t time);
 
+void synapse_dynamics_process_neuromodulator_event(
```
neworderofjamie: Never pass non 32-bit data types to functions in ARM code - it results in loads of unnecessary bit-masking operations

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_neuromodulated_impl.c
```diff
@@ -0,0 +1,455 @@
...
+uint32_t weight_update_constant_component;
+uint32_t last_dopamine_spike_time = 0;
+weight_state_t weight_state;
+
+
+//---------------------------------------
+// Dopamine trace is a simple decaying trace similarly implemented as pre and
+// post trace.
+static inline post_trace_t add_dopamine_spike(
+        uint32_t time, int16_t concentration, int16_t* last_dopamine_level) {
```
neworderofjamie: Again, passing 16-bit integers to functions is inefficient - C will believe you know what you're doing and mask the results of all the 32-bit ARM operations down to 16-bit

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_neuromodulated_impl.c
```diff
@@ -0,0 +1,455 @@
...
+        post_event_history_t *post_event_history) {
+
+    use(&trace);
+
+    // Calculate EXP components in JK's weight update equation
+    int16_t decay_eligibility_trace = DECAY_LOOKUP_TAU_C(
+        time - last_non_dopamine_spike_time);
+    int16_t decay_dopamine_trace = DECAY_LOOKUP_TAU_D(
+        time - post_event_history -> last_dopamine_spike_time);
+    int16_t third_exp_component;
```
neworderofjamie: You also shouldn't use 16-bit integers to STORE stuff that's going to live in registers or on the stack - ARM registers are 32-bit AND it results in bit masking operations whenever you access them

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_neuromodulated_impl.c
```diff
@@ -0,0 +1,455 @@
...
+
+    use(&trace);
+    use(&last_post_trace);
+
+    // Calculate EXP components in JK's weight update equation
+    int16_t decay_eligibility_trace = DECAY_LOOKUP_TAU_C(
+        time - last_non_dopamine_spike_time);
+    int16_t decay_dopamine_trace = DECAY_LOOKUP_TAU_D(
+        time - post_event_history -> last_dopamine_spike_time);
+    int16_t third_exp_component;
```
neworderofjamie: Again you shouldn't use 16-bit integers to STORE stuff that's going to live in registers or on the stack - ARM registers are 32-bit AND it results in bit masking operations whenever you access them

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_neuromodulated_impl.c
```diff
@@ -0,0 +1,455 @@
...
+    const post_trace_t last_post_trace =
+        history->traces[history->count_minus_one];
+    post_events_add(time, history, timing_add_post_spike(time, last_post_time,
+                                                         last_post_trace));
+}
+
+//--------------------------------------
+void synapse_dynamics_process_neuromodulator_event(
+        uint32_t time, int16_t concentration, uint32_t neuron_index) {
+    log_debug("Adding neuromodulation event to trace at time:%u concentration: %d", time, concentration);
```
neworderofjamie: 16-bit parameters baaad

mmikaitis:
@rowleya I have a brief idea on how to integrate this but some parts are missing, maybe you have some ideas?

1. New synapse dynamics are introduced to accomodate dopaminergic spikes. These dynamics have special functions to add and process dopamine spikes and they call functions in timing dependence rule to process usual pre- and post- spikes.

2. Timing dependence rules should be left intact, however, they do not modify weights but they modify eligibility traces.

3. Eligibility trace has all the parameters that weight has, so weight  dependence can become eligibility trace dependence. Additionally eligibility trace has timing constant because it is decaying.

4.Weights are updated simply by multiplying two traces on each spike pair processing call. However weights should still have A_plus/A_minus limits but do not require sophisticated weight dependence rules - I think.

So I am not sure how to tie this all up so that if user specified neuromodulation=True in PyNN script all the points that I have listed are wired up in the selected binary properly. Thoughts?

mmikaitis:
@rowleya I think I have a pretty good first version of this learning rule - please review / pass to someone else responsible! Cheers.

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_izhikevich_neuromodulation.c
```diff
@@ -0,0 +1,483 @@
...
+        "Adding neuromodulation event to trace at time:%u concentration:%d",
+        time, concentration);
+
+    // Get post event history of this neuron
+    post_event_history_t *history = &post_event_history[neuron_index];
+    const uint32_t last_post_time = history->times[history->count_minus_one];
+    const post_trace_t last_post_trace =
+        history->traces[history->count_minus_one];
+
+    // Update neuromodulator level reaching this post synaptic neuron
```
mmikaitis: @oliverrhodes note that this comment is misleading - we need to know not just dopamine level at the post-synaptic neuron, but a history of dopamine spikes in order to be able to integrate them (see post_events_with_da.h). I will update the comments.

dkfellows:
The code does not currently seem to be in a state which builds. I've brought it up to date with some of the recent changes we've been doing elsewhere, but the problems seem to date back a few versions.

You can explore the testing builds on travis above; an example build-failing error is this:
```
make[1]: *** No rule to make target `builds/IF_curr_exp_stdp_izhikevich_neuromodiulation', needed by `all'.  Stop.
```
That looks thoroughly fatal to me…

mmikaitis:
Fixed a typo. Now it can run the build but there are errors due to synapse_dynamics using functions in synapse_row.h incorrectly. As discussed, we will need to think how to integrate this properly without introducing new synapse dynamics like it is now. :)

mmikaitis:
@oliverrhodes is this PR in your agenda somewhere? If you have full hands I am actually happy to take this again and try put it into some mergeable form. :)

I believe @rowleya was looking at this also at some point (based on spinnaker mailing list info). Maybe we three need to try to schedule another meeting about this learning rule.

bstroh:
Hi, I am using the SpiNNaker board for development at my University. I would like to use neuromodulation in my PhD project and I was wondering what the status of this branch is, will it be merged? :)

rowleya:
This is a bit out of date, but I think we still intend merging it eventually.  I am unsure when this will be at this point though...

dkfellows:
See also SpiNNakerManchester/sPyNNaker8#229 and SpiNNakerManchester/PyNN8Examples#24

dkfellows:
Code is fully clean. Only sensible remaining question is:

- [ ] Do we want to add this feature?

mmikaitis:
> I think this is now at the point where it needs to be (re)considered for merging. There are still outstanding issues which I guess can be debated:
> 
> 1. This doesn't go the whole way into integrating with our STDP methods, mainly because @mmikaitis has chosen to use int32s where we tend to use int16s (and call them weight_t).  I've stopped going the whole way along this route because I wasn't sure whether I would lose too much accuracy by doing so.
> 2. A related issue is that this setup will only work with the multiplicative weight rule used in the examples (see [SpiNNakerManchester/PyNN8Examples#24](https://github.com/SpiNNakerManchester/PyNN8Examples/pull/24)).  It would be a bit more work on top of what is here to add in the capability to use an additive weight dependence, for example.  [At the moment if anyone tries to do so it will tell them it's not implemented].
> 3. This does work (as far as I can tell from my examples and testing - also in the above repository) alongside structural plasticity, at least in the sense that it returns sensible-looking results.
> 
> Mantas: do you have any further comments/suggestions?

Hi Andrew,

Nice to see that you were looking into this old code!

No further comments, just that regarding point 1, int32 was used because apart from the weight, each plastic synapse also has an eligibility trace. So this is not about the accuracy (weights are still 16-bit) but that we needed extra 16 bits per synapse for those traces.

Best regards,
Mantas

andrewgait:
Just noting that this is marked as "incomplete" for now as it's not planned for the next release, but the next step on it is probably a review from someone (@rowleya ?).

spynnaker/pyNN/utilities/bit_field_utilities.py
```diff
@@ -203,9 +222,28 @@ def write_bitfield_init_data(
...
-                    machine_edges.append(machine_edge)
+
+            # If neuromodulation is turned on on this edge then treat
+            # edges with reward or punishment receptors differently
+            post_vertex = in_edge.post_vertex
+            neuromodulation = in_edge.is_neuromodulated(post_vertex)
+
+            receptor_type = in_edge.synapse_information[0].receptor_type
+            if neuromodulation and (receptor_type == "reward" or
+                                    receptor_type == "punishment"):
```
rowleya, line 233: This should ask the synapse type whether the receptor type is neuromodulated

rowleya: This should probably also go through the synapse information rather than rely on the first as two projections with different parameters could mess this up!

spynnaker/pyNN/extra_algorithms/splitter_components/splitter_abstract_pop_vertex_neurons_synapses.py
```diff
@@ -564,10 +569,30 @@ def get_in_coming_vertices(
...
+        # Work out if there is a neuromodulated edge
+        app_vertex = self._governed_app_vertex
+        neuromodulation = edge.is_neuromodulated(app_vertex)
+
+        # If the incoming edge targets the reward or punishment receptors
+        # then it needs to be treated differently
+        receptor_type = edge.synapse_information[0].receptor_type
+        if neuromodulation and (receptor_type == "reward" or
+                                receptor_type == "punishment"):
+            # Increment the vertex index so non-neuromodulated edges work
```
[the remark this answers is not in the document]
rowleya: (Should probably also look through the synapse information in case there are different receptor types)

spynnaker/pyNN/models/neuron/synaptic_matrices.py
```diff
@@ -355,27 +356,49 @@ def __in_edges_by_app_edge(self, incoming_projections, routing_info):
...
             app_edge = proj._projection_edge
 
             # Skip if already done
             if app_edge in in_edges_by_app_edge:
                 continue
 
+            # Is there a neuromodulated edge?
+            post_vertex = app_edge.post_vertex
+            neuromodulation = app_edge.is_neuromodulated(post_vertex)
+            receptor_type = app_edge.synapse_information[0].receptor_type
```
rowleya, line 370: Might want to iterate over the synapse information and get all receptor types

rowleya (dismissed):
Looking close, but there are a couple of things to check; the change to synapses.c could be breaking for some things!

spynnaker/pyNN/models/neural_projections/projection_application_edge.py
```diff
@@ -109,6 +124,19 @@ def delay_edge(self):
...
+    def is_neuromodulated(self, post_vertex):
+        """ Check if neuromodulation is set on any edge into the post_vertex
+
+        :rtype: bool
+        """
+        if self.__is_neuromodulated is None:
+            for proj in post_vertex.incoming_projections:
+                dynamics = proj._synapse_information.synapse_dynamics
+                if are_dynamics_stdp(dynamics):
+                    self.__is_neuromodulated = dynamics.neuromodulation
```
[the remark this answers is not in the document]
andrewgait: There aren't currently any examples with a neuromodulated and non-neuromodulated STDP dynamics on the same edge.  I suppose there may be future cases where this is the case though, so you're probably right that this needs to stay true in a mixed case.

neural_modelling/src/neuron/synapses.c
```diff
@@ -202,29 +202,48 @@ static inline bool process_fixed_synapses(
...
+            uint32_t index = synapse_row_sparse_index(
+                synaptic_word, synapse_index_mask);
+            // In case this is punishment synapse, invert dopamine level
+            // to cause depression.
+            if (synapse_type == 3) {
+                concentration = ~concentration + 1;
+            }
+            synapse_dynamics_process_neuromodulator_event(time,
+                concentration, index, synapse_type);
+        } else {
```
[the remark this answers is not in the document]
andrewgait: I guess again here that there isn't an example using a different synapse type where the ID could be > 1 (though I think the dual-excitatory synapse is the only current one we have?) along with neuromodulation, so it wouldn't have been noticed yet.  I agree though that this could be moved into an impl file somehow.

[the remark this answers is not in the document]
rowleya: I would think this would already break the dual-excitatory synapse... it might be that we don't have a test for this that actually checks the result (it wouldn't fail to run, but it would likely break the results).  I think EProp might also make use of the multiple synapse types though I am not certain...

spynnaker/pyNN/models/neural_projections/projection_application_edge.py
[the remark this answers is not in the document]
andrewgait: In fact, given some of the sensible comments further on in this PR, it would probably be a good idea to write some tests where there are mixed dynamics objects on the same edge in order to make sure nothing is broken in those cases...

mmikaitis:
Hi all. Great to see this has been merged! Many thanks.

andrewgait:
> Hi all. Great to see this has been merged! Many thanks.

It's gone through quite a few iterations since but thank you for coming up with the prototype in the first place!

Issue #361, opened 2017-08-01
Too long random back-off in SpikeSourcePoisson


chanokin:
In SpikeSourcePoisson the back off could end up beeing too long if the population is big:

https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/spynnaker/pyNN/models/spike_source/spike_source_poisson.py#L401

could be fixed with:

`
spec.write_value(random.randint(0, min(SpikeSourcePoisson._n_poisson_vertices,  MICROSECONDS_PER_SECOND//machine_time_step)))
`

andrewgait:
Fixed in https://github.com/SpiNNakerManchester/sPyNNaker/pull/367

alan-stokes:
closed!

Issue #271, opened 2017-02-03
Low rate Poisson source causes an error


rowleya:
Referring to the code here:
https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/spynnaker/pyNN/models/spike_source/spike_source_poisson.py#L317

When the rate is very low, the expected spikes per timestep is very low.  This can cause one of two problems:
1) When spikes_per_timestep is 0, a division by 0 is performed.
2) When spikes_per_timestep is close to 0, time_between_spikes is very large and doesn't fit in a UINT32.

Since this code is only there to stop multiple spikes being sent all at once, the simple solution is to make time_between_spikes = 0 when spikes_per_timestep is <= 1, only performing the division when spikes_per_timestep > 1.

andrewgait:
Added suggested code in https://github.com/SpiNNakerManchester/sPyNNaker/pull/367

Is there an actual low Poisson rate example out there somewhere that this can be tested against further?

alan-stokes:
closed

PR #367, opened 2017-08-03
Add low rate value check

andrewgait:
- followed instructions by AR to fix issue when spikes_per_timestep is
zero or small.

Followed instructions in issue: https://github.com/SpiNNakerManchester/sPyNNaker/issues/271

Tested against all examples I can find that use SpikeSourcePoisson, no problems.  Not sure whether we have any example that uses a low Poisson rate as described in the issue - couldn't find any.

### DIFF
--- a/spynnaker/pyNN/models/spike_source/spike_source_poisson.py
+++ b/spynnaker/pyNN/models/spike_source/spike_source_poisson.py
@@ -398,7 +398,9 @@ class SpikeSourcePoisson(
             spec.write_value(data=key)
 
         # Write the random back off value
-        spec.write_value(random.randint(0, self._n_poisson_machine_vertices))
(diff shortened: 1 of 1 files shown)


alan-stokes:
if ya cant find one, why not build a test to verify it?

andrewgait:
Will see what I can do later this afternoon...

andrewgait:
Made another edit to fix https://github.com/SpiNNakerManchester/sPyNNaker/issues/361

Also, discovered that one of the tests in https://github.com/SpiNNakerManchester/sPyNNaker7/blob/master/p7_integration_tests/test_spike_source/test_poisson_spike_source.py (and the equivalent in sPyNNaker8) has a low enough rate that this change affects it, so there's no need to write another one.

PR #369, opened 2017-08-04
matplotlib subplot index starts at 1, not 0

andrewgait:
- noticed this from error thrown at the end of integration_test in
sPyNNaker7/p7_integration_tests/test_various
test_synfire_izk_curr_exp.py when attempting to plot using subplot


alan-stokes:
another reason why python is a pain. who would have thunk to think from 1 instead of zero. lol

apdavison:
(The 1-indexing is to maximize compatibility with MATLAB)

Issue #244, opened 2016-09-26
Memory allocation is not working correctly


rowleya:
Sometimes a script with enough neurons and enough run time will over-allocate memory.  This means that everything seems fine until the Executing Data Specifications and Loading phase at which point the core reports that it can't allocate memory.

alan-stokes:
i think you mean under allocate memory. a over allocation would be fine and we most likely always do that. It would be interesting to get a small example that does this. because its highly likely that the synaptic matrix estimates must be failing at this point.

rowleya:
It over allocates the SDRAM as in it allocates too much.  It doesn't under-allocate though it might under-estimate.  Note that there is no check on this; DSG in FEC could do a check though and fail if too much SDRAM is allocated.

alan-stokes:
would be much better to fail in dsg. dse is a nightmare to debug. it underallocs sdram region in dsg?/ sdram estimate for partitioning?, and writes too much to it during dsg.

rowleya:
If it writes too much to a region, DSG might or might not fail.  This is because you don't actually know if it has written too much, as pointers can move and you can add for loops etc.  There could probably be a basic check for write and write_array which is aware of set_write_pointer - it can then keep track of the furthest known write pointer and if this is outside of the region, it can raise an error (see SpiNNakerManchester/DataSpecification#8).

An additional check can be done in FEC where after the DSG is called, the size of each region is added to the total size for the chip on which the data is to be loaded.  This can then raise an exception if that overflows (see SpiNNakerManchester/SpiNNFrontEndCommon#72).

This issue is still needed to fix the actual overflow though.

alan-stokes:
think this is solved from the pull request about dsg memory errors. If this is the case, its likely this guy can be closed.

alan-stokes:
closing assuming fixed. reopen if proved otherwise

PR #373, opened 2017-08-15
fix sdram overuse

rowleya:
Added a multiplication to avoid too much SDRAM being used in very large networks (comes up with the Microcircuit example)

### DIFF
--- a/spynnaker/pyNN/models/neuron/synaptic_manager.py
+++ b/spynnaker/pyNN/models/neuron/synaptic_manager.py
@@ -44,6 +44,9 @@ _SYNAPSES_BASE_SDRAM_USAGE_IN_BYTES = 0
 _SYNAPSES_BASE_N_CPU_CYCLES_PER_NEURON = 10
 _SYNAPSES_BASE_N_CPU_CYCLES = 8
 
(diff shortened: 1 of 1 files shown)


rowleya:
Fixes #244

alan-stokes:
magic number please name

alan-stokes:
isn't this to avoid too little sdram being used? not too much sdram (as then id expect a divide) and it smacks of a fudge. Is there a better solution?

rowleya:
A fudge it is, but the memory use is statistical.  This certainly allows us to run the microcircuit, which we can't without...

rowleya:
It avoids us over-using the SDRAM in the synapses - if we do this on all cores on a chip we get an "out-of-memory" error when allocating recording regions (as these are done by the core now).

alan-stokes:
ok, used vs allocated logic viewpoint. fair cope. and i know, just find fudges to be things that bite us later. lol.

rowleya:
This is unlikely to cause new issues as it is just adding more memory to the estimate - so you might get less neurons per core with a large network I guess.  We might get more similar issues later that mean we need to revise this upwards again

Issue #370, opened 2017-08-08
"Exception: Only rows of up to 255 entries are allowed" - What does it mean to a general PyNN user?


mmikaitis:
Exception occurs in the following small test cases:

* Population 1 with 1 neuron; Population 2 with 127 neurons; All-to-All connection; STDP.
* Population 1 with 100 neurons; Population 2 with 127 neurons; All-to-All connection; STDP.
* Population 1 with 1 neuron; Population 2 with 256 neurons; All-to-All connection; **No STDP**.

Exception **does not** occur in the following test cases:

* Population 1 with 1 neuron; Population 2 with 126 neurons; All-to-All connection; STDP.
* Population 1 with 300 neurons; Population 2 with 126 neurons; All-to-All connection; STDP.
* Population 1 with 1 neuron; Population 2 with 255 neurons; All-to-All connection; **No STDP**.
* Population 1 with 127 neurons; Population 2 with 127 neurons; One-To-One connection; STDP.
* Population 1 with 127 neurons; Population 2 with 400 neurons; One-To-One connection; STDP.

So it looks like this error means that Population 2 cannot be larger than 126 neurons if there is at least one neuron in Population 1 that will connect to all of the neurons in Population 2 through plastic synapses? I.e. **No neuron can connect to more than 126 neurons when STDP is enabled**?

How does 126 or 127 relate to 255? 126x2=252 and 127x2=254. What is meant by "entries" and "rows" in terms of PyNN vocabulary. How does STDP affect these numbers.

Error needs clarification on what needs to be fixed in user's PyNN script.

Thanks, - Mantas :)

mmikaitis:
It would be worth also considering 3 population test cases. I.e. Population 1 with 1 neuron, Population 2 with 126 neurons, Population 3 with 126 neurons. Two All-To-All projections 1 -> 2 and 1 -> 3. Will this cause the exception?

rowleya:
I believe that this is fixed by #374.  

I didn't get a match for the numbers you give in terms of limits, but for STDP it will depend on the rules as to the maximum number of synapses per row.  When not using STDP, I couldn't replicate this.  If you have a script that does so, please attach it!

mmikaitis:
Ahh I see, I saw these errors with neuromodulation rule which probably requires more memory per row. So the only time this exception appears using simple STDP rule, pair based, is when post-synaptic population is larger than 255 neurons (Simple test attached).

[...]

PR #374, opened 2017-08-16
Updated error message when the synapse row gets too big

rowleya:
This should make it easier for the user to update their script and make it work...


rowleya:
Fixes #370

alan-stokes:
bounced back as looks like black magic, better would be to ask the dynamics how many synapses it can fit into x bytes. But this requires more changes which wont be ready for release. so bounced till future release.

rowleya:
This has been updated as requested - please look again!

PR #375, opened 2017-08-17
Fix read direct matrix

rowleya:
Fixed an issue where the direct matrix was not being read back correctly.  A test has been added.


rowleya:
Fixes SpiNNakerManchester/sPyNNaker7#23

PR #378, opened 2017-08-22
Add first_id and last_id to the population

rowleya:
These are apparently things that PyNN populations should have


alan-stokes:
bounced back as the pynn 0.8 already have first_id and last_id that are private variables. I do note that they are not properties. These are used by describe, id_to_index etc. So with these proposed changes, youd have two first_ids etc which wont sync up. More thought needed here.

alan-stokes:
Its also worth noting that PyNN 0.7 impl we have has no idea about first and last id, and the describe just sets them to None. Our PyNN 0.75 also doesn't give the id_to_index(self, id) function at all.

spynnaker/pyNN/models/pynn_population_common.py
```diff
@@ -19,6 +19,10 @@
 
 
 class PyNNPopulationCommon(object):
+
+    # Keep a global neuron id for print headers
+    _next_global_neuron_id = 0
```
dkfellows: This appears to be unreferenced elsewhere.

rowleya: Oops - should probably go now...

PR #382, opened 2017-08-23
Fix the read back of parameters when there are globals

### LINKED ISSUE #380: Problem with injection context when using set after run with IZK_cond_exp
There seems to be an issue when using e.g. pop.set('i_offset', 1) when pop uses the IZK_cond_exp model.

rowleya:
Fixes #380


alan-stokes:
can i propose a constant for that 4? word_to_bytes_converter?

I've been using such a named constant in the codebase for speeding up data extraction, so putting it in here will hopefully clear up our code a little.

alan-stokes:
please fix 2 ticks.

alan-stokes:
also, could I propose that it goes into FEC constants, instead of being a private variable? I'll bet that that magic number is used shed loads around the place.

rowleya:
In this case it is _BYTES_PER_PARAMETER, which, whilst it is 4, is a specific assumption that we use here

alan-stokes:
fair cope.

Issue #402, opened 2017-10-05
DistanceDependentProbabilityConnector cannot be instantiated 


uahic:
While instantiation of the DistanceDependentProbabilityConnector the following error occurs

```
/home/schulze/.local/lib/python2.7/site-packages/spynnaker/pyNN/models/neural_projections/connectors/distance_dependent_probability_connector.pyc in __init__(self, d_expression, allow_self_connections, safe, verbose, n_connections)
     56         # TODO: Work out how this can be done statistically
     57         expand_distances = self._expand_distances(self._d_expression)
---> 58         pre_positions = self._pre_population.positions
     59         post_positions = self._post_population.positions
     60 

AttributeError: 'NoneType' object has no attribute 'positions'
```

the file causing this issue is located at
<your-python2.7-systemfolder>/spynnaker/pyNN/models/neural_projections/connectors/distance_dependent_probability_connector.py (Line 58)

I am not quite familiar with the pyNN architecture but I assume its impossible for the connector to know already on construction which are the pre and post populations will be.

```
import spynnaker8 as p
from pyNN.random import RandomDistribution
import pylab
import numpy
import itertools
from pyNN.utility.plotting import Figure, Panel

p.setup(timestep=1.0)

def create_grid(n, label, dx=1., dy=1.):
    grid_structure = p.Grid2D(dx=dx, dy=dy, x0=0, y0=0)
    return p.Population(n*n, p.IF_curr_exp(), structure=grid_structure,
                        label=label)

n = 10
tau_exc = 1.
tau_inh = 1.

# Network
grid = create_grid(n, n)

# Connectors
dist_dep_exc = 'exp(-abs(d)/{tau_exc})'.format(tau_exc=tau_exc)
dist_dep_inh = 'exp(-abs(-d)/{tau_inh})'.format(tau_inh=tau_inh)

import ipdb
# ipdb.set_trace()

exc_connector = p.DistanceDependentProbabilityConnector(dist_dep_exc, allow_self_connections=False)
                                                        # weights=5.,
                                                        # delays='abs(d)')

inh_connector = p.DistanceDependentProbabilityConnector(dist_dep_inh, allow_self_connections=False)
                                                        # weights=5.,
                                                        # delays=0.1)
# Wire grid
p.Projection(grid, grid, exc_connector)
p.Projection(grid, grid, inh_connector)

p.run(10000)
```

andrewgait:
[...]

PR #406, opened 2017-10-13
Fix connectors that use grids so that they work

andrewgait:
Solution of issue https://github.com/SpiNNakerManchester/sPyNNaker/issues/402 - at least for small populations.  More work required on larger grids.  The grid sizes in the newly-created integration tests work, but increasing them leads firstly to weird-looking results and then eventually to crashing code.

Pull requests for sPyNNaker7 and sPyNNaker8 required to get the integration tests which can be used as a basis for further work on larger grids.

### DIFF
--- a/spynnaker/pyNN/models/neural_projections/connectors/distance_dependent_probability_connector.py
+++ b/spynnaker/pyNN/models/neural_projections/connectors/distance_dependent_probability_connector.py
@@ -1,6 +1,7 @@
 from spynnaker.pyNN.utilities import utility_calls
 from .abstract_connector import AbstractConnector
 from spinn_utilities.safe_eval import SafeEval
+from pacman.model.decorators import overrides
 import logging
 import numpy
 import math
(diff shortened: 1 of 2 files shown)


dkfellows:
I suggest that Christian reviews this due to it touching on differences between PyNN versions.

andrewgait:
I think we need to wait for the next version of PyNN before merging this as I believe the required change to the PyNN code has been made on the PyNN master branch, but hasn't made its way into a release yet.  We may need to make more modifications to this pull request and the associated ones in sPyNNaker7 and sPyNNaker8 once that happens.

Christian-B:
I am not aware of this part of pynn changing between 0.8 and 0.9 as that is mainly a neo change

Christian-B:
Loooonnnngggg term this should be in the latest PyNN format but lets do that when we drop support for PyNN 0.7

andrewgait:
I have just noticed that the required change to "fix" this has been made in PyNN (as far as I can tell, at release 0.9.2).  So I'll work on getting this merged at some point soon...

andrewgait:
This is up to date with master and ready to review again now.  However, I am also still trying to work out why the distance-dependent connector in particular isn't behaving as it used to for larger grid sizes, so don't merge it yet...

andrewgait:
OK, I think this is working properly now and can be looked at again.

Issue #395, opened 2017-09-21
Duplication of STDP builds


rowleya:
There appear to be stdp and stdp_mad versions of a lot of builds.  It doesn't seem possible to use the non-mad stdp versions, so these should be removed.

andrewgait:
I'm a bit confused what you mean here; it doesn't seem as though

https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_impl.c

and

https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c

do exactly the same thing at the moment.  If they are doing different things, then are we not building something different when MAD is specified?  

I'll have a look at this in more detail in the near future...

andrewgait:
OK, so in the Python code, get_vertex_executable_suffix() in

https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_stdp.py

returns a binary name with "_stdp_mad" in it, along with the timing/weight rules; there are no examples in the (Python) toolchain of a binary name being returned with just "_stdp" in it, so I agree - any C builds of non-mad STDP code are unnecessary at the moment as they cannot currently be used from a PyNN script, and they should probably be removed from the toolchain.

alan-stokes:
merged

PR #409, opened 2017-10-16
Remove any non-MAD STDP builds

andrewgait:
- These can currently not be accessed from the Python side, so it
  seemed sensible to take them out of the build process

Related to issue https://github.com/SpiNNakerManchester/sPyNNaker/issues/395


andrewgait:
Doing this also demonstrated that there are little to no integration/unit tests of STDP timing rules apart from the spike pair rule with Additive/Multiplicative weight dependence; we should probably write some.

mmikaitis:
Hi Andrew. You can also remove functions post_events_get_window() and post_events_next() from post_events.h as those functions are not used in the realistic (i.e. with delays) or "MAD" stdp model.

andrewgait:
Thanks Mantas, I'll add that to my list of things to check.

PR #444, opened 2017-12-01
in cased called with a numpy array

Christian-B:


spynnaker/pyNN/models/neuron/synapse_types/synapse_type_exponential.py
```diff
@@ -39,16 +40,18 @@ def data_type(self):
...
-        # decay = e^(-ts / tau) as an unsigned long fract
-        tau.apply_operation(
-            lambda x: int(numpy.exp(-ts / x) * ulfract)),
-
-        # init = (tau / ts) * (1 - e^(-ts / tau)) as an unsigned long fract
-        tau.apply_operation(
-            lambda x: int((x / ts) * (1.0 - numpy.exp(-ts / x)) * ulfract))
-    )
+    # decay = e^(-ts / tau) as an unsigned long fract
+    decay = lambda x: int(numpy.exp(-ts / x) * ulfract)  # noqa E731
```
rowleya: This disables a QA check but I don't believe it should.  It would be better to make decay (and init later) into functions and pass them that way, as the function is used more than once in the code.

See https://stackoverflow.com/questions/45790269/pep-8-warning-do-not-use-a-lambda-expression-use-a-def-for-a-defaultdict-lambd

spynnaker/pyNN/models/neuron/synapse_types/synapse_type_exponential.py
```diff
@@ -39,16 +40,18 @@ def data_type(self):
...
+    # init = (tau / ts) * (1 - e^(-ts / tau)) as unsigned long fract
+    init = lambda x: int((x / ts) * (1.0 - numpy.exp(-ts / x)) * ulfract)  # noqa E731,
```
rowleya: See above re: removing noqa and replacing with a function.

rowleya: Having discussed this with Christian, I withdraw the objection, as the lambda function is dependent on other functions of values, which would make the code less readable.

PR #445, opened 2017-12-05
Keep data in matrix

Christian-B:
This avoids converting v ect data from a matrix to a list or (id, time, value) tuples and then back again for Neo.

Note: get_sampling_interval(self, variable): is in preparation for different ones per variable


Christian-B:
MUST be done together with:
https://github.com/SpiNNakerManchester/sPyNNaker7/pull/47
https://github.com/SpiNNakerManchester/sPyNNaker8/pull/91

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -1,19 +1,107 @@
...
+        :param application_vertex:
+        :param variable: PyNN name for the variable (V, gsy_ihn ect
+        :type variable: str
+        :param n_machine_time_steps:
+        :return:
+        """
+        vertices = graph_mapper.get_machine_vertices(application_vertex)
+        progress = ProgressBar(
+                vertices, "Getting {} for {}".format(variable, label))
+        sampling_interval = self.get_sampling_interval(variable)
```
rowleya: I don't think this is a thing - there is self.get_neuron_sampling_interval...

spynnaker/pyNN/models/recording_common.py
```diff
@@ -16,13 +16,19 @@
...
 
         :param population: the population to record for
+        :param simulator: the spinnaker control class
         """
 
         self._population = population
+
+        self._sampling_interval = None
+        if sampling_interval is not None:
+            self._sampling_interval = sampling_interval
```
rowleya: or just ```self._sampling_interval = sampling_interval```

spynnaker/pyNN/models/recording_common.py
```diff
@@ -61,6 +67,10 @@ def _record(self, variable, new_ids, sampling_interval, to_file):
         # Get bit array of indices to record for this variable
         indices = self._indices_to_record[variable]
 
+        # update sampling interval
+        if sampling_interval is not None:
+            self._sampling_interval = sampling_interval
```
rowleya: Probably don't need to do it again...

spynnaker/pyNN/models/recording_common.py
```diff
@@ -101,12 +111,59 @@ def _set_spikes_recording(self):
...
+
+        (data, ids, sampling_interval) = self._get_recorded_matrix(variable)
+        return self.pynn7_format(data, ids, sampling_interval)
+        (data, ids, sampling_interval) = self._get_recorded_matrix(variable)
+        return self.pynn7_format(data, ids, sampling_interval)
+
+    def _get_recorded_matrix(self, variable):
         """ method that contains all the safety checks and gets the recorded
-        data from the vertex
+        data from the vertex in P7NN0.7 format
```
rowleya: P7NN -> PyNN.  I thought this wasn't in PyNN 7 format though...

spynnaker/pyNN/models/recording_common.py
```diff
@@ -115,42 +172,46 @@ def _get_recorded_variable(self, variable):
...
                 "This population has not been set to record {}"
                 .format(variable))
 
-        elif not sim.has_ran:
+        if not sim.has_ran:
             logger.warn(
                 "The simulation has not yet run, therefore {} cannot"
                 " be retrieved, hence the list will be empty".format(
                     variable))
             data = numpy.zeros((0, 3))
```
rowleya: No worse than it used to be - however gsyn is (0, 4) in PyNN 7...

spynnaker/pyNN/models/spike_source/spike_source_array.py
```diff
@@ -174,6 +174,10 @@ def set_recording_spikes(self, new_state=True):
         self._requires_mapping = not self._spike_recorder.record
         self._spike_recorder.record = new_state
 
+    @overrides(AbstractSpikeRecordable.get_spikes_sampling_interval)
+    def get_spikes_sampling_interval(self):
+        return globals_variables.get_simulator().machine_time_step
```
rowleya: It may just be that this isn't done yet, but I think this should return ```self._spike_recorder.get_spikes_sampling_interval```

spynnaker/pyNN/models/spike_source/spike_source_poisson.py
```diff
@@ -532,6 +532,10 @@ def set_recording_spikes(self, new_state=True):
             self._spike_recorder.record != new_state)
         self._spike_recorder.record = new_state
 
+    @overrides(AbstractSpikeRecordable.get_spikes_sampling_interval)
+    def get_spikes_sampling_interval(self):
+        return globals_variables.get_simulator().machine_time_step
```
rowleya: As with the SpikeSourceArray...

spynnaker/pyNN/models/recording_common.py
Christian-B: Actually Worse. Not sure sampling_interval belongs here at all.

spynnaker/pyNN/models/recording_common.py
Christian-B: Actually Worse. Not sure sampling_interval belongs here at all.

spynnaker/pyNN/models/recording_common.py
Christian-B: Agreed

spynnaker/pyNN/models/recording_common.py
Christian-B: Yes and NO.
You are correct that in Pynn0.7 gsyn is (0,4) BUT
In sPyNNaker7/spynnaker7/pyNN/models/pynn_population.py method get_gsyn this method was and still is called twice once for gsyn_exc and once for gsyn_inh and then the results are merged.

spynnaker/pyNN/models/spike_source/spike_source_array.py
Christian-B: I have not yet decided where to keep the spike sample_interval but YES it should be in one place and YES it should be where the parameters are passed to the C doe so that is worth considering.

spynnaker/pyNN/models/spike_source/spike_source_poisson.py
Christian-B: In this case NO.  As this class has stuff like generate_data_specification this is where the sampling interval belongs.

rowleya: OK, but then the data might be replicated - there is also a sampling interval in the recorder (unless there isn't)

spynnaker/pyNN/models/recording_common.py
rowleya: Ahh OK

spynnaker/pyNN/models/spike_source/spike_source_array.py
rowleya: OK - this isn't critical for this pull request in any case

spynnaker/pyNN/models/spike_source/spike_source_poisson.py
rowleya: As above, this isn't yet critical, as long as it works in the next one ;)

Christian-B: As above I am not sure sampling Interval belongs in recorder.

Keep trying to test that but on top of an unstable master this is proving difficult.

spynnaker/pyNN/models/common/neuron_recorder.py
Christian-B: fixed

spynnaker/pyNN/models/recording_common.py
Christian-B: Will look at this in the selective recording PR

spynnaker/pyNN/models/recording_common.py
Christian-B: Will look at this in the selective recording PR

spynnaker/pyNN/models/spike_source/spike_source_array.py
Christian-B: Will look at this later when I do selective recording of spikes

spynnaker/pyNN/models/spike_source/spike_source_poisson.py
Christian-B: Will look at this later when I do selective recording of spikes

spynnaker/pyNN/models/recording_common.py
Christian-B: In Selective record sampling intervals are held in NeuronRecorder and no longer in RecordingCommon

spynnaker/pyNN/models/recording_common.py
Christian-B: In Selective record sampling intervals are held in NeuronRecorder and no longer in RecordingCommon

alan-stokes:
having looked,accepting that its acutally missing the sampling pushed down to c code, so this would go boom if the samplign was anything but the machine time step. but the strucutre is fine

Christian-B:
part of https://github.com/SpiNNakerManchester/sPyNNaker/pull/461

PR #450, opened 2017-12-14
fixes issues with machines and spalloc

alan-stokes:
allows to adapt to spalloc machines which come with different formats as the original one it was mapped to.

has a fec related one as well.
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/256


andrewgait:
Merging, think this red cross is an issue on Travis.

PR #461, opened 2018-01-24
Index recording

Christian-B:
This and its sister PR (utils, FEC, p7 and p8) are ready to be pulled in.

There is a minor difference of the last spikes being lost or spikes being moved by one timestep.
I think this may be the system moving the spikes and not the recorder missing them.

Still to do but can be in a later PR.

1. Use views instead of indexes parameter

2. Improve auto pause resume cost models to allocate remaining SDRAm much better.


Christian-B:
Must be done at the same time as: https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/272

Christian-B:
includes https://github.com/SpiNNakerManchester/sPyNNaker/pull/445 and https://github.com/SpiNNakerManchester/sPyNNaker/pull/449

Christian-B:
Builds on spike_recording https://github.com/SpiNNakerManchester/sPyNNaker/pull/462
which build on selective_record https://github.com/SpiNNakerManchester/sPyNNaker/pull/449
which build on keep_data-in_matrix https://github.com/SpiNNakerManchester/sPyNNaker/pull/445

spynnaker/pyNN/models/recording_common.py
```diff
@@ -20,9 +20,11 @@ def __init__(self, population):
         """ object to hold recording behaviour
 
         :param population: the population to record for
+        :param simulator: the spinnaker control class
```
rowleya: I don't think this has actually been added as a parameter...

spynnaker/pyNN/models/recording_common.py
```diff
@@ -91,22 +92,45 @@ def _set_v_recording(self):
         """
         self._population._vertex.set_recording("v")
 
-    def _set_spikes_recording(self):
+    def _set_spikes_recording(self, sampling_interval, indexes=None):
```
rowleya: Is there a need for this empty method?

spynnaker/pyNN/models/spike_source/spike_source_array.py
```diff
@@ -167,13 +167,24 @@ def is_recording_spikes(self):
         return self._spike_recorder.record
 
     @overrides(AbstractSpikeRecordable.set_recording_spikes)
-    def set_recording_spikes(self, new_state=True):
+    def set_recording_spikes(
+            self, new_state=True, sampling_interval=None, indexes=None):
+        if sampling_interval is not None:
+            logger.warning("Sampling interval currently not sopported for "
```
rowleya: sopported -> supported

spynnaker/pyNN/models/spike_source/spike_source_array.py
```diff
@@ -167,13 +167,24 @@ def is_recording_spikes(self):
...
+                           "SpikeSourceArray so being ignored")
+        if indexes is not None:
+            logger.warning("Indexes currently not sopported for "
```
rowleya: sopported -> supported

spynnaker/pyNN/models/common/abstract_spike_recordable.py
```diff
@@ -45,3 +56,14 @@ def get_spikes(
         :return: A numpy array of 2-element arrays of (neuron_id, time)\
                 ordered by time
         """
+
+    @abstractmethod
+    def get_spikes_sampling_interval(self):
+        """ Returns the current sampling interval for spikes
+        :return: Sampling interval in micro seconds
+
+        if self.is_recording_spikes():
```
rowleya: Is the content here meant to be commented out?

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -1,58 +1,520 @@
...
-    def get_dtcm_usage_in_bytes(self):
-        return self.N_BYTES_PER_NEURON * sum(self._record.values())
+        data_size = self.get_buffered_sdram_per_record(
+            variable,  vertex_slice)
+        if rate == 1:
+            return data_size
+        else:
+            return data_size / rate
+
+    def get_sampling_overflow_sdram(self, vertex_slice):
```
rowleya: Not needed if get_buffered_sdram_per_timestep is changed to get_n_timesteps_in_buffer_space.

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
```diff
@@ -252,14 +251,26 @@ def mark_no_changes(self):
 
     def _get_buffered_sdram_per_timestep(self, vertex_slice):
```
rowleya: Not needed if get_buffered_sdram_per_timestep is changed to get_n_timesteps_in_buffer_space.

spynnaker/pyNN/models/neuron/population_machine_vertex.py
```diff
@@ -130,8 +144,9 @@ def get_minimum_buffer_sdram_usage(self):
 
     @overrides(AbstractReceiveBuffersToHost.get_n_timesteps_in_buffer_space)
     def get_n_timesteps_in_buffer_space(self, buffer_space, machine_time_step):
+        safe_space = buffer_space - self._overflow_sdram
```
rowleya: If get_buffered_sdram_per_timestep is changed to get_n_timesteps_in_buffer_space, this can now call neuron_recorder.get_n_timesteps_in_buffer_space.

spynnaker/pyNN/models/neuron/population_machine_vertex.py
```diff
@@ -47,13 +47,27 @@ class PopulationMachineVertex(
 
     def __init__(
             self, resources_required, is_recording, minimum_buffer_sdram_usage,
-            buffered_sdram_per_timestep, label, constraints=None):
+            buffered_sdram_per_timestep, label, constraints=None,
```
rowleya: If get_buffered_sdram_per_timestep is changed to get_n_timesteps_in_buffer_space, this should take the neuron_recorder instead of buffered_sdram_per_timestep and overflow_sdram.

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
```diff
@@ -270,16 +281,18 @@ def create_machine_vertex(
...
         minimum_buffer_sdram = recording_utilities.get_minimum_buffer_sdram(
-            buffered_sdram_per_timestep, n_machine_time_steps,
-            self._minimum_buffer_sdram)
+            buffered_sdram, self._minimum_buffer_sdram)
+        overflow_sdram = self._neuron_recorder.get_sampling_overflow_sdram(
+            vertex_slice)
         vertex = PopulationMachineVertex(
             resources_required, is_recording, minimum_buffer_sdram,
-            buffered_sdram_per_timestep, label, constraints)
+            buffered_sdram_per_timestep, label, constraints, overflow_sdram)
```
rowleya: If get_buffered_sdram_per_timestep is changed to get_n_timesteps_in_buffer_space, this should pass in the neuron_recorder instead of buffered_sdram_per_timestep and overflow_sdram.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -1,58 +1,520 @@
...
+                # Non recoding neurons write to an extra slot
+                n_neurons += 1
+            out_spike_words = int(math.ceil(n_neurons / 32.0))
+            out_spike_bytes = out_spike_words * self.N_BYTES_PER_WORD
+            return self.N_BYTES_FOR_TIMESTAMP + out_spike_bytes
         else:
+            return self.N_BYTES_FOR_TIMESTAMP + \
+                        n_neurons * self.N_BYTES_PER_VALUE
+
+    def get_buffered_sdram_per_timestep(self, variable, vertex_slice):
```
rowleya: It might make more sense for this to be get_n_timesteps_in_buffer_space (as this is the actual call that gets made to the vertices where this is used).  This would then avoid having to return a fractional value when only some timesteps are being recorded.  Places that would have to be changed to make this work are (these are also tagged in the review):
1. You wouldn't need get_sampling_overflow_sdram below
1. You wouldn't need _get_buffered_sdram_per_timestep in abstract_population_vertex.
1. PopulationMachineVertex would need to be passed the neuron recorder.  
1. PopulationMachineVertex would then call neuron_recorder.get_n_timesteps_in_buffer_space from its own version of that method.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -1,58 +1,520 @@
...
+                fragment = (record[:, 1:] / float(DataType.S1615.scale))
+            else:
+                missing_str += "({}, {}, {}); ".format(
+                    placement.x, placement.y, placement.p)
+                # Start the fragment for this slice empty
+                fragment = numpy.empty((expected_rows, n_neurons))
+                for i in xrange(0, expected_rows):
+                    time = i * sampling_interval
+                    # Check if there is data for this timestep
+                    indexes = numpy.where(record[:, 0] == time)
```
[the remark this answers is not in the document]
Christian-B: yes fixed

neural_modelling/src/neuron/neuron.c
```diff
@@ -116,31 +149,103 @@ static inline void _print_neuron_parameters() {
...
+        //Not the first run so do not reset counters
+        return;
+    }
+    if (global_record_params->spike_rate == 0){
+        // Setting increment to zero means v_index will never equal v_rate
+        spike_increment = 0;
+        // Index is not rate so does not record. Nor one so we never reset
+        spike_index = 2;
+    }
+    else {
```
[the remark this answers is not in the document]
Christian-B: I prefer the else on the same line so fixed

neural_modelling/src/neuron/neuron.c
```diff
@@ -116,31 +149,103 @@ static inline void _print_neuron_parameters() {
...
 //! \brief does the memory copy for the neuron parameters
 //! \param[in] address: the address where the neuron parameters are stored
 //! in SDRAM
 //! \return bool which is true if the mem copy's worked, false otherwise
 bool _neuron_load_neuron_parameters(address_t address){
     uint32_t next = START_OF_GLOBAL_PARAMETERS;
 
-    log_info("loading neuron global parameters");
+    log_info("loading parameters");
+    //log_info("loading global record parameters");
```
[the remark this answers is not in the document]
Christian-B: Currently turning on  make SPYNNAKER_DEBUG=DEBUG caused an 'ITCM' overflowed by over 2000 bytes so not wasting time doing log_debug statements at this time.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -1,58 +1,520 @@
...
+            # There is one column for time and one for each neuron recording
+            n_rows = record_length // row_length
+            # Converts bytes to ints and make a matrix
+            record = (numpy.asarray(record_raw, dtype="uint8").
+                      view(dtype="<i4")).reshape((n_rows, (n_neurons + 1)))
+            # Check if you have the expected data
+            if not missing_data and n_rows == expected_rows:
+                # Just cut the timestamps off to get the fragment
+                fragment = (record[:, 1:] / float(DataType.S1615.scale))
+            else:
```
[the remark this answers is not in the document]
Christian-B: Actually worse it falls in the else part.

Fix applied.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -1,58 +1,520 @@
...
+                if neurons_recording < vertex_slice.n_atoms:
+                    # For spikes the overflow position is also returned
+                    neurons_recording += 1
+            # Read the spikes
+            n_words = int(math.ceil(neurons_recording / 32.0))
+            n_bytes = n_words * self.N_BYTES_PER_WORD
+            n_words_with_timestamp = n_words + 1
+
+            # for buffering output info is taken form the buffer manager
+            neuron_param_region_data_pointer, data_missing = \
```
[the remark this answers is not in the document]
Christian-B: Test show it did actually work but agree it is strange to read and process data when you don't need any.

Fixed

neural_modelling/src/neuron/neuron.c
```diff
@@ -261,23 +387,67 @@ bool neuron_initialise(address_t address, uint32_t recording_flags_param,
         }
     }
 
+    //! Setting this to MAX so it can be used to detect first run
+    recording_flags = 0xFFFFFFFF;
```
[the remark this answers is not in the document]
Christian-B: Agreed fixed

neural_modelling/src/neuron/neuron.c
```diff
@@ -116,31 +149,103 @@ static inline void _print_neuron_parameters() {
 #endif // LOG_LEVEL >= LOG_DEBUG
 }
 
+
+void _reset_record_counter(){
+    if (recording_flags != 0xFFFFFFFF){
```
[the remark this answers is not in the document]
Christian-B: See reply above

spynnaker/pyNN/models/recording_common.py
Christian-B: it had been and the replaced with cleaner getting via global variables

spynnaker/pyNN/models/spike_source/spike_source_array.py
Christian-B: That's what you get for hiring someone who spells like Alan.

spynnaker/pyNN/models/common/abstract_spike_recordable.py
Christian-B: Yes or better yet just deleted!  It is an abstractmethod after all.

spynnaker/pyNN/models/common/neuron_recorder.py
Christian-B: get_n_timesteps_in_buffer_space is an UGLY approach that needs stripping out.

That assumes all processors are given an equal share of the remaining SDRAM and then the auto pause resume puts a break when the first one runs out.

Long term a better solution is to get the total sdram per timestep for ALL processor on the chip. Compute the max run time and then allocate the SDram per processor. But that is work for another day,

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
Christian-B: See above on why get_n_timesteps_in_buffer_space is bad!

spynnaker/pyNN/models/neuron/population_machine_vertex.py
Christian-B: See above on why get_n_timesteps_in_buffer_space is bad!

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
Christian-B: See above on why get_n_timesteps_in_buffer_space is bad!

spynnaker/pyNN/models/common/neuron_recorder.py
rowleya: I am fine with leaving this for another day - it just seemed like you were going out of your way to do per-timestep when this code would find it easier to work out the number of timesteps in a given buffer space (the code would be much clearer in that case). 

I have started a ticket for some possible ideas here: SpiNNakerManchester/SpiNNFrontEndCommon/293

PR #466, opened 2018-02-08
Get synaptic input as array

oliverrhodes:
Laying the foundations for multi-receptor and multi-compartment neurons


oliverrhodes:
@rowleya - I've made the changes we discussed, but had to add back the input_type_get_input_value() function to enable scaling of the synaptic input to resolve g_syn in order to record. This flow will hopefully be tidied up in the follow-up work to update ring buffer scaling.

oliverrhodes:
@andrewgait - please take a look once you're back in the office

neural_modelling/src/neuron/input_types/input_type.h
```diff
@@ -10,17 +10,18 @@ typedef struct input_type_t* input_type_pointer_t;
 //! \param[in] value The value of the input before scaling
 //! \param[in] input_type The input type pointer to the parameters
 //! \return The value of the input after scaling
-static input_t input_type_get_input_value(
-    input_t value, input_type_pointer_t input_type);
+static input_t* input_type_get_input_value(
```
rowleya: This probably doesn't need to return the pointer, as with the other functions...

neural_modelling/src/neuron/input_types/input_type.h
```diff
@@ -10,17 +10,18 @@ typedef struct input_type_t* input_type_pointer_t;
 //! \param[in] value The value of the input before scaling
```
rowleya: You might want to change this to \param[in/out] and update the comment

neural_modelling/src/neuron/input_types/input_type.h
```diff
@@ -10,17 +10,18 @@ typedef struct input_type_t* input_type_pointer_t;
...
 //! \param[in] input_type The input type pointer to the parameters
 //! \return The value of the input after scaling
-static input_t input_type_get_input_value(
-    input_t value, input_type_pointer_t input_type);
+static input_t* input_type_get_input_value(
+    input_t* value, input_type_pointer_t input_type,
+	uint16_t num_receptors);
 
 //! \brief Converts an excitatory input into an excitatory current
 //! \param[in] exc_input The total excitatory input this timestep - note that
```
rowleya: As above re updating the comment and in/out

neural_modelling/src/neuron/input_types/input_type.h
```diff
@@ -10,17 +10,18 @@ typedef struct input_type_t* input_type_pointer_t;
...
 //! \param[in] input_type The input type pointer to the parameters
 //! \param[in] membrane_voltage The membrane voltage to use for the input
 //! \return The excitatory input current
-static input_t input_type_convert_excitatory_input_to_current(
-    input_t exc_input, input_type_pointer_t input_type,
+static void input_type_convert_excitatory_input_to_current(
+    input_t* exc_input, input_type_pointer_t input_type,
     state_t membrane_voltage);
 
 //! \brief Converts an inhibitory input into an inhibitory current
```
rowleya: As above re-comment (on line below, but github won't let me do that!)

neural_modelling/src/neuron/models/neuron_model.h
```diff
@@ -32,8 +32,9 @@ void neuron_model_set_global_neuron_params(
 //! \return state_t which is the value to be compared with a threshold value
 //!     to determine if the neuron has spiked
 state_t neuron_model_state_update(
-    input_t exc_input, input_t inh_input, input_t external_bias,
-    neuron_pointer_t neuron);
+		uint16_t num_excitatory_inputs, input_t* exc_input,
```
rowleya: May want to update the comments again

neural_modelling/src/neuron/input_types/input_type.h
oliverrhodes: As discussed, I will leave this in due to how the function is used in neuron.c

neural_modelling/src/neuron/input_types/input_type.h
oliverrhodes: Done

neural_modelling/src/neuron/models/neuron_model.h
oliverrhodes: Done

neural_modelling/src/neuron/input_types/input_type.h
oliverrhodes: Done

neural_modelling/src/neuron/input_types/input_type.h
oliverrhodes: Done

rowleya:
I have approved this, and was about to merge, but then wondered, does this have a knock-on effect on the new model templates?

andrewgait:
Yes it will have an effect; I'm guessing not a huge amount though.  There aren't any extra "input_type"s defined there at the moment, so it will only be the neuron update function and the synapse_type changes that will need to be made.  I'd be happy to look at that next week unless Oliver would prefer to do so...

oliverrhodes:
Afternoon chaps, I've just done a PR to update sPyNNaker8NewModelTemplate to accommodate these changes (same branch name). 

As discussed, I haven't done the same for sPyNNaker7NewModelTemplate, so this will effectively be the end of our support for that repository.

If we're happy, then I think this is ready to merge....

PR #470, opened 2018-02-19
Pass down max feasible atoms per core

oliverrhodes:
This update is required to allow changing of the number of neuron_id_bits in the synaptic row. (neuron_id_bits should default to 8 for the standard models, but can be reduced when extra bits are required in the synaptic row for additional synapses etc).


alan-stokes:
after a quick glance at this, im a bit confused. why isnt max_atoms_per_core not usable for this? max_feasible_atoms_per_core sounds like the same semantics to max_atoms_per_core.

oliverrhodes:
Hi Alan,

The difference is that the user can't change the max_feasible_atoms_per_core. Instead this provides a way for us to calculate how many bits are required in the synaptic word to identify the neurons (this value is hardcoded on the C side, so we don't want things changing on the Python side).

For example, if we just used max_atoms_per_core, with the default value of 255, then we'd calculate that 8 bits were required for neuron IDs. However, if the user manually changed this, say to 63, then we'd now calculate that only 6 bits were required for neuron IDs, however our C code would still be hardcoded to use 8, and hence we'd get a mismatch. 

Maybe it would be sensible to rename DEFAULT_MAX_ATOMS_PER_CORE to MAX_FEASIBLE_NEURONS_PER_CORE, and set this per build. Then we could simply pass this to the AbstractPopulationVertex __init__() ?

alan-stokes:
hi oliver. so to ask what seems like a silly question. instead of it being hard coded in the c code. couldn't it be passed down from the python to the c code as a config param?

at which point you can still only need one type of max atoms?

rowleya:
So the issue here is that max_atoms_per_core is set by the user whereas max_feasible_atoms_per_core is something that is defined in the C code, and some parts of the python code need to know about.  The two things are completely different.  Thus it is correct to keep these separate.

alan-stokes:
from what ive read, this doesnt sound true. the c code should be able to configure itself for the atoms its given. If it cant, then id like to understand why?, the max atoms may be set by the end user, but there's already a hard top limit here, so that limit needs adjusting to reflect this new "feasible" point. At the moment, theres nothign ehre that i can see which reflects a distinction between max user and max c.

rowleya:
OK, it is *possible* to make it that the C code configures itself for this, but this is another step down the path.  This pull request will help move in that direction.  I think the code does reflect the change in the hard top limit - this is set as DEFAULT_MAX_ATOMS_PER_CORE in all the models, and exactly this value is then passed to the AbstractPopulationVertex.  The max_atoms_per_core is also initially set to this value, and then this is allowed to vary by user choice (so long as it is always less than or equal to the default).

rowleya:
I guess a summary of this change is that the original default max atoms per core is now passed down through the python code, where it wasn't before.

alan-stokes:
but this change wouldn't be needed if the work to change the c code to configure itself was done, at which point, this becomes reverted. so how hard would it be to do that configuration?

rowleya:
It would be quite hard without upsetting a lot of the code I believe.  The main issue would be in the generation of masks to be used in the code - currently a #define is used, but this would need to be replaced by a variable with the appropriate value for the number of neurons per core.  I can have a quick look to see...

rowleya:
Changed in #473, which pulls on to this branch if you like it...

alan-stokes:
i think #473 is much better than this one. this one felt like a sticky plaster, whereas #473 looks like just an extra parameter and its processing in c code.

oliverrhodes:
Hi Andrew,

Please can you review the latest changes to this branch when you get a chance - things have changed a bit since the merge of structural plasticity into master. 

I've made a few updates to get things building: of note is the removal of IF_cond_exp_stdp_mad_pair_additive_structural from the list of builds as it no longer fits into ITCM, and I've also switched synapse_dynamics_initialise to return an address again (it was returning a bool).

Otherwise this now seems to be working OK: I've tested with models containing 2 excitatory and 2 inhibitory synapses; and also the STDP integration tests still run successfully. Therefore, if structural plasticity still functions correctly, I think this is nearly ready to merge.

Oliver

alan-stokes:
IF_cond_exp_stdp_mad_pair_additive_structural from the list of builds as it no longer fits into ITCM

whys that? is it print statements? does the chrisitan's iobuf fixes going to get it back under control?

rowleya:
There have been a lot of changes to master - I guess this one just pushes that model over the edge.  We could look at changing the flags to optimise for space and see if that helps for now.

Christian's stuff should help here of course, and we will also have to start working out the details of the split neuron-synapse model, and get that implemented at some point...

rowleya:
Note - there is a flake8 error in here...

oliverrhodes:
Petrut has now confirmed this functions correctly with structural plasticity, so I think it's ready to merge.

Note that IF_cond_exp_stdp_mad_pair_additive_structural has been commented from the build as it exceeds ITCM with version 5.4 of the compiler.

rowleya:
I have managed to re-enable the build of IF_cond_exp_stdp_mad_pair_additive_structural.  This was done by compiling for space efficiency rather than time.  This requires:
SpiNNakerManchester/SpiNNFrontEndCommon#308.

PR #473, opened 2018-02-27
Updates to make the code use n_neurons for ring buffer sizing

rowleya:
This adjusts the code to use n_neurons instead of passing down the max feasible atoms per core.  This not only appears to work correctly, but it also reduces the number of instructions by 2 within that part of the code!  Not sure why this should be, but it works in any case...

This has the advantage that the ring buffer size is now related to the number of neurons per core.  This doesn't make it as small as it can be (it is sized to the next closest power of 2 of the number of neurons -
 to make it exact would require other changes to be made in the future), but the reduction is still useful.

Note that this is a pull request to pass_down_max_feasible_atoms_per_core.  As such, it currently fails due to a dependency on spinn_common branch by the same name (but different to the name of this branch ;).

### DIFF
--- a/neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c
+++ b/neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c
@@ -12,6 +12,13 @@
 #include "timing_dependence/timing.h"
 #include <string.h>
 #include <debug.h>
+#include <utils.h>
+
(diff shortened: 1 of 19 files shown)


alan-stokes:
id rubber this, but dont know how the rest are considered. but i approve of this one

rowleya:
@oliverrhodes Are you happy with this idea?

alan-stokes:
would need whatever those x's are about to be fixed. as a note.

Issue #453, opened 2017-12-21
RNG error with FixedTotalNumberConnector 


apdavison:
This line:

https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/spynnaker/pyNN/models/neural_projections/connectors/multapse_connector.py#L68

calls the PyNN RNG object with a distribution name 'multinomial' and parameters in the form of a list. As of PyNN 0.8, the RNGs support only a limited number of distributions with standardized names, and 'multinomial' is not in that list.

I have created PyNN ticket https://github.com/NeuralEnsemble/PyNN/issues/555 to address the PyNN side of this, but you will need to change the parameters to a dictionary (keys `n` and `pvals`) instead of a list.

(Unfortunately, the parameters still need to be a list for PyNN 0.7)

```
Traceback (most recent call last):
  File "tmp/example.py", line 55, in <module>
    net.simulate()
  File "/tmp/job2972132055770751070.tmp/tmp/network.py", line 391, in simulate
    self.sim.run(self.sim_dict['t_sim'])
  File "/opt/git/sPyNNaker8/spynnaker8/__init__.py", line 515, in run
    return __pynn_run(simtime, callbacks=callbacks)
  File "/usr/local/lib/python2.7/dist-packages/pyNN/common/control.py", line 110, in run
    return run_until(simulator.state.t + simtime, callbacks)
  File "/usr/local/lib/python2.7/dist-packages/pyNN/common/control.py", line 92, in run_until
    simulator.state.run_until(time_point)
  File "/opt/git/sPyNNaker8/spynnaker8/spinnaker.py", line 128, in run_until
    self._run(tstop - self.t)
  File "/opt/git/sPyNNaker8/spynnaker8/spinnaker.py", line 175, in _run
    AbstractSpiNNakerCommon.run(self, duration_ms)
  File "/opt/git/sPyNNaker/spynnaker/pyNN/abstract_spinnaker_common.py", line 312, in run
    AbstractSpinnakerBase._run(self, run_time)
  File "/opt/git/SpiNNFrontEndCommon/spinn_front_end_common/interface/abstract_spinnaker_base.py", line 850, in _run
    self._get_machine(total_run_time, n_machine_time_steps)
  File "/opt/git/SpiNNFrontEndCommon/spinn_front_end_common/interface/abstract_spinnaker_base.py", line 1342, in _get_machine
    inputs, algorithms, outputs, "machine_generation")
[...]

PR #474, opened 2018-02-27
Multapse connector edits

andrewgait:
Editing to make the MultapseConnector / FixedTotalNumberConnector work; fixing issue https://github.com/SpiNNakerManchester/sPyNNaker/issues/453 but also adding arguments to allow for self connections and replacement within the connector.


andrewgait:
Something like this, I think... after some arguing with numpy.concatenate... :-)

andrewgait:
Good point; code updated accordingly.  Looks a lot less confusing now! :-)

spynnaker/pyNN/models/neural_projections/connectors/multapse_connector.py
```diff
@@ -69,9 +79,8 @@ def _update_synapses_per_post_vertex(self, pre_slices, post_slices):
             prob_connect = [
                 float(pre.n_atoms * post.n_atoms) / float(n_connections)
                 for pre in pre_slices for post in post_slices]
-            self._synapses_per_edge = self._rng.next(
-                1, distribution="multinomial", parameters=[
-                    self._num_synapses, prob_connect])
+            self._synapses_per_edge = self.get_rng_next(
```
rowleya: Quick test - this doesn't seem to exist...

andrewgait: It had to move out to sPyNNaker7/8 as the "multinomial" keyword was removed between PyNN 0.7 and 0,8.  I believe they plan to add this back in again but for now I've split it up until they actually fix it.  (It's not in 0.9.2 yet as far as the documentation says...)

rowleya: Ahh OK, I missed this.  Best thing to do then is to add this as an @abstractmethod to show that you know that it isn't defined here.

PR #479, opened 2018-03-05
Index based prob connector

andrewgait:
Add an index-based probability connector (functionality added between PyNN0.7 and 0.8) but not supported by us until now.  

The idea is that the user supplies a probability expression which is dependent upon the indices i and j of the two populations being connected, and connections are then created based upon this expression.


spynnaker/pyNN/models/neural_projections/connectors/index_based_probability_connector.py
```diff
@@ -0,0 +1,190 @@
...
+        # note: this only needs to be done once
+        if self._probs is None:
+            # numpy array of probabilities using the index_expression
+            self._probs = numpy.array([[_index_expr_context.eval(
+                self._index_expression, i=i, j=j)
+                for j in range(self._n_post_neurons)]
+                for i in range(self._n_pre_neurons)])
+
+    @overrides(AbstractConnector.get_delay_maximum)
+    def get_delay_maximum(self):
```
rowleya: I am guessing this should call the update function above, to ensure self._probs is filled in...

spynnaker/pyNN/models/neural_projections/connectors/index_based_probability_connector.py
```diff
@@ -0,0 +1,190 @@
...
+            Population.
+        """
+        super(IndexBasedProbabilityConnector, self).__init__(safe, verbose)
+        self._index_expression = index_expression
+        self._allow_self_connections = allow_self_connections
+
+        self._probs = None
+
+    def _update_probs_from_index_expression(self, pre_vertex_slice,
+                                            post_vertex_slice):
```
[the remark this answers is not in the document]
andrewgait: I agree - this is probably there from having copied the structure from a different connector where the update function did use the pre- and post- slices.

spynnaker/pyNN/models/neural_projections/connectors/index_based_probability_connector.py
andrewgait: Agreed.

PR #491, opened 2018-03-22
[Structural plasticity] Random delays, per projection information 

dkfellows:
These are a few tweaks that came in after I approved the branch


dkfellows:
It'd be really great if you reduced the number of mysterious constants in the code. The ones in `__write_post_to_pre_table` in `synapse_dynamics_structural_common.py` are particularly weird.

pabogdan:
@dkfellows you are correct. I don't know where my brain was when I wrote those bit masks

alan-stokes:
is this still waiting on me to do something? or someone else?

pabogdan:
@alan-stokes this is waiting on me. I will let you know when all of the requested changes are done!

alan-stokes:
k, just trying to prune my requested reviews, before getting back to python to java conversions..........prograstianting. 

plus, trying to find lucas c++ many board code, dont really want to spend the time reinventing the wheel

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c
```diff
@@ -93,7 +93,8 @@ static inline final_state_t _plasticity_update_synapse(
     // Process events in post-synaptic window
     while (post_window.num_events > 0) {
         const uint32_t delayed_post_time = *post_window.next_time
-                                           + delay_dendritic;
+//                                           + delay_dendritic;
+                                           + 0;
```
rowleya: I am slightly worried about the effect of this on normal plasticity.  I can't remember the details, but is this correct?

spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py
```diff
@@ -527,13 +539,18 @@ def __write_presynaptic_information(self, spec, application_graph,
             # Custom header for commands / controls
             # currently, controls = True if the subvertex (on the current core)
             # is part of this population
+            # In other words, controls is 0 if connection is considered FF
+            # otherwise the connections if lat
             controls = current_key in np.asarray(subpopulation_list)[:0]
             spec.write_value(data=int(controls), data_type=DataType.UINT16)
+            # TODO Write connection type
```
rowleya: Is this now done below?

spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py
```diff
@@ -615,12 +632,16 @@ def __write_post_to_pre_table(self, spec, app_vertex, post_slice,
                     pre_vertex_slice = graph_mapper._slice_by_machine_vertex[
                         row[2].pre_vertex]
                     pre_vertex_id = source - pre_vertex_slice.lo_atom
-                    masked_pre_vertex_id = pre_vertex_id & (2 ** 17 - 1)
+                    masked_pre_vertex_id = \
+                        pre_vertex_id & \
+                        SynapseDynamicsStructuralCommon.BIT_MASK_16_BIT
```
rowleya: This isn't the same as it was previously (2 ** 17 - 1 vs 2 ** 16 - 1).  Is this correct?

spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py
```diff
@@ -615,12 +632,16 @@ def __write_post_to_pre_table(self, spec, app_vertex, post_slice,
...
 
                     # Select population index
                     pop_index = population_to_subpopulation_information. \
                         keys().index(row[1].pre_vertex)
-                    masked_pop_index = pop_index & (2 ** 9 - 1)
+                    masked_pop_index = \
+                        pop_index & \
+                        SynapseDynamicsStructuralCommon.BIT_MASK_8_BIT
```
rowleya: As above - is this correct?

spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py
```diff
@@ -637,12 +658,19 @@ def __write_post_to_pre_table(self, spec, app_vertex, post_slice,
                     subpop_index = np.argwhere(
                         sorted_info_list['lo_atom'] ==
                         pre_vertex_slice.lo_atom).ravel()[0]
-                    masked_sub_pop_index = subpop_index & (2 ** 9 - 1)
+                    masked_sub_pop_index = \
+                        subpop_index & \
+                        SynapseDynamicsStructuralCommon.BIT_MASK_8_BIT
```
rowleya: As above...

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c
pabogdan: It depend what correct means. The default implementation on spinnaker is and has been for a while to apply a `delay_dendritic` backpropagation delay, however this behaves incorrectly for autapses. I believe Steve wanted us to resolve this somehow.

spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py
pabogdan: yes it is!

spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py
pabogdan: This is correct now (previously a bug). For clarity, the definition of BIT_MASK_X_BIT could be replaced with 0xFF (for 8bit) and 0xFFFF (for 16bit).

spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py
pabogdan: Copy paste:
`This is correct now (previously a bug). For clarity, the definition of BIT_MASK_X_BIT could be replaced with 0xFF (for 8bit) and 0xFFFF (for 16bit).`

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c
rowleya: I am looking back at the description in our paper, and trying to work out which is the thing we do.  Is the problem that if delay_dendritic is used instead of 0, STDP never has any effect on the autapse?  Or is it that the autapse is always depressed?

Additionally, if we use 0 here, what effect does this have on STDP on non-autapses?

rowleya: I think I remember now that you want autapses to be depressed, because a spike can never have caused itself.  I guess our current code disallows this so autapses always continue to exist... I will have to think if there is an easy solution to this.

pabogdan:
We still have some todos:

- [ ] Check that the data types match between parameters in Python and C
- [ ] Modularise C (e.g. random partner for formation or last neuron to have spiked partner for formation)
- [ ] Modularise Python (as above)

Will try to look at it a bit more and see if I can also add the possibility for delays drawn from a uniform distribution.

spynnaker/pyNN/models/spike_source/spike_source_poisson_vertex.py
```diff
@@ -230,30 +359,8 @@ def create_machine_vertex(
             constraints, label)
 
     @property
-    def rate(self):
-        return self._rate
-
-    @rate.setter
-    def rate(self, rate):
-        self._rate = utility_calls.convert_param_to_numpy(rate, self._n_atoms)
-
```
andrewgait: Isn't it supposed to be possible to do e.g. population.set(rate=100) from a top-level PyNN script?  Doesn't this remove that possibility?  (see e.g. https://github.com/SpiNNakerManchester/sPyNNaker8/blob/master/p8_integration_tests/test_change_neuron_parameters_between_runs/test_change_parameter.py)

pabogdan:
I am now confident that the current changes (sans back propagation delays) function correctly for topographic map formation, MNIST classification and motion detection. There are still some restrictions for the current model (interpret these as TODOS):

- [ ] Layer size limit (when populations are too big there are too many pre-synaptic vertices which will generate information which might not fit in DTCM). Filtering edges based on distance would help here.

- [ ] the parameters sigma form ff and lat  can't be too big (we generate LUTs of probabilities based on them)

- [ ] can't use arbitrary spaces -- currently only limited to integer locations on a rectangular grid

- [ ] on-chip synapse expander should be adjusted to deal with structural plasticity (i.e. ensure it adds padding to each synaptic row if needed and that it genereates the appropriate data structure for Structural Plasticity)

I might add some items later, but this is the functionality that I can think of immediately.

rowleya:
Replaced with #687.  To be deleted once this has been merged...

pabogdan:
https://github.com/SpiNNakerManchester/sPyNNaker/pull/687 also invalidates this PR.

alan-stokes:
brnach to be deleted? or are we storing it for future for some unknown reason?

PR #493, opened 2018-03-27
Unbreak setup

dkfellows:
The setup script was badly broken, as it depended on being called in a very precise way that wasn't very useful even for the one script that used it. It also pretended to be an executable plain shell script, which wasn't sensible at all. This fixes all of that stuff.

References
-
* https://stackoverflow.com/a/246128/301832
* https://stackoverflow.com/a/29835459/301832


neural_modelling/setup.bash
```diff
@@ -0,0 +1,10 @@
+# THIS IS A BASH SCRIPT INTENDED TO BE SOURCED!
+
+# Setup script to define the location of the files and the required binaries
+
+set -a
+
+# Define the location of this library
+export NEURAL_MODELLING_DIRS="$( CDPATH= cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
rowleya: set -a does this, but we could instead remove set -a...

PR #494, opened 2018-03-27
Array connector

andrewgait:
Work to support using an ArrayConnector on the toolchain; basically, as simple as it sounds: the user supplies a 2D array of neuron IDs specifying the connections and this is simply passed into the synaptic blocks as required.

### DIFF
--- a/spynnaker/pyNN/models/neural_projections/connectors/__init__.py
+++ b/spynnaker/pyNN/models/neural_projections/connectors/__init__.py
@@ -1,5 +1,6 @@
 from .abstract_connector import AbstractConnector
 from .all_to_all_connector import AllToAllConnector
+from .array_connector import ArrayConnector
 from .distance_dependent_probability_connector \
     import DistanceDependentProbabilityConnector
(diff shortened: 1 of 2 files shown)


andrewgait:
I ought also to have said in the description that although this sounds straightforward, it actually gives us a basis on which to set up the CSAConnector, which will be the next one I look at.  (CSA: Connection Set Algebra, there's a paper on it that I've read and a Python implementation which can be easily pip-installed and from what I've checked today, the results of using it can be quite easily converted into an array and fed through an ArrayConnector).

andrewgait:
Just realised that this is a really basic version of the array connector which only works when the array is the same size as the number of (pre)-neurons... I'll need to do some more work to make this work for a general array.

alan-stokes:
oh well, new pr it be then i guess.

andrewgait:
Yeah - no problem, it's good to have something to work from anyway.

andrewgait:
New PR is at: https://github.com/SpiNNakerManchester/sPyNNaker/pull/496

PR #495, opened 2018-03-27
Fix poisson spike source on resume

oliverrhodes:
This fixes an issue with the Poisson spike source, where on resume all source neurons have their time-to-spike set to zero, and hence all spike during the first timestep. This can flood the network with spikes leading to dropped packets, spike buffer overflow, and ring buffer saturation.

@Andrew, please don't merge this yet - the fix works, but I'm not sure if this is masking a different/additional issue with the reading and writing of the Poisson parameters on resume.


rowleya:
Having looked at this now, I have found a way to fix this through SpiNNFrontEndCommon.  See SpiNNakerManchester/SpiNNFrontEndCommon#306 for a fix.

oliverrhodes:
Unfortunately this still appears to be a problem with the SpikeSourcePoisson.

When running a simulation with a SpikeSourcePoisson which changes rate three times: [0, 6, 0], on starting the period of rate=6 Hz, all Poisson neurons fire, indicating they have all had their time_to_spike set to zero. 

For the rate changes: [1, 6, 1], this behaviour is not observed indicating a corner case when starting from a rate of 0.

The code in this PR to recalculate time_to_spike on resume fixes this issue, however may be masking reloading of incorrect data.

It does, however, improve the correctness of the rate representation on the neuron population, as it forces the Poisson neurons to update their time_to_spike immediately. This otherwise happens on next firing, which could be some time in the future if switching from a low rate.

rowleya:
Just realised that I have introduced a multicore bug in this - going to fix it now...

rowleya:
OK, fixed again - the issue was that the same seed was used on every core of a given source.  This has now been resolved.

rowleya:
My own testing of this shows that it works now, so merging; we can always open a new issue if it causes any further issues.

PR #498, opened 2018-03-29
Moved to use DSG more closely to avoid having code in 2 places

rowleya:
This relies on SpiNNakerManchester/SpiNNFrontEndCommon#306.


oliverrhodes:
This fixes the issue I saw with the Poisson Spike source: data is no longer re-written after a run, when before the run the rate was changed. 

@alan-stokes  please merge when you've reviewed

Issue #492, opened 2018-03-26
ExternalDeviceLifControl population sends membrane potentials as int instead of a S1615


georghinkel:
Neurons using the neuron model above are supposed to regularly send their membrane potentials as a multicast command packet with payload where the payload equals to the current membrane potential. They do that, but the membrane potential is sent as an integer instead of an S1615. This loses a lot of precision and sometimes drops the whole point of these devices as these membrane potentials often fall below 1.

alan-stokes:
Hi George. 

to understand why this is the case, you need to revisit what the ExternalDeviceLifController was originally designed to interface with. aka a push bot. The values that the lab manual for its protocol states are in the range of 1 to 100, and so the controller converts to that format. It is not hard to duplicate the code and replace the int with a s1615, or even better, augment the code to switch format types based off init parameters, but currently this is not implemented as noone has asked. 

I recommend you augment the ExternalDeviceLifController c and python to do the switch and produce a pull request for us to review, as us getting to it may take longer than yourself. 

regards
Alan

rowleya:
One thing to note is that the output value is between 1 and 100, but the internal representation is a fixed point value - thus it is just chopped off at the end.  I don't think you can just change this, as existing devices might not be compatible with the fixed point format; instead it might be good to have a format flag which would indicate whether conversion is desired.  Alternatively, fixed-point numbers are treated as integers internally - therefore using larger numbers and then just treating the value as fixed-point would also work (using a power of 2 multiplier for example).

andrewgait:
The fix for this has been merged, so closing this issue.

PR #500, opened 2018-04-03
Add option to ExternalLifNeurons to shift the value range of payloads

georghinkel:
This is to implement #492:

> Neurons using the neuron model above are supposed to regularly send their membrane potentials as a multicast command packet with payload where the payload equals to the current membrane potential. They do that, but the membrane potential is sent as an integer instead of an S1615. This loses a lot of precision and sometimes drops the whole point of these devices as these membrane potentials often fall below 1.

The way this is implemented is to interpret the integer value `value_as_payload` as a scaling factor instead of a boolean: If this scaling factor is 0, then no payload is required. In other cases, the value depicts an integer scaling factor. This scaling factor is set to a default of 1 in the Python API in order to give the equivalent behavior for existing code.

Unfortunately, I have no idea whether there are unit tests for the neuron models. Therefore, I did not alter any tests.


georghinkel:
The build seems to fail because of the following lines:

> The command "flake8 examples spynnaker pyNN-spiNNaker-src" exited with 1.
> flake8 unittests

However, all I can see above is that all tests still pass. Do you have an idea why Travis failed?

Christian-B:
After running the unittest travis does pep8 style checking using a program called flake8.

As the Travis output (reachable by the details link above) shows

**$ flake8 examples spynnaker pyNN-spiNNaker-src
spynnaker/pyNN/external_devices_models/threshold_type_multicast_device_control.py:54:80: E501 line too long (98 > 79 characters)**

length lines is one of many restrictions in pep8

georghinkel:
I corrected the line length and fixed the test coverage problem. Is there anything else I need to do to get this merged?

andrewgait:
I don't think so - it just needs one of us to review it and merge.  Someone should get to it in the near future...

PR #501, opened 2018-04-06
No dots in include

Christian-B:
Got rid of all or most of the #include ".. UGLINESS!


neural_modelling/src/neuron/builds/neural_build.mk
```diff
@@ -131,12 +135,15 @@ $(foreach obj, $(STDP), $(eval $(call stdp_rule, $(obj))))
 $(WEIGHT_DEPENDENCE_O): $(WEIGHT_DEPENDENCE) $(SYNAPSE_TYPE_H)
 	-mkdir -p $(dir $@)
 	$(CC) -D__FILE__=\"$(notdir $*.c)\" -DLOG_LEVEL=$(PLASTIC_DEBUG) $(CFLAGS) \
+	        $(INCLUDE_PLASTICITY_HEADERS) \
 	        -include $(SYNAPSE_TYPE_H) -o $@ $<
 
 $(TIMING_DEPENDENCE_O): $(TIMING_DEPENDENCE) $(SYNAPSE_TYPE_H) \
                         $(WEIGHT_DEPENDENCE_H)
+	echo TIMING_DEPENDENCE_O
```
dkfellows: ????

neural_modelling/src/delay_extension/Makefile
```diff
@@ -3,3 +3,5 @@ BUILD_DIR = build/
 SOURCES = delay_extension.c
 
 include ../common.mk
+
+CFLAGS += -I $(NEURAL_MODELLING_DIRS)/src/common
```
dkfellows: I believe that there shouldn't usually be a space between `-I` and the directory

neural_modelling/src/neuron/c_main.c
```diff
@@ -15,13 +15,13 @@
  *
  */
 
-#include "../common/in_spikes.h"
+#include <in_spikes.h>
 #include "neuron.h"
```
dkfellows: Surely we should be consistent here and use `#include <…>` for all of these?

neural_modelling/src/neuron/builds/neural_build.mk
Christian-B: debug gone

neural_modelling/src/delay_extension/Makefile
Christian-B: Our pattern is to include the space

See:  lines 54 and 55 in
https://github.com/SpiNNakerManchester/spinnaker_tools/blob/master/make/Makefile.common

The space may not be required but that is out of scope for this PR

neural_modelling/src/neuron/c_main.c
Christian-B: The difference is that <in_spikes.h> is brought in by an -I ... while "neuron.h" is local to c_main.c

May be possible/ cleaner to do all as <...> but that is also out of scope of this PR

PR #502, opened 2018-04-13
CSA connector

andrewgait:
Implementation of a Connection Set Algebra (Djurfeldt 2012) connector in sPyNNaker.

For information on the python implementation of CSA see https://github.com/INCF/csa

This implementation does not at present do any work related to statistical analysis or probabilities relating to calculating the possible maximum connections as some other connectors do; instead, it "assumes the worst" may happen (i.e. an all-to-all connector).  This is likely to be something that we need to consider in the future, or it can be debated on this PR if that seems more appropriate.

### DIFF
--- a/spynnaker/pyNN/models/neural_projections/connectors/csa_connector.py
+++ b/spynnaker/pyNN/models/neural_projections/connectors/csa_connector.py
@@ -1,5 +1,6 @@
 from spinn_utilities.overrides import overrides
 from .abstract_connector import AbstractConnector
+
 import logging
 import numpy
 import csa
@@ -26,7 +27,8 @@ class CSAConnector(AbstractConnector):
         super(CSAConnector, self).__init__(safe, verbose)
         self._cset = cset
         # think this is probably needed here
-        self._arrays_set = False
+        self._pair_list = None
+        self._full_connection_set = None
 
     @overrides(AbstractConnector.get_delay_maximum)
     def get_delay_maximum(self):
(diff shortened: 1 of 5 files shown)


alan-stokes:
looks good to me

spynnaker/pyNN/models/neural_projections/connectors/csa_connector.py
```diff
@@ -0,0 +1,159 @@
...
+        post_hi = post_vertex_slice.hi_atom
+
+        # this is where the magic needs to happen somehow
+        if self._full_cset is None:
+            self._full_cset = [x for x in csa.cross(
+                range(self._n_pre_neurons),
+                range(self._n_post_neurons)) * self._cset]
+
+        # use CSA to cross the range of this vertex's neurons with the cset
+        self._pair_list = csa.cross(
```
rowleya: Don't think this needs to be stored since it is calculated at every call - it could instead be returned

spynnaker/pyNN/models/neural_projections/connectors/csa_connector.py
```diff
@@ -0,0 +1,159 @@
...
+            range(pre_lo, pre_hi+1),
+            range(post_lo, post_hi+1)) * self._full_cset
+
+        # Get the lists of pre and post neurons that result from this
+        self._pre_neurons = [x[0] for x in self._pair_list]
+        self._post_neurons = [x[1] for x in self._pair_list]
```
rowleya: I think this and self._pre_neurons could be returned rather than stored as they are calculated at each call

spynnaker/pyNN/models/neural_projections/connectors/csa_connector.py
```diff
@@ -0,0 +1,159 @@
...
+
+        if self._verbose:
+            print('full cset: ', self._full_cset)
+            print('this vertex pair_list: ', self._pair_list)
+            print('this vertex pre_neurons: ', self._pre_neurons)
+            print('this vertex post_neurons: ', self._post_neurons)
+
+        n_connections = len(self._pre_neurons)  # size of the array created
```
rowleya: Should this be the size of the pair list?

spynnaker/pyNN/models/neural_projections/connectors/csa_connector.py
andrewgait: Agreed - done.

spynnaker/pyNN/models/neural_projections/connectors/csa_connector.py
andrewgait: Don't even need to be returned as they can be obtained from pair_list whenever required - done.

spynnaker/pyNN/models/neural_projections/connectors/csa_connector.py
andrewgait: pair_list and pre_neurons are the same size, but as pre_neurons has been removed I have changed this too.

andrewgait:
Note: there is an error in the CSA code when using python 3, which I have detailed here (with a fix): https://github.com/INCF/csa/issues/10

rowleya:
I guess there isn't much we can do about the error here - as this is a separate connector, I see no reason not to merge this :)

andrewgait:
If you pip(3)-install csa into a location you can easily edit (as I have done for now - I've been trying to keep python2 and python3 as separate as possible), then it's fine - you can just edit it as described in that issue.

Issue #437, opened 2017-11-24
speed-up FromListConnector


chanokin:
When the lists are large searching for incomming connections takes a long time because we're searching the entire list every time [(line 220)](https://github.com/SpiNNakerManchester/sPyNNaker/blob/63f73083d81febad18bdc08a8db09fab5144e3ef/spynnaker/pyNN/models/neural_projections/connectors/from_list_connector.py#L220)

Would it be safe to delete the items we already found in the list after parsing them to the synaptic block? This would reduce the search space every time a vertex slice is processed, thus reducing the time to process it.

alan-stokes:
issue is youd need to clone the connection list before you could do anything. The end user gave you that object, changing it would be a breach of api (just imagine a 2 run phase where the end user uses the same list in a new projection and then wonders why its empty). The cloning would take a while, but after that you could. Youd need to measure to see how it performs. and just ebcuase we asked once, does not mean we wont ask again, the partitioner is a tril and error process.

rowleya:
I think that we clone the list into a numpy array anyway to make the whole thing faster.  If the connections are then removed from this list at the point when the synaptic block was generated, that should be faster, although as Alan points out, there isn't anything in there that says it only has to be done once...

Other solutions would be to look at why the numpy array is so slow in the first place, which I think is due to multiple filterings of the list.  I would guess that there are solutions that involve writing some C code that more directly accesses and filters the numpy array directly.

andrewgait:
Looking at this code a bit more, it seems to me as though there's a lot of repetition and duplication that could be unnecessary - I may be wrong, but I think it should only be necessary to set the `mask` variable once per block - currently it is set in every function where it is required, which seems like overkill to me.  In other connectors the equivalent function is first set to None, then updated by the first function that calls it, and then after that in this instance no further update of it would be necessary and any other function requiring the mask could just simply ask for it.

andrewgait:
The PR at https://github.com/SpiNNakerManchester/sPyNNaker/pull/632 is related to this - I'd say this issue can probably be closed once that branch is merged.

andrewgait:
PR #632 has been merged so closing this issue.

PR #507, opened 2018-04-26
Update from list connector

andrewgait:
After reading https://github.com/SpiNNakerManchester/sPyNNaker/issues/437 I went and looked at the FromListConnector; there does seem to be quite a bit of repetition and duplication in there.  This attempts to avoid that where possible; things still work on the scripts with small number of neurons that use this connector in PyNN8Examples.  It would be interesting (to me at least) to see if these changes make any differences with larger models... ? @chanokin do you have anything you could test this on?


rowleya:
I believe the recent changes to connectors in general (where we no longer look at the connectivity of subsections of the matrix) make these changes redundant.

andrewgait:
Closing this as it's no longer relevant due to other PRs such as https://github.com/SpiNNakerManchester/sPyNNaker/pull/525.

PR #508, opened 2018-04-26
Dictionary for log messages

Christian-B:
https://github.com/SpiNNakerManchester/spinnaker_tools/pull/75

Major make changes including moves of all makefiles


neural_modelling/src/neuron/synapses.c
```diff
@@ -60,7 +60,7 @@ static inline void _print_synaptic_row(synaptic_row_t synaptic_row) {
 
     for (uint32_t i = 0; i < n_fixed_synapses; i++) {
         uint32_t synapse = fixed_synapses[i];
-        uint32_t synapse_type = synapse_row_sparse_type(synapse);
+        uint32_t synapse_type = synapse_row_sparse_type(synapse, synapse_index_bits);
```
dkfellows: What are the changes in this file doing?

Christian-B: It is a bug fixed found on the way.
The signature changed in:
https://github.com/SpiNNakerManchester/sPyNNaker/commit/7af04a9e59f7b9abbd7574dcd6fc2ddf34023959

but was only discovered when I was able to build with debug on.

PR #524, opened 2018-05-23
on no spikes returns an empty shaped array

Christian-B:
Actual fix for bug discovered by @SimonDavidson 

Return an empty array with the correct shape


Christian-B:
This PR is independent of https://github.com/SpiNNakerManchester/sPyNNaker8/pull/147 this fixes the issue the other is now a safety patch.

Issue #519, opened 2018-05-17
Connector cannot be re-used (really sPyNNaker8 but code is here)


rowleya:
In PyNN, a connector should be able to be reused in multiple connections.  In PyNN 7, the weight and delay are part of that connection, but in PyNN 8 they are not.  The current implementation sets the weights and delays in the connection when used from PyNN 8, but this means that two connections that use the same connector but different weights and delays will not work; the weights and delays get overwritten.

Example script:
```
p.setup()
inp = p.Population(1, p.SpikeSourceArray(spike_times=[]))
out = p.Population(1, p.IF_curr_exp())

connector = p.OneToOneConnector()

proj_1 = p.Projection(inp, out, connector, p.StaticSynapse(weight=1.0, delay=1.0))
proj_2 = p.Projection(inp, out, connector, p.StaticSynapse(weight=2.0, delay=2.0))

p.run(1)

print proj_1.get(("weight", "delay"), "list")
print proj_2.get(("weight", "delay"), "list")

p.end()
```

alan-stokes:
drat, this is a rather difficult one to solve, till we disconnect from 0.75. 

at moment, i think we'd need to clone the connector (to sticky plaster it), as our low level stuff still assumes weight and delay are in the connector, as they were in 0.75. 

the proper fix would be to remove weights from the connector totally, but thats going to have massive ramifications to the rest of the python, and would require a pynn 0.75 to be either rewritten to accommodate the move, or dropped. 

im guessing we could but a safety check in the connection for the time being, to stop users using it multiple times with different weights / delays.

rowleya:
I think the answer to this really is to remove the weights and delays from the connector, and instead use the PyNN8 concept of having a synapse type which contains these, and which is then passed to the connector at the point it needs them.  This would also involve moving other appropriate methods out of the connector and into the synapse type.  I don't think this is too difficult, though it does involve a fair bit of code mangling.

Note that for PyNN 7, you would then do something like the opposite of what we are currently doing for PyNN 8 i.e. split the given connector into a connector and synapse type.

alan-stokes:
yep. which is why its a rather big change. not something for an afternoon. lol. 
thus why i said maybe in the short term, a safety check can be put in until we get a few days to fix it fully.

andrewgait:
Started looking at this in https://github.com/SpiNNakerManchester/sPyNNaker/pull/523 - have added safety check, will look at fixing it more fully when I have time to do so.

[...]

PR #525, opened 2018-05-24
Move weights and delays

andrewgait:
Further work on https://github.com/SpiNNakerManchester/sPyNNaker/issues/519 - with a little bit of hacking this appears to work for both PyNN 0.7 and 0.8+. More testing is probably required.


spynnaker/pyNN/models/neuron/synapse_io/synapse_io_row_based.py
```diff
@@ -62,6 +62,13 @@ def get_sdram_usage_in_bytes(
             machine_time_step, in_edge):
         # pylint: disable=too-many-arguments, too-many-locals, arguments-differ
 
+        # Tell the connector the weights and delays it needs to use
```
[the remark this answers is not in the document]
andrewgait: To some extent the connector is going to have to have weights at any one time depending upon which projection it’s encapsulated in.  But I agree that perhaps it’s better if the weights and delays are stored in the synapse type and not the connector if at all possible; I’ll have another look at this and see if it’s possible to somehow access the correct synapse type depending upon the projection the connector is encapsulated in...

rowleya:
I think this is heading in the right direction, but I would advise against storing either delay or weight in the connector, including setting it as a side effect.  Any method that needs either or both can be given them from outside.

spynnaker/pyNN/models/neural_projections/connectors/from_list_connector.py
```diff
@@ -105,22 +142,22 @@ def get_n_connections_to_post_vertex_maximum(self):
         return numpy.max(numpy.bincount(self._conn_list["target"]))
 
     @overrides(AbstractConnector.get_weight_mean)
-    def get_weight_mean(self):
+    def get_weight_mean(self, weights):
```
rowleya: This needs to do the check, but for weights.

spynnaker/pyNN/models/neural_projections/connectors/from_list_connector.py
```diff
@@ -105,22 +142,22 @@ def get_n_connections_to_post_vertex_maximum(self):
...
         return numpy.mean(numpy.abs(self._conn_list["weight"]))
 
     @overrides(AbstractConnector.get_weight_maximum)
-    def get_weight_maximum(self):
+    def get_weight_maximum(self, weights):
```
rowleya: This can do the check again...

spynnaker/pyNN/models/neural_projections/connectors/from_list_connector.py
```diff
@@ -105,22 +142,22 @@ def get_n_connections_to_post_vertex_maximum(self):
...
         # pylint: disable=too-many-arguments
         return numpy.amax(numpy.abs(self._conn_list["weight"]))
 
     @overrides(AbstractConnector.get_weight_variance)
-    def get_weight_variance(self):
+    def get_weight_variance(self, weights):
```
rowleya: Another case of the check being required.

spynnaker/pyNN/models/neural_projections/connectors/from_list_connector.py
```diff
@@ -105,22 +142,22 @@ def get_n_connections_to_post_vertex_maximum(self):
...
         # pylint: disable=too-many-arguments
         return numpy.var(numpy.abs(self._conn_list["weight"]))
 
     @overrides(AbstractConnector.create_synaptic_block)
     def create_synaptic_block(
-            self, pre_slices, pre_slice_index, post_slices,
+            self, weights, delays, pre_slices, pre_slice_index, post_slices,
```
rowleya: Final place where the check is required, this time for both weights and delays.  This will have to generate the weights and delays for the block, though this might just use the internal lists.

spynnaker/pyNN/models/neuron/synapse_io/synapse_io_row_based.py
```diff
@@ -61,17 +61,25 @@ def get_max_row_info(
...
                 max_delay_supported + numpy.finfo(numpy.double).tiny)
 
+        # get delays to pass into connector
+        delays = None
+        if hasattr(synapse_info.synapse_dynamics, 'delay'):
+            delays = synapse_info.synapse_dynamics.delay
+        else:
+            # delays come directly from the connector?
+            delays = synapse_info.connector.get_delay()
+
```
rowleya: I think this could just now use synapse_info.delay?  I don't see that the dynamics needs to be involved...  remember these are the actual delay values, not theoretical maximums or anything else that the synapse dynamics would provide.

spynnaker/pyNN/models/neuron/synapse_io/synapse_io_row_based.py
```diff
@@ -186,14 +194,27 @@ def get_synapses(
...
+        if hasattr(synapse_info.synapse_dynamics, 'weight'):
+            weights = synapse_info.synapse_dynamics.weight
+        else:
+            weights = synapse_info.weight
+
+        if hasattr(synapse_info.synapse_dynamics, 'delay'):
+            delays = synapse_info.synapse_dynamics.delay
+        else:
+            delays = synapse_info.delay
+
```
rowleya: Again, I am not sure why the synapse dynamics is involved here...

spynnaker/pyNN/models/neuron/synaptic_manager.py
```diff
@@ -425,23 +427,34 @@ def _get_ring_buffer_to_input_left_shifts(
...
+                    if hasattr(synapse_dynamics, 'weight'):
+                        weights = synapse_dynamics.weight
+                    else:
+                        weights = synapse_info.weight
+
+                    delays = None
+                    if hasattr(synapse_dynamics, 'delay'):
+                        delays = synapse_dynamics.delay
+                    else:
+                        delays = synapse_info.delay
```
rowleya: Again, not sure why synapse dynamics is involved here...

spynnaker/pyNN/models/pynn_projection_common.py
```diff
@@ -62,15 +62,26 @@ def __init__(
...
-            connector, synapse_dynamics_stdp, synapse_type)
+        # Set and store synapse information for future processing
+        if hasattr(synapse_dynamics_stdp, 'weight'):
+            self._synapse_information = SynapseInformation(
+                connector, synapse_dynamics_stdp, synapse_type)
+        else:
+            self._synapse_information = SynapseInformation(
+                connector, synapse_dynamics_stdp, synapse_type,
+                connector.get_weight(), connector.get_delay())
+
```
rowleya: Rather than doing this here, I think this function should take a synapse type (not the same as the existing synapse type!), which can provide the weights and delays.  PyNN8 should provide this anyway...

spynnaker/pyNN/models/neuron/synapse_io/synapse_io_row_based.py
andrewgait: Looking at this a bit more, there's some strange naming going on here.  Basically in this function, synapse_dynamics_stdp is the "synapse_type" that was passed in by the user in a Projection.  I guess it's called that here (it's called "synapse_type" in the sPyNNaker8 Projection class) to avoid confusion with the synapse_type in the neuron model... anyway, this isn't really using the dynamics, it's essentially copying in what the user has defined in their Projection.

andrewgait: I should clarify a bit further: here, it's called "synapse_dynamics", but at the top end where this is passed in (in pynn_projection_common) it's called "synapse_dynamics_stdp".  We should possibly come up with a whole different name instead I guess...

spynnaker/pyNN/models/neural_projections/connectors/abstract_connector.py
```diff
@@ -67,7 +67,7 @@ def set_space(self, space):
         """
         self._space = space
 
-    def _set_weights_and_delays(self, weights, delays, allow_lists):
+    def set_weights_and_delays(self, weights, delays):
```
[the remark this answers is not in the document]
andrewgait: I actually think this function is necessary - I think the confusion has arisen because the merge with master didn't work properly, so every other function had delays / weights as an argument, when in fact in master they don't use them any more.  But it's necessary for a different reason now - it's the function that's being used to pass in the weights and delays from the synapse information in each case.  (And in the case of the FromListConnector, making sure that the weights and/or delays supplied by the user in their list don't get overwritten).

spynnaker/pyNN/models/neural_projections/connectors/from_list_connector.py
```diff
@@ -38,52 +37,90 @@ def __init__(self, conn_list, safe=True, verbose=False):
...
+            other_element_column_names.append(column_names[element])
+        if element_index:
+            other_conn_list = conn_list[:, element_index]
+            other_conn_list.dtype.names = other_element_column_names
+
+        # hand over split data
+        return source_destination_conn_list, weights, delays, other_conn_list
+
     @overrides(AbstractConnector.set_weights_and_delays)
     def set_weights_and_delays(self, weights, delays):
```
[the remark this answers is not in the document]
andrewgait: I think there's something in this that I still need to work out - namely that a user could supply only weights, or only delays (assuming that's allowed by PyNN, which does seem to be the case).  I actually worked on something similar to this in a later PR, so this could be dealt with in that PR (see https://github.com/SpiNNakerManchester/sPyNNaker8/pull/188 and https://github.com/SpiNNakerManchester/sPyNNaker/pull/586) rather than in this one?

spynnaker/pyNN/models/neural_projections/connectors/abstract_connector.py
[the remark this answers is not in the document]
rowleya: I have looked back at this in previous iterations.  It doesn't look like connectors in the master version ever took in the weight or delay parameters, so I think these have been added here...

spynnaker/pyNN/models/neural_projections/connectors/from_list_connector.py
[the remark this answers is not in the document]
rowleya: The above fix should work in general since you can check for weights and delays separately.  Basically, if you store the weights and/or delays from the FromListConnector constructor, you can then check if they have been provided when any functions are called, and use the ones provided in the list if you have them.

spynnaker/pyNN/models/neural_projections/connectors/abstract_connector.py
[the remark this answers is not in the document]
andrewgait: Now I’m confused - I’m sure at least some of the functions here did take weights and/or delays as parameters - (_get_delay_maximum(..) and _get_weight_maximum(..) certainly did...), but looking back it does look as though I added a number of them on this branch.  I’m of the opinion that using the set_weights_and_delays(..) function here to send the weights and delays in is a better way of doing it than passing them into the functions, but if you have an opinion as to which is best then we can go with that...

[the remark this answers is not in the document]
andrewgait: (And of course overriding this function in the FromListConnector as required...)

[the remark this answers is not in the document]
rowleya: The underscore functions did and do, as these are static utility functions in the abstract class.  The other functions didn't but probably should - this is what is needed to avoid storing the weights and delays in the classes, which is needed to allow the connector to be used in multiple projections.  This *could* be done by requiring that each call to the functions is preceded by a call to set_weights_and_delays but this seems error prone, as it will only take one missed call to make it all fall over.

PR #530, opened 2018-06-15
Synapse expander on chip

rowleya:
This is the on-chip synaptic expander.  The main body of this work has been to do this expansion on chip, but a couple of other things have gone in here too, that all contribute to synaptic expansion being done as efficiently as possible.  This includes:
1. Modification of the synaptic manager to only calculate row lengths and weight scaling once per post-machine-vertex rather than once for every pre-post machine-vertex pair.  This speeds things up fairly drastically.
1. Modification of all connectors in response to the above.  This has simplified a lot of the connectors.
1. Update of the master population table to be 16-word aligned rather than 1-word aligned.  This allows for the table to use any amount of SDRAM, which allows up to 255 neurons per core on the microcortical column.
1. Update to STDP to change how generation works.  This was needed for the generator.  In the original version, the data was written by the synapse structure.  In this version the structure just says how many half-words it contains and which half-word is the data.  NOTE: Future change is to also allow the header half-words to be specified rather than all 0s.  This has not yet been implemented here.
1. A couple of bug fixes for reading back of data, and for the use of direct matrices, which are now supported again.

A couple of things that are still to do, but do not stop this being merged:
1. The STDP changes mentioned above.
1. Add Gary's connection generation for convolution networks.


neural_modelling/src/synapse_expander/delay_expander.mk
```diff
@@ -0,0 +1,7 @@
+APP = delay_expander
+BUILD_DIR = build/
+SOURCES = rng.c param_generator.c connection_generator.c delay_expander.c
+
+include ../common.mk
+
```
andrewgait: I think @Christian-B was attempting to avoid the use of ../ in makefiles in a previous PR - can it be done here?

spynnaker/pyNN/models/neural_projections/connectors/array_connector.py
```diff
@@ -28,89 +28,45 @@ def __init__(
...
             post_slice_index, pre_vertex_slice, post_vertex_slice,
             synapse_type):
-        n_connections = self._get_n_connections(pre_vertex_slice,
-                                                post_vertex_slice)
+        pre_neurons = self._array[0]
+        post_neurons = self._array[1]
+        mask = ((pre_neurons >= pre_vertex_slice.lo_atom) and
+                (pre_neurons <= pre_vertex_slice.hi_atom) and
+                (post_neurons >= post_vertex_slice.lo_atom) and
+                (post_neurons <= post_vertex_slice.hi_atom))
```
andrewgait: This is certainly much simpler than what I had previously!

spynnaker/pyNN/models/neural_projections/connectors/from_list_connector.py
```diff
@@ -262,8 +157,3 @@ def conn_list(self):
     @conn_list.setter
     def conn_list(self, new_value):
         self._conn_list = new_value
```
andrewgait: This is quite similar to what I tried to do in https://github.com/SpiNNakerManchester/sPyNNaker/pull/507 - perhaps that PR isn't necessary any more.

spynnaker/pyNN/models/neuron/synapse_dynamics/abstract_synapse_dynamics.py
```diff
@@ -65,45 +65,29 @@ def get_delay_maximum(self, connector):
         """
         return connector.get_delay_maximum()
 
-    def get_delay_variance(
-            self, connector, n_pre_slices, pre_slice_index, n_post_slices,
-            post_slice_index, pre_vertex_slice, post_vertex_slice):
+    def get_delay_variance(self, connector):
```
andrewgait: A few things to consider here in conjunction with https://github.com/SpiNNakerManchester/sPyNNaker/pull/525 - but again, not enough to stop this from being merged now.  I'd be happy to do the merge from this PR into that one if necessary.

spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_stdp.py
```diff
@@ -207,20 +213,30 @@ def get_plastic_synaptic_data(
...
-        # Get the plastic data
+        # Get the plastic data by inserting the weight into the half-word
+        # specified by the synapse structure
         synapse_structure = self._timing_dependence.synaptic_structure
-        plastic_plastic = synapse_structure.get_synaptic_data(connections)
-        plastic_headers = numpy.zeros(
-            (n_rows, self._n_header_bytes), dtype="uint8")
+        n_half_words = synapse_structure.get_n_half_words_per_connection()
+        half_word = synapse_structure.get_weight_half_word()
+        plastic_plastic = numpy.zeros(
```
andrewgait: I hadn't noticed this before now, but "plastic_plastic" seems a somewhat unnecessary name - what's wrong with simply using "plastic" here instead?

spynnaker/pyNN/models/neural_projections/connectors/from_list_connector.py
```diff
@@ -47,53 +46,6 @@ def __init__(self, conn_list, safe=True, verbose=False):
         self._delays = None
         self._converted_weights_and_delays = False
 
-    @staticmethod
-    def _split_conn_list(conn_list, column_names):
-        """ takes the conn list and separates them into the blocks needed
```
andrewgait: Taking out this function leads to an error as it is called from the sPyNNaker8 from_list_connector; it might be best to move this function up there rather than keeping it here... again it depends somewhat on how much PyNN 0.7 support we want to bother with.

andrewgait:
To confirm, the recent edits I've made on this now mean that the tests with single-neuron STDP models now pass - so this is good for merging as far as I'm concerned.  Might be worth running the integration tests again to be certain...

rowleya:
The issue is not that the CSA connector has not been tested, but that the changes in this branch were done before CSAConnector was merged - thus it also needs to be changed to meet the new interface

andrewgait:
OK, done for now for this branch - will open a new branch for making csa / csa-plot optional.

PR #535, opened 2018-07-23
tau_refrac_timesteps is an integer

andrewgait:
Unit tests in sPyNNaker8 (at least) were failing in python3, as numpy.ceil returns a float and tau_refrac_timesteps is an integer.


andrewgait:
(I ought to say too that it wasn't just the unit tests that were failing - most examples were failing to run at all.)

PR #538, opened 2018-08-08
Remove this code as it can wrap and do strange things when T1 low

oliverrhodes:


alan-stokes:
hi oliver. 

those liens are used to spread the firing over a timer tick so that it doesn't peak and cause dropped packets over many neurons on many cores fire. 

Maybe the code needs to be adjusted so that it handles the wrap, but removing it will cause other issues.

rowleya:
I believe that this has been fixed in #539, so closing this.

PR #539, opened 2018-08-09
Updated to avoid issues of wrap around of timer

rowleya:
This should avoid additional delays when the timer tick is close to over-running (or has overrun).  It could create dropped packets in those circumstances (but then we are supposed to avoid those circumstances anyway).


rowleya:
This could do with testing on any known error cases where the timer has wrapped in the past...

PR #540, opened 2018-08-22
Fix plastic target neuron mask for getting weights

oliverrhodes:


spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_stdp.py
```diff
@@ -282,9 +282,10 @@ def read_plastic_synaptic_data(
             data_fixed.size, dtype=self.NUMPY_CONNECTORS_DTYPE)
         connections["source"] = numpy.concatenate(
             [numpy.repeat(i, fp_size[i]) for i in range(len(fp_size))])
-        connections["target"] = (data_fixed & 0xFF) + post_vertex_slice.lo_atom
+        connections["target"] = ((data_fixed & (n_neuron_id_bits+1))
+                                 + post_vertex_slice.lo_atom)
         connections["weight"] = synapse_structure.read_synaptic_data(
-            fp_size, pp_without_headers)
+            pp_size, pp_without_headers)
```
andrewgait: I'm not sure this is right - for a start, it fails the tests in https://github.com/SpiNNakerManchester/sPyNNaker8/tree/master/p8_integration_tests/test_stdp - but I'm not completely sure what the correct fix is.  It looks like perhaps the connections array needs to be initialised differently above (using e.g. data_plastic instead of data_fixed) ?

PR #542, opened 2018-09-07
Add required package

dkfellows:
NB: `csa` should be in `setup.py` as well as `requirements.txt`; this was making the `master` build of sPyNNaker7 fail


andrewgait:
There's also a fail on most if not all Travis builds at the moment due to something related to pytest_cov and coverage, but I'm not sure how that should be resolved...

PR #543, opened 2018-09-07
Citiation

alan-stokes:
adds citation and manifest files. start point for further work on creating a doi for the tools


dkfellows:
You had syntax errors in your YAML. I fixed them for you and redid the `address` to match about what https://citation-file-format.github.io/1.0.3/specifications/ specifies.

dkfellows:
How can we know if the contents of the `MANIFEST.in` file are correct? When I try running `python setup.py -n sdist`, I get these warnings (in amongst a _lot_ of output):
```
warning: sdist: standard file not found: should have one of README, README.rst, README.txt
warning: check: missing meta-data: either (author and author_email) or (maintainer and maintainer_email) must be supplied
```
(I also get an error, but that's probably because I used the `-n` — dry-run — option. That doesn't bother me.)

alan-stokes:
thanks on the fixes. the latest error is this weird dependnacy thing. but it also says vera is reporting errors. it seems vera isnt running on thie machine so cant test. but as ive not touched the c code......

alan-stokes:
deleting as wrong branch name

PR #549, opened 2018-10-08
Reorder variables to match C code

oliverrhodes:
There is a misalignment between the C code struct, and what the Python code writes, meaning you end up with incorrect neural dynamics.

Also the global time step which gets written is too big, so you get strange behaviour after the first spike.

### DIFF
--- a/spynnaker/pyNN/models/neuron/neuron_models/neuron_model_izh.py
+++ b/spynnaker/pyNN/models/neuron/neuron_models/neuron_model_izh.py
@@ -25,10 +25,10 @@ UNITS = {
 class NeuronModelIzh(AbstractNeuronModel):
 
     __slots__ = [
-        "_a", "_b", "_c", "_d", "_i_offset", "_v_init", "_u_init"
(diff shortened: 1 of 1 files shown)


spynnaker/pyNN/models/neuron/neuron_models/neuron_model_izh.py
```diff
@@ -78,7 +78,7 @@ def has_variable(self, variable):
     @overrides(AbstractNeuronModel.get_global_values,
                additional_arguments={'machine_time_step'})
     def get_global_values(self, machine_time_step):
-        return [machine_time_step]
+        return [machine_time_step/1000]
```
rowleya: This should probably be float(machine_time_step) / 1000.0, as machine_time_step might be < 1000

alan-stokes: should also remove magic number and replace with a named constant. Had a little look around our constants. Seems we're missing that constant. so likely needs a few places to change

PR #565, opened 2018-10-30
Buffer cleanup

Christian-B:
Updates scripts to get data by placement and directly from BufferedReceivingData

Must be done at the same time as: SpiNNakerManchester/SpiNNFrontEndCommon#363

EDIT: must be at same time as SpiNNakerManchester/SpiNNFrontEndCommon#372


dkfellows:
Related to SpiNNakerManchester/SpiNNFrontEndCommon#370 perhaps?

Issue #566, opened 2018-11-06
Incorrect edge pruning 


pabogdan:
The presence of a One To One Connector between two populations can cause undesired effects if an additional projection exists between the same two population with a different connector (e.g. All To All). In the script below I have a Spike Source Array (SSA) and an IF_curr_exp, each with 2 neurons forced to partition them as 1 / core. I have 2 projections between these 2 populations: 1:1 and all:all. The 1:1 weights is ignorable, the all:all weight causes the post neurons to spike. 
[**DESIRED EFFECT**] a spike from one of the neurons (id 0) in the SSA should cause both post neurons to spike
[**OBSERVED EFFECT**] only the first post neuron (id 0) fires because the other machine edge has been pruned

The bug is caused by https://github.com/SpiNNakerManchester/sPyNNaker/blob/e0aa35fab3646aa5de0a65ef312e734a718614ef/spynnaker/pyNN/models/neural_projections/projection_machine_edge.py#L37-L43

Script location ([**ACTION**] needs wrapping as a test and inclusion into integration test batch): https://github.com/pabogdan/neurogenesis/blob/sp_random_delays/sundry/pynn8_integration_test_one_to_one_connector.py

[**WRONG**] target pop spikes with 1:1 connector: `[[0. 2.]]` (ignore small bug in plotting, id should read 0 rather than 1)
![wrong_spikes_one_to_one_connector](https://user-images.githubusercontent.com/1470639/48061097-bfeb5000-e1b5-11e8-970f-3cb81900c1a2.png)

[**CORRECT**] target pop spikes without 1:1 connector: `[[0. 2.] [1. 2.]]`

![correct_spikes_one_to_one_connector](https://user-images.githubusercontent.com/1470639/48061100-c4176d80-e1b5-11e8-8d6b-655523ebdcba.png)

andrewgait:
Note: the behaviour described here was observed using the release version; against current master, with both connectors in place, neither of the neurons in target population spike.

andrewgait:
The fix for this has been merged, and it also seems unaffected by other recent changes, so closing this.

PR #567, opened 2018-11-06
Fix so that multiple connectors on the same edge behave correctly

andrewgait:
Fixing issue https://github.com/SpiNNakerManchester/sPyNNaker/issues/566 - @rowleya this affects the synapse expander a bit in that I think we have to remove the direct matrix option... ?


rowleya:
I believe that the changes in #576 fix the issue here where direct row processing is not working.  I will therefore wait to merge this until #576 is done...

PR #568, opened 2018-11-10
Fixes to allow Poisson to run forever

rowleya:
Simple changes that allow the Poisson source to run forever

### DIFF
--- a/spynnaker/pyNN/connections/spynnaker_poisson_control_connection.py
+++ b/spynnaker/pyNN/connections/spynnaker_poisson_control_connection.py
@@ -8,6 +8,7 @@ from data_specification.enums import DataType
 from spinn_front_end_common.utilities.connections import LiveEventConnection
 from spinn_front_end_common.utilities.exceptions import ConfigurationException
 from spinn_front_end_common.utilities.constants import NOTIFY_PORT
+from decimal import Decimal
 
 _MAX_RATES_PER_PACKET = 32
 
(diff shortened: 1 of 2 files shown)


spynnaker/pyNN/connections/spynnaker_poisson_control_connection.py
```diff
@@ -100,7 +101,8 @@ def _assemble_message(id_to_key_map, neuron_id_rates, pos):
         for _ in range(_MAX_RATES_PER_PACKET):
             neuron_id, rate = neuron_id_rates[pos]
             key = id_to_key_map[neuron_id]
-            message.add_key_and_payload(key, int(round(rate * scale)))
+            message.add_key_and_payload(
+                key, int(round(Decimal(str(rate)) * scale)))
```
andrewgait: Is this necessary because `scale` is a `Decimal`, and therefore `rate` needs to be as well?  Seems a somewhat convoluted way of doing things, but if it's necessary then it's fine...

spynnaker/pyNN/models/spike_source/spike_source_poisson_vertex.py
```diff
@@ -191,7 +194,7 @@ def get_resources_used_by_atoms(
             [self._spike_recorder.get_sdram_usage_in_bytes(
                 vertex_slice.n_atoms, self._max_spikes_per_ts(
                     vertex_slice, n_machine_time_steps, machine_time_step),
-                self._N_POPULATION_RECORDING_REGIONS) * n_machine_time_steps],
+                n_machine_time_steps)],
```
andrewgait: If _N_POPULATION_RECORDING_REGIONS is no longer being used then it should be deleted at the top of the class...

andrewgait:
Also I presume that this depends on https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/369 ?

spynnaker/pyNN/connections/spynnaker_poisson_control_connection.py
rowleya: This is an unfortunate consequence of using Decimal.  This is also how it is done in DataSpecification.

spynnaker/pyNN/models/spike_source/spike_source_poisson_vertex.py
rowleya: Indeed it should - fixed!

PR #569, opened 2018-11-12
Extractor (Changes to BufferManager calls)

Christian-B:
Replaces https://github.com/SpiNNakerManchester/sPyNNaker/pull/565

Must be done at same time as: https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/370

### DIFF
--- a/spynnaker/pyNN/models/common/neuron_recorder.py
+++ b/spynnaker/pyNN/models/common/neuron_recorder.py
@@ -180,9 +180,6 @@ class NeuronRecorder(object):
                     for index in self._indexes[SPIKES])
                 if neurons_recording == 0:
                     continue
-                if neurons_recording < vertex_slice.n_atoms:
-                    # For spikes the overflow position is also returned
-                    neurons_recording += 1
             # Read the spikes
             n_words = int(math.ceil(neurons_recording / 32.0))
             n_bytes = n_words * self.N_BYTES_PER_WORD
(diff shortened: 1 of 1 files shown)


spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -105,20 +105,21 @@ def get_matrix_data(
...
 
             row_length = self.N_BYTES_FOR_TIMESTAMP + \
                 n_neurons * self.N_BYTES_PER_VALUE
 
             # There is one column for time and one for each neuron recording
             n_rows = record_length // row_length
-            # Converts bytes to ints and make a matrix
-            record = (numpy.asarray(record_raw, dtype="uint8").
-                      view(dtype="<i4")).reshape((n_rows, (n_neurons + 1)))
+            if record_length > 0:
```
alan-stokes: not sure what this bits are for. suspisious of data recorder changes over the raw_data. so flagging for someone whos more understanding this block.

Christian-B: Ok Caught me. That is actually a bug fix that I sneaked in here.  It addresses the case where there is no spike data returned from the machine. Rowley's numpy magic fails if there are zero bytes in the data.

andrewgait: I'm pretty sure this is fine - not surprising that the case where no spike data is returned wasn't really dealt with, as that rarely happens.  As far as I understand it, happy for this to be merged.

PR #580, opened 2019-01-04
Sdram per timestep

Christian-B:


spynnaker/pyNN/models/spike_source/spike_source_poisson_vertex.py
```diff
@@ -174,7 +174,7 @@ def get_resources_used_by_atoms(
 
         # build resources as i currently know
         container = ResourceContainer(
-            sdram=SDRAMResource(self.get_sdram_usage_for_atoms(vertex_slice)),
+            sdram=ConstantSDRAM(self.get_sdram_usage_for_atoms(vertex_slice)),
```
rowleya: Shouldn't this return a variable size as it uses recording?

rowleya:
I have been looking through this.  One thing it does is force the software to use auto pause and resume.  I think this is OK, as the user can now disable auto pause and resume by setting the minimum time to run for to their script run time.  There is also an option to do live buffered recording, but as we know that this doesn't really work, perhaps this needs to be disabled here also.

Christian-B:
> Looks like there is a bit of work to be done here for the Poisson vertex, as otherwise it might not do auto pause and resume. It currently relies on the old way of doing things...

Poisson Vertex if the current system could ever trigger and auto pause resume YES must be fixed.

Will Sit with @rowleya to work out how.

Christian-B:
> I have been looking through this. One thing it does is force the software to use auto pause and resume.

It should not force more auto pause resume if any thing less.

In the old way the partition-er was give a minimum sdram to save for recording and would split up based on that.   This way gives a cost model and a minimum time to run to partition. If the partition-er is leaving too many neurons on each core then the default minimum is probably too low.    

In the current system the auto pause integration tests with artificially reduced memory where splitting it into machine vetrexs of 2 neurons each. Different to what it did 1 years ago and not really desirable.

Assuming that the minimum time is set to partition at the desirable level the auto pause may well run with larger steps (each run longer) as the new way allocates the sdram for all vertexes on the chip based on need and not just a total/number of vertexes.  The auto pause calculations try to run as long as possible (based on placements) and ignore the minimum timesteps used for partitioning.

Future work could make a smarter placer that tries to spread the most sdram greedy vertexes across several chips.

spynnaker/pyNN/models/spike_source/spike_source_poisson_vertex.py
Christian-B: I admit I do not understand the poisson full. But if its need for sdram changes with longer runs than 100% agreed.
Exactly why I named the change from SDRAMResource to ConstantSDRAM to force the decision.

That said

Christian-B: I have improved the poisson cost models significantly now

PR #582, opened 2019-01-15
Add robot control types

rowleya:
Adds a "type" to the data to be sent at fixed intervals for robot control.  This means that the voltage can be interpreted in different ways (the NRP needs the voltage to be a fixed-point number but the pushbot needs a signed integer).


dkfellows:
I've tried to merge this code up to date, but I'm not sure if it is correct: there seems to be both a scaling factor and a type control factor now, and I don't know how they interact…

PR #584, opened 2019-01-16
Routing table compression tied to bit field generator

alan-stokes:
still working on.


alan-stokes:
circumvented by c version

Issue #589, opened 2019-01-29
math domain error, SpikeSourcePoisson with FromListConnector


andrewgait:
While trying to sort out a problem for someone, I came across something that might be wrong; the question is whether the user is doing something daft or not.  The following script throws a ValueError because it tries to take the log of zero; it only does this because (a) it uses a FromListConnector with variable delays so that the delay variance is non-zero, and (b) the rate in the SpikeSourcePoisson population has been set to zero by the user.  Should a user be allowed to do this or should we tell them not to set a rate of zero?

(In the code in synaptic_manager.py, it basically makes a choice between using the "rate" value, or using "spikes_per_second" as set in the config file, so we can quite easily make it take the correct choice, or tell the user they're doing something strange...)

```
import spynnaker8 as sim
sim.setup(timestep=1.0)
poisson = sim.Population(2, sim.SpikeSourcePoisson(
    duration=1.0e10, rate=0.0, start=0.0))
receiver = sim.Population(2, sim.IF_curr_exp())

connections = [
(0, 0, 1.0, 7.0),
(1, 0, 2.0, 7.0),
(0, 1, 3.0, 7.0),
(1, 1, 4.0, 8.0)
]

conn = sim.FromListConnector(connections)
syn = sim.StaticSynapse(weight=0.5, delay=1)
sim.Projection(presynaptic_population=poisson,
               postsynaptic_population=receiver,
               connector=conn, synapse_type=syn, receptor_type="excitatory")

receiver.record(["v", "spikes"])
poisson.record("spikes")

sim.run(200.0)
```

rowleya:
The tools should cope with a rate of 0.  In particular, the issue here comes down to the interaction with the FromListConnector which I don't fully understand; if you replace it with an AllToAllConnector it works fine.  I am guessing then that it is the values used that causes the issue?

andrewgait:
The AllToAllConnector has a fixed delay value for every connection between the populations (from the StaticSynapse).  The FromListConnector allows you to set a different delay for each connection between the populations.  If the pre-vertex is a SpikeSourcePoisson then spikes_per_second gets set to the rate value; then if the variance across the delays is non-zero, this spikes_per_second value is used in the "_ring_buffer_expected_upper_bound" calculation, where, if it's zero, it causes the ValueError.  I think the solution to it is basically to use the rate value if it's zero, but to default back to the config value otherwise.

andrewgait:
[...]

PR #590, opened 2019-01-29
Deal with a zero rate SSPoisson case correctly

andrewgait:
Fix for issue https://github.com/SpiNNakerManchester/sPyNNaker/issues/589 - this should allow rate=0 in all cases.


spynnaker/pyNN/models/neuron/synaptic_manager.py
```diff
@@ -461,7 +461,17 @@ def _get_ring_buffer_to_input_left_shifts(
                     spikes_per_second = self._spikes_per_second
                     if isinstance(app_edge.pre_vertex,
                                   SpikeSourcePoissonVertex):
-                        spikes_per_second = app_edge.pre_vertex.rate
+                        rate = app_edge.pre_vertex.rate
+                        # If non-zero rate then use it; otherwise keep default
+                        if hasattr(rate, "__getitem__"):
+                            sumrate = 0
+                            for i in range(len(rate)):
```
rowleya: probably easier to do sumrate = sum(rate)

andrewgait: Definitely is - fixed.

spynnaker/pyNN/models/neuron/synaptic_manager.py
```diff
@@ -461,7 +461,10 @@ def _get_ring_buffer_to_input_left_shifts(
                     spikes_per_second = self._spikes_per_second
                     if isinstance(app_edge.pre_vertex,
                                   SpikeSourcePoissonVertex):
-                        spikes_per_second = app_edge.pre_vertex.rate
+                        rate = app_edge.pre_vertex.rate
+                        # If non-zero rate then use it; otherwise keep default
+                        if (sum(rate) != 0):
```
rowleya: Almost - you need some of what you had before, as you can only sum things that are lists (unless this is always guaranteed to be a list).

andrewgait: It's guaranteed to be a list (SpikeSourcePoissonVertex.rate is set using utility_calls.convert_param_to_numpy, which enforces that single values become single-valued arrays).

rowleya:
Updated to account for new information available.  I have tested it against the case in the issue and this seems to be fine.  I will merge when green...

PR #595, opened 2019-02-18
Remove synapse type header as not now needed

oliverrhodes:
Change implemented to enable building of a neuron model with non-standard neuron implementation and STDP.

This update works for me!


andrewgait:
I did also see a few extra warnings during the build process as well - I'll have another double-check through these tomorrow and see what they're pointing at.  For now though this works fine.

andrewgait:
Warnings found and removed, merging now.

PR #596, opened 2019-02-22
Exception is users tries something that currently broken

Christian-B:
Current selective record is broken if indexes (or view used to create indexes) has repeats or out of order neuron ids.

Not sure this is worth fixing so just going boom!


Christian-B:
Actually there is no reason not to allow this. Need to work out a warning if user sets indexes and the reguest data on the whole population.

alan-stokes:
this brnach needed?

Christian-B:
NOPE. Left just to annoy @alan-stokes

PR #598, opened 2019-02-25
Fix duplicate error and sort for consitency.

Christian-B:
Bugfix if user asks for duplicates in recording indexes.

Note: If the users sets recording on a view gets data on the whole population the data comes out in numerical order. If they also call get_data on the view (or a similar one) goes out in the view order and can include duplicates.


Christian-B:
When merged also delete:
https://github.com/SpiNNakerManchester/sPyNNaker/tree/jenkins2

PR #604, opened 2019-03-05
Fix duplicate error and sort for consitency.

Christian-B:
population view fix

Found with Jenkin2 testing but independant


Christian-B:
separate pr ignore

Issue #612, opened 2019-03-07
FromListConnector with wieghts and more than 255 cause exception


Christian-B:
If you use a fromList or fromFile connector with more than 255(I think) and weights and delays it throws

pynnaker.pyNN.exceptions.SynapseRowTooBigException: The connection between pop1 with 200 atoms and pop2 with 300 atoms has more synapses (300) than can currently be supported on this implementation of PyNN (255 for this connection type). Please reduce the size of the target population, or reduce the number of neurons per core.

Without weights and delays it works

Script:
import spynnaker8 as sim

sources = 200
destinations = 300
aslist = []
for s in range(sources):
    for d in range(destinations):
        aslist.append((s, d, 5, 2))

sim.setup(1.0)
sim.set_number_of_neurons_per_core(sim.IF_curr_exp, 100)
pop1 = sim.Population(sources, sim.IF_curr_exp(), label="pop1")
pop2 = sim.Population(destinations, sim.IF_curr_exp(), label="pop2")
synapse_type = sim.StaticSynapse(weight=5, delay=2)
projection = sim.Projection(
    pop1, pop2, sim.FromListConnector(aslist),
    synapse_type=synapse_type)
sim.run(0)
from_pro = projection.get(["weight", "delay"], "list")
print(len(from_pro))
sim.end()

Christian-B:
I think the error is a Safety check that is no longer valid.

i tried 3000 destinations and that worked.

Ok it lost packets as I had 3000 source neurons all firing at the same time.

But the connector worked even 300 destinations. 
(Did like a one to one)

Christian-B:
import spynnaker8 as sim

sources = 3000
destinations = 3000
aslist = []
for s in range(sources):
        aslist.append((s, s, 5, 2))

sim.setup(1.0)
sim.set_number_of_neurons_per_core(sim.IF_curr_exp, 10)
pop1 = sim.Population(sources, sim.SpikeSourceArray(spike_times=[0]), label="input")
pop2 = sim.Population(destinations, sim.IF_curr_exp(), label="pop2")
synapse_type = sim.StaticSynapse(weight=5, delay=2)
projection = sim.Projection(
    pop1, pop2, sim.FromListConnector(aslist),
    synapse_type=synapse_type)
pop2.record("spikes")
sim.run(20)
from_pro = projection.get(["weight", "delay"], "list")
print(len(from_pro))
spikes = pop2.spinnaker_get_data("spikes")
[...]

PR #614, opened 2019-03-08
A four character fix.  Oops.

andrewgait:
Fix for issue https://github.com/SpiNNakerManchester/sPyNNaker/issues/612.


Christian-B:
And interesting and I assumed desired side effect (over just removing the safety code) is that not the parishioner splits the neurons up more which results in longer runs being possible and I assume less spikes dropped.

Christian-B:
Tested against the 9000000 test from https://github.com/SpiNNakerManchester/sPyNNaker/issues/612

PR #621, opened 2019-03-20
No more layout enums

dkfellows:
This is a branch where I'm trying to make our code use this:
```c
typedef struct {
    uint32_t bar;
    uint32_t grill;
    uint16_t quux[];
} foo_t;
foo_t *foo = (foo_t *) address;
op_with(foo->bar, foo_grill, foo->quux);
```
instead of:
```c
enum foo_e {
    BAR_INDEX,
    GRILL_INDEX,
    QUUX_INDEX
};
op_with(address[BAR_INDEX], address[GRILL_INDEX], (uint16_t *) &address[QUUX_INDEX]);
// or even worse, this (which I've seen things like):
// op_with(address[BAR_INDEX], address[GRILL_INDEX], (uint16_t *)(((uint8_t *)address) + (QUUX_INDEX * sizeof(uint32_t) / sizeof(uint8_t)));
```
The differences get more profound once we use types other than these basic ones, with casts and weird address arithmetic.

See also SpiNNakerManchester/SpiNNFrontEndCommon#395 which is part of the same programme of changes.


rowleya:
Having merged in master, this seems to cause RTEs and other issues.  See:
http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/sPyNNaker8%20Integration%20Tests/detail/no-more-layout-enums/5/pipeline/

Issue #618, opened 2019-03-15
OneToOneConnector can not handle big difference in delays


Christian-B:
Script:
import spynnaker8 as sim

sim.setup(timestep=1.0)
input_pop2 = sim.Population(2, sim.SpikeSourceArray([0]),  label="input")
pop = sim.Population(2, sim.IF_curr_exp(), label="pop")

c2 = sim.Projection(input_pop2, pop, sim.OneToOneConnector(),
                    sim.StaticSynapse(weight=[0.3, 1.0],  delay=[1, 32]))
sim.run(1)
print(c2.get(['weight', 'delay'], 'list'))
sim.end()

error:
Exception: Too much synaptic memory has been written: 32 of 17 

-
Notes:
from list connector can handle the differing delays
all to all can to (adjusting size of inpUt_pop to 1)
Works ok if the delays are closer together.
So  delay=[1, 2] works as does delay=[31, 32]

Christian-B:
delay=[17, 120] works

delay[1, 17] does not

rowleya:
My initial guess is that this is down to the fact that OneToOneConnector can use direct synapses, but in this case it can't; so the memory estimation is probably made on the assumption that it can use direct synapses but the data is generated on the basis that it can't.

PR #624, opened 2019-03-25
Compare using both the max and the min, otherwise mixed cases are missed

andrewgait:
Fix for https://github.com/SpiNNakerManchester/sPyNNaker/issues/618 - the case where the delays were an array containing mixed values where some required delay extensions and some did not was not dealt with correctly.

### DIFF
--- a/spynnaker/pyNN/models/neural_projections/connectors/one_to_one_connector.py
+++ b/spynnaker/pyNN/models/neural_projections/connectors/one_to_one_connector.py
@@ -43,9 +43,9 @@ class OneToOneConnector(AbstractGenerateConnectorOnMachine):
 
         slice_min_delay = min(delays)
         slice_max_delay = max(delays)
-        if slice_max_delay >= min_delay and slice_max_delay <= max_delay:
+        if min_delay <= slice_max_delay <= max_delay:
             return 1
-        if slice_min_delay >= min_delay and slice_min_delay <= max_delay:
+        if min_delay <= slice_min_delay <= max_delay:
             return 1
         return 0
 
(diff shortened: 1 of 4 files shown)


spynnaker/pyNN/models/neural_projections/connectors/one_to_one_connector.py
```diff
@@ -43,7 +43,9 @@ def get_n_connections_from_pre_vertex_maximum(
 
         slice_min_delay = min(delays)
         slice_max_delay = max(delays)
-        if slice_min_delay >= min_delay and slice_max_delay <= max_delay:
+        if min_delay <= slice_max_delay <= max_delay:
+            return 1
+        if min_delay <= slice_min_delay <= max_delay:
```
dkfellows: Right now, this is effectively an “or” that returns 1 if _either_ delay is in range. Should it instead check if _both_ delays are in range? If so, it can become:
```python
if min_delay <= slice_min_delay <= slice_max_delay <= max_delay:
    return 1
```

rowleya: I don't think this is quite right so far - it is meant to be an overlap check.  See for an example: 
https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/spynnaker/pyNN/models/neural_projections/projection_machine_edge.py#L44

rowleya: Obviously the above is checking the overlap of something different, but the idea is similar.

andrewgait: It needs to be a separate check for each delay as the values that are passed in for "min_delay" and "max_delay" are different depending on the circumstances (in this case, whether a delay extension is required or not).  The way it's written works properly in that it gives the correct answer in each scenario, but I agree there might be a more intuitive way of writing it.  The problem is that it is an "overlap check", but it's performed over two separate calls to the function...

andrewgait: Rewritten to use or, hopefully making it a bit clearer.

andrewgait:
This PR also now fixes https://github.com/SpiNNakerManchester/sPyNNaker/issues/633 as well (the two issues are vaguely related to each other).

PR #627, opened 2019-03-29
C version of router compressor with bit field

alan-stokes:
this encapsulates the bit-field generation from previous branch, and calls the router table compressor with bit fields and then redoes synaptic expansion so that the router compressor can use that memory.


alan-stokes:
clsoing as dead branch

PR #631, opened 2019-04-05
Stdp with alpha and ca2

such-a-git:
Added stdp make files for IF_curr_alpha and IF_curr_exp_ca2_adaptive, fully tested and working


neural_modelling/src/neuron/structural_plasticity/synaptogenesis/topographic_map_impl.c
```diff
@@ -601,7 +601,7 @@ bool synaptogenesis_dynamics_formation_rule(void)
     uint no_elems = number_of_connections_in_row(
         synapse_row_fixed_region(rewiring_dma_buffer.row));
     if (no_elems >= rewiring_data.s_max) {
-        log_error("row is full");
+        log_debug("row is full");
         return false;
```
andrewgait: The rest of this is fine - I was just wondering whether this change is related to Peter's branch(es), or do you want it included here?

such-a-git: It is related to Peters branch so doesn't need be included, however, it was necessary to running properly. Peter seemed fine with it being changed too as it had little impact and did not need to error out anyway.

andrewgait: Ok, in that case we'll keep it in here.

Issue #636, opened 2019-04-11
stdp can also run out of SDRAM


Christian-B:
Script:  (stripped down vogels_2011  Not using vogels)
import spynnaker8 as sim

NUM_EXCITATORY = 2000

sim.setup(timestep=1.0, min_delay=1.0, max_delay=10.0)
sim.set_number_of_neurons_per_core(sim.IF_curr_exp, 10)

ex_pop = sim.Population(NUM_EXCITATORY, sim.IF_curr_exp)
in_pop = sim.Population(NUM_EXCITATORY / 4, sim.IF_curr_exp)

ex_pop.record(['spikes'])

sim.Projection(ex_pop, in_pop, sim.FixedProbabilityConnector(0.02),
               receptor_type='excitatory',
               synapse_type=sim.StaticSynapse(weight=0.03))
sim.Projection(ex_pop, ex_pop, sim.FixedProbabilityConnector(0.02),
               receptor_type='excitatory',
               synapse_type=sim.StaticSynapse(weight=0.03))

sim.Projection(in_pop, in_pop,  sim.FixedProbabilityConnector(0.02),
               receptor_type='inhibitory',
               synapse_type=sim.StaticSynapse(weight=-0.3))

ie_projection = sim.Projection(
    in_pop, ex_pop, sim.FixedProbabilityConnector(0.02),
    receptor_type='inhibitory', synapse_type=synapse_type=sim.StaticSynapse(weight=-0.3))

sim.run(1200000)
sim.end()

Error during:
            spinn_front_end_common/interface/interface_functions/chip_runtime_updater.py

2019-04-11 09:30:56 ERROR: 1, 0, 12: RUN_TIME_EXCEPTION (API) IF_curr_exp_std
2019-04-11 09:30:56 ERROR: r0=0x00000013 r1=0x00000000 r2=0x00004000 r3=0x00003304
2019-04-11 09:30:56 ERROR: r4=0x603937F0 r5=0x00400018 r6=0x00000001 r7=0x00000000
....
2019-04-11 09:30:59 ERROR: 1, 0, 12: Could not allocate recording region 0 of 8094032 bytes (recording.c: 589)
.....

Christian-B:
related to https://github.com/SpiNNakerManchester/sPyNNaker/issues/620

Christian-B:
Removed stdp from original script and it still fails.

Christian-B:
[...]

PR #637, opened 2019-04-12
reimplented get_max_atoms_per_core

Christian-B:
Fix for SpiNNakerManchester/sPyNNaker#636

Merge at same time as other PRs to fix that.


spynnaker/pyNN/models/spike_source/spike_source_array_vertex.py
```diff
@@ -166,10 +166,6 @@ def clear_spike_recording(self, buffer_manager, placements, graph_mapper):
     def set_model_max_atoms_per_core(new_value=sys.maxsize):
         SpikeSourceArrayVertex._model_based_max_atoms_per_core = new_value
```
alan-stokes: _model_based_max_atoms_per_core  does not exist inside the class to begin with. so broken if i were to put sockets on here

Christian-B: This is the same as master so out of scope for this PR.

Issue #601, opened 2019-02-27
“v” is not a valid parameter for IF_curr_exp.


Christian-B:
related to 
https://github.com/SpiNNakerManchester/sPyNNaker/issues/599

email with @apdavison 
in reply to:
sim.setup(1.0)
pop = sim.Population(1, sim.IF_curr_exp(v=-64, tau_syn_E=1), {}, label="pop")
He said:
"The script shouldn’t even get that far, though, as “v” is not a valid parameter for IF_curr_exp."

Alternatives are:
1. Allow the v but log a non standard warning
2. Move v to additional parameters
3. A combination of 1 and 2

I assume the same applies to other state variables.

apdavison:
For reference, this should look like:
`pop = sim.Population(1, sim.IF_curr_exp(tau_syn_E=1), initial_values={'v': -64}, label="pop")`

andrewgait:
Just to add a note: what @apdavison suggests here works already as far as I can tell; it should be possible to at least direct users to do things this way rather than the other less correct way.  As someone commented on a different issue, this could simply be a SpiNNaker "feature" if we wanted it to be...

Christian-B:
I agree as long as the "offical" way works there is no reason to break other ways.
A log once warning is suitable.

Christian-B:
fixed with https://github.com/SpiNNakerManchester/sPyNNaker/pull/671

PR #640, opened 2019-04-18
warning if initial_values set with cell params

Christian-B:
Compromise for https://github.com/SpiNNakerManchester/sPyNNaker/issues/601


spynnaker/pyNN/models/neuron/additional_inputs/additional_input_ca2_adaptive.py
```diff
@@ -27,7 +27,10 @@ def __init__(self,  tau_ca2, i_ca2, i_alpha):
             DataType.S1615,   # i_ca_2
             DataType.S1615])  # i_alpha
         self._tau_ca2 = tau_ca2
-        self._i_ca2 = i_ca2
+        if i_ca2 is None:
```
alan-stokes: cant this be default values in the __init__?

rowleya:
Unfortunately, this looks like it will break "default_initial_values".  I have instead created #671 which should do the same thing without breaking this.

Christian-B:
replaced with https://github.com/SpiNNakerManchester/sPyNNaker/pull/671

PR #632, opened 2019-04-06
Faster from list

rowleya:
Makes the FromListConnector faster by pre-splitting the connection list and then using the cached version.  This has now been tested by users and seems to make a big difference.

PR #643, opened 2019-05-03
from list machine edge filtering up to date with spynnaker master 240419

rjames91:
Automatic empty edge filtering if someone is using a from list connector. May be worth merging with master for the benefit of those experimenting with CNN style networks.

I think the conn_matrix method of doing this is working correctly and efficiently. I would suggest some further testing to be 100% sure though.


andrewgait:
Hi Rob, thanks for this.  I wonder whether it might be possible for you to test whether this works in conjunction with the changes in https://github.com/SpiNNakerManchester/sPyNNaker/pull/632/ ?  If not then I'll get round to looking at it at some point (and reviewing this at the same time), but I'm busy with something else right now if you have time instead...

rowleya:
I agree that the idea of filtering is good.  In the current master, there is a self.__split_conn_list which is a dict in which the key is (pre_vertex_slice.hi_atom, post_vertex_slice.hi_atom), which should then be usable for filtering.  This probably needs an extra access method in the from list connector so that if isinstance(connector, FromListConnector): n_conenctions = connector.n_connections(pre_vertex_slice, post_vertex_slice). This should then give you what you want directly in the connector without additional processing.

andrewgait:
There's a problem here, though - isn't the split_conn_list only populated when building the synaptic matrix (i.e. after filtering has happened)?  The method that Rob has added to in projection_machine_edge.filter_edge is (I think) called before connector.create_synaptic_block().  Or are you suggesting that the filtering should be done entirely within the FromListConnector?

rowleya:
Well, the information is cached, so you could populate the cache at filtering time if that was more useful.  The cache requires that the list of pre-slices and post-slices haven't changed, but they shouldn't between filtering and DSG.  Note that the computation of the split only happens once in total and then the cache is used forever more provided the slices stay the same.

andrewgait:
@rjames91 - this has been rewritten now to use the newer version of the FromListConnector.  I think it's still doing what you wanted it to, but if you could check it using any appropriate scripts you have then that would be appreciated.

rjames91:
ok sounds good to me, I don't have any specific test scripts for this. A simple one would be to run a script with two 1000 neuron populations connected with a from list projection where the list only contains a single entry (e.g. (0,0)). By changing the neuron indices in this list you'll be able to see if it filters the relevant core to core edges.

andrewgait:
Thanks.  I think this is working now from brief testing if someone else wants to review it.

spynnaker/pyNN/models/neural_projections/connectors/from_list_connector.py
```diff
@@ -235,6 +236,14 @@ def __repr__(self):
     def conn_list(self):
         return self.__conn_list
 
+    @property
+    def split_conn_list(self):
+        return self.__split_conn_list
+
+#     @property
+#     def conn_matrix(self):
+#         return self.__conn_matrix
```
rowleya: These should be removed

spynnaker/pyNN/models/neural_projections/connectors/from_list_connector.py
```diff
@@ -18,6 +18,7 @@ class FromListConnector(AbstractConnector):
     """
     __slots__ = [
         "__conn_list",
+        "__conn_matrix",
```
rowleya: This can be removed

spynnaker/pyNN/models/neural_projections/connectors/from_list_connector.py
```diff
@@ -235,6 +236,14 @@ def __repr__(self):
     def conn_list(self):
         return self.__conn_list
 
+    @property
+    def split_conn_list(self):
```
rowleya: It might make more sense to make this something like "get_n_connections(pre_slices, post_slices, pre_slice, post_slice)".  Then you can create the split_conn_list in here if it doesn't already exist.

spynnaker/pyNN/models/neural_projections/projection_machine_edge.py
```diff
@@ -43,6 +43,26 @@ def filter_edge(self, graph_mapper):
...
+                pre_app_vertex = graph_mapper.get_application_vertex(
+                    self.pre_vertex)
+                post_app_vertex = graph_mapper.get_application_vertex(
+                    self.post_vertex)
+                pre_slices = graph_mapper.get_slices(pre_app_vertex)
+                post_slices = graph_mapper.get_slices(post_app_vertex)
+                # run through connection list and check for any connections
+                # between the pre and post vertices that could be filtered
+                try:
+                    synapse_info.connector._split_connections(
```
rowleya: See comment in the from_list_connector - once get_n_connections exists, you can just call this method.

spynnaker/pyNN/models/neural_projections/projection_machine_edge.py
```diff
@@ -43,6 +43,26 @@ def filter_edge(self, graph_mapper):
...
+                        pre_slices, post_slices)
+                    split_list = synapse_info.connector.split_conn_list[
+                        (pre_hi, post_hi)]
+                    if len(split_list) == 0:
+                        n_filtered += 1
+                except ValueError:
```
rowleya: I don't think the try/except is needed as there is nothing that would be expected to go wrong here if you already know it is a from_list_connector.

andrewgait:
I think I've resolved everything in the above review.  Will run the integration tests again now.

rowleya:
Seems to pass everything, so I will merge it

PR #663, opened 2019-06-11
No need to check the same thing twice

andrewgait:
Looks like this was left in accidentally during a merge - thanks to @lucaperes for spotting it.


rowleya:
Looks like this has already been removed in another PR (the check has completely gone now).

PR #666, opened 2019-06-28
Variable-rate Poisson spike source

pabogdan:
Extracted from the Structural plasticity branch. A bit more testing is required.

Another pull request is also required in sPyNNaker8.

Example use:
```python
import spynnaker8 as sim
from pynn_object_serialisation.functions import intercept_simulator, restore_simulator_from_file
from spynnaker8.extra_models import SpikeSourcePoissonVariable
import numpy as np


runtime = 5000
sim.setup(timestep=1.0, min_delay=1.0, max_delay=144.0)
N_layer = 200  # number of neurons in each population

t_stim = 250
number_of_slots = int(runtime / t_stim)
range_of_slots = np.arange(number_of_slots)
slots_starts = np.ones((N_layer, number_of_slots)) * (range_of_slots * t_stim)
durations = np.ones((N_layer, number_of_slots)) * t_stim
rates = np.random.randint(10, 20, size=(N_layer, number_of_slots))

# Spike sources
poisson_spike_source = sim.Population(N_layer, SpikeSourcePoissonVariable(
    starts=slots_starts, rates=rates, durations=durations), label='poisson_source')
sim.run(runtime)
```

this script should create a poisson spike source with 200 neurons, each changing to a random firing rate after each 250 ms.


dkfellows:
While I think that, _scientifically speaking_, encoding information in just the spike rate is wrong (akin to saying the only information in human speech is in the volume!), since actual scientists use the spike rate as an input, it is useful to be able to change the rate by programming it as described. It also seems to be not a huge change from the user perspective. 

Reviewing the change itself on my phone on the train... Tricky. 😉

spynnaker/pyNN/models/spike_source/spike_source_poisson_variable.py
```diff
@@ -0,0 +1,34 @@
...
+_population_parameters = {"seed": None}
+
+DEFAULT_MAX_ATOMS_PER_CORE = 500
+
+
+class SpikeSourcePoissonVariable(AbstractPyNNModel):
+
+    default_population_parameters = _population_parameters
+
+    def __init__(self, rates=[1.0], starts=[0], durations=None):
```
dkfellows: Alas, this is Bad Practice in Python because it is using a read-write parameter as a default. Can we use a `tuple` instead? (i.e., `tuple(1.0)` and `tuple(0)`)

rowleya: I suspect these should be required parameters instead - surely you always want to supply a list of rates and start times...

pabogdan:
https://github.com/SpiNNakerManchester/sPyNNaker/pull/687 already contains this so we can safely close this PR.

alan-stokes:
yet again. deleting branch or storing for magic pixies?

rowleya:
Answered with a delete

Issue #600, opened 2019-02-27
pop.initialize( silently does nothing.


Christian-B:
Related to:
https://github.com/SpiNNakerManchester/sPyNNaker/issues/599

Email with @apdavison 
I would expect it to raise an Exception, as pop.initialize() should only be called at the start of a simulation, or after sim.reset().

andrewgait:
Making notes on this: currently pop.initialize(...) does "work" in between two runs as long as no pop.set(...) is also called (there must be some form of overwriting going on here that takes the current state off the machine when set() is called).  

This should most likely be changed so that the current state overrides any call to initialize(), and on top of that there should at least be a warning thrown (and perhaps an exception) if the user tries to initialize() anything in between two runs without having called reset() first.

andrewgait:
This was fixed by https://github.com/SpiNNakerManchester/sPyNNaker/pull/669, so closing.

Issue #599, opened 2019-02-26
pop.set(.. followed by pop.reset(... causes state variables not to be reloaded.


Christian-B:
Script:
import spynnaker8 as sim

sim.setup(1.0)
pop = sim.Population(1, sim.IF_curr_exp, {}, label="pop")
pop.set(i_offset=1.0)
pop.set(tau_syn_E=1)
pop.record(["v"])
sim.run(5)
pop.set(tau_syn_E=1)
#sim.run(5)
v1 = pop.spinnaker_get_data('v')
sim.reset()
sim.run(5)
#sim.run(5)
v2 = pop.spinnaker_get_data('v')
sim.end()
print(v1)
print(v2)

If you uncomment the second runs the states are reset correctly.

If you do 
pop.initialize(v=.60) nar the reset ONLY v is reset the other states gsyn_exc remain unchanged.
(note not checkable with this script as gsyn_exc is zero)

Christian-B:
Or more likely is is setting the state variables but to the values it obtained to do second pop.set(tau_syn_E=1)

andrewgait:
In a similar vein to the above script,

import spynnaker8 as sim

sim.setup(1.0)
pop = sim.Population(1, sim.IF_curr_exp, {}, label="pop")
pop.set(i_offset=1.0)
pop.record(["v"])

pop.set(tau_syn_E=1)
sim.run(5)
pop.set(tau_syn_E=1)
sim.run(5)
v1 = pop.spinnaker_get_data('v')

sim.reset()
pop.set(tau_syn_E=1)
sim.run(5)
pop.set(tau_syn_E=1)
sim.run(5)
v2 = pop.spinnaker_get_data('v')

sim.end()

print(v1)
print(v2)

reset() goes back to the state at the last call to set() (i.e. all the parameters AND state variables at that point).  I'm not sure this is the expected behaviour in this script... ?

rowleya:
I think we already knew that this was broken (but good to have a ticket for it).  There have been several attempts to fix it but nothing has properly addressed the issue yet...  This is linked to the fact that state variables and parameters are all tied in to a single data structure for each neuron, so treating them differently is difficult. The current execution cycle says that after a reset, load the last data specification.  Unfortunately this means that it will load the state variables too.  Another option would have been to go to the first data specification after a reset, but of course this means that set since the last run would then be broken instead!

[...]

PR #669, opened 2019-07-02
Better reset support

rowleya:
Changes associated with SpiNNakerManchester/SpiNNFrontEndCommon#450

### DIFF
--- a/neural_modelling/src/delay_extension/delay_extension.c
+++ b/neural_modelling/src/delay_extension/delay_extension.c
@@ -189,7 +189,7 @@ static bool initialize() {
     if (!simulation_initialise(
             data_specification_get_region(SYSTEM, address),
             APPLICATION_NAME_HASH, &timer_period, &simulation_ticks,
-            &infinite_run, SDP, DMA)) {
+            &infinite_run, &time, SDP, DMA)) {
         return false;
     }
     simulation_set_provenance_function(
--- a/neural_modelling/src/neuron/c_main.c
+++ b/neural_modelling/src/neuron/c_main.c
@@ -131,7 +131,7 @@ static bool initialise() {
     if (!simulation_initialise(
             data_specification_get_region(SYSTEM_REGION, address),
             APPLICATION_NAME_HASH, &timer_period, &simulation_ticks,
-            &infinite_run, SDP, DMA)) {
+            &infinite_run, &time, SDP, DMA)) {
         return false;
     }
     simulation_set_provenance_function(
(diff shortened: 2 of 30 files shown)


andrewgait:
Will review this soon, but just doing some simple tests so far suggests that this is behaving more correctly than previously, but there's still something not quite right - it seems as though I can do pop.set(...) after sim.reset(), but pop.initialize(...) after sim.reset() appears to have no effect.  

(Both pop.set() and pop.initialize() correctly do nothing at present if they are called after sim.run() without a sim.reset() - it was suggested at one point that this should raise an exception but I think it's better to silently ignore it or perhaps a warning message would be a good idea?)

rowleya:
There was an issue where I had put something in the wrong place - I believe this is now resolved.

I also got tricked when looking at this because of Neo segments.  Remember that after calling reset, the next set of data is in segments[1]...

andrewgait:
That has fixed it.  (I was using spinnaker_get_data(..) in my script to get around the issues with neo segments...)

andrewgait:
There's still something not quite right here.  In the new test I just pushed to the sPyNNaker8 branch at p8_integration_tests/quick_test/test_multi_call_examples/test_reset_initialize.py, if I only do a pop.initialize() between the third and fourth runs (i.e. I comment out the pop.set on line 33), then it (incorrectly) initializes the voltage to -61.  With the pop.set() included the initialize() is ignored.

rowleya:
It should be possible to initialise without calling reset - this is just like set for state variables (I believe).  If reset is called, it should return to the values set at time == 0, so run, run, initialise, run, reset means that the values go back to before the first run I believe.  If you do run, reset, initialise, run, reset, the values should go back to just after the reset, because the time was == 0 at this point.  Set should not have an effect however, so I will look at why this is happening.

andrewgait:
Going off the exchange at these issues: https://github.com/SpiNNakerManchester/sPyNNaker/issues/600 and https://github.com/SpiNNakerManchester/sPyNNaker/issues/599 I'd suggest that initialize() shouldn't do anything if it's not called in conjunction with reset(), or at least that's what Andrew Davison seems to be suggesting...

rowleya:
Oops, I had forgotten about this.  OK, so initialise when time is not 0 should raise an exception.  I will update this now.

rowleya:
OK, this has been updated and I now think it is doing the right thing!

alan-stokes:
there seems to be some left over changes to do wtih connectors slices in here which are not related to the change mapping. 
the actual changes to change mapping im happy with.

rowleya:
Ah, that *is* specific to this branch.  In this case, the connection slice changes are to ensure that if you reset and re-generate the connections, you should get the same connectivity as you did the first time.

alan-stokes:
bugger must have missed that. ok fair enough

PR #671, opened 2019-07-05
Initial values 2

rowleya:
A replacement for #640 which uses the existing infrastructure to check for the assignment of initial values through the constructor.  Note that this version keeps the values of the initial values in tact, so the default_initial_values still works.


rowleya:
Note - please close #640 if this gets merged instead.

rowleya:
Oops, missed a place where the indirection was missed.  I have added a test to ensure this doesn't go wrong again.

PR #672, opened 2019-07-08
Fix to the indexing of synapse info

rowleya:
This still wasn't quite right - I had an example where it was failing, but this fixes it.  Can you check again with your own example and see that I haven't broken it again?

### DIFF
--- a/spynnaker/pyNN/models/neuron/synaptic_manager.py
+++ b/spynnaker/pyNN/models/neuron/synaptic_manager.py
@@ -7,6 +7,7 @@ import struct
 import numpy
 import scipy.stats  # @UnresolvedImport
 from scipy import special  # @UnresolvedImport
+from six import iteritems
 from spinn_utilities.helpful_functions import get_valid_components
 from data_specification.enums import DataType
 from spinn_front_end_common.utilities.helpful_functions import (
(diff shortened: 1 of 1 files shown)


rowleya:
It was something Basab's student was working on.  It did involve STDP as well.  I think the issue was that the population was receiving connections from various sources, but never the same one twice (so index should have always been 0).  The connector was also always specified rather than being reused.  I don't know if this was the issue since the connector was being used to identify the index previously.  I will try to make a small example version to see if I can replicate it.

rowleya:
Here is a simple version that shows the issue:
```
import spynnaker8 as p

p.setup(1.0)

source_1 = p.Population(10, p.IF_curr_exp(), label="source_1")
source_2 = p.Population(10, p.IF_curr_exp(), label="source_2")

target = p.Population(10, p.IF_curr_exp(), label="target")

proj_1 = p.Projection(
    source_1, target, p.FixedProbabilityConnector(0.3),
    p.StaticSynapse(weight=1.0))

proj_2 = p.Projection(
    source_2, target, p.FixedProbabilityConnector(0.2),
    p.StaticSynapse(weight=2.0))

p.run(10)

print(proj_1.get("weight", "list"))
print(proj_2.get("weight", "list"))

p.end()
```

In this case, the first projection works, but the second prints an empty list.

PR #674, opened 2019-07-09
Fix save

rowleya:
This finally should fix the indexing issue, which was broken again when an application vertex was split by partitioning!  This also goes with a sPyNNaker8 branch which is fixing another issue but also running the integration tests for this fix...


rowleya:
Still not quite there - I'll keep trying!

andrewgait:
This was how I fixed this issue last time I saw it: https://github.com/SpiNNakerManchester/sPyNNaker/pull/668/files - not sure how that fits with this fix.

rowleya:
Not sure - I removed that bit as it was causing other issues, but I have something in place now that just asks the master pop table for the index, since that is where it is supposed to be from...

rowleya:
OK, it now seems to work everywhere finally!

Issue #613, opened 2019-03-08
projection.get before run fails for some conntors


Christian-B:
In some cases getting weights before running and only accessing them after works in others it does not!

Example script: (pick one of the porjections only of course

import spynnaker8 as sim

sources = 2
destinations = 3

sim.setup(1.0)
pop1 = sim.Population(
    sources, sim.SpikeSourceArray(spike_times=[0]), label="input")
pop2 = sim.Population(destinations, sim.IF_curr_exp(), label="pop2")
synapse_type = sim.StaticSynapse(weight=5, delay=2)
projection = sim.Projection(
    pop1, pop2, sim.OneToOneConnector(),
    synapse_type=synapse_type)
projection = sim.Projection(
    pop1, pop2, sim.FromListConnector([[0,0]]),
    synapse_type=synapse_type)
from_pro = projection.get(["weight", "delay"], "list")
sim.run(1)
print(len(from_pro))
sim.end()

Christian-B:
As we now support.
sim.run(0)

I wonder if the real fix is require user to add a sim.run(0) if they want weights at time 0

And then nuke the whole ConnectionHolder work around!

andrewgait:
This fails for connectors which use the on-chip expander (the ConnectionHolder is not populated with connections in these instances until after run(...) has completed).  So it may be the case that the best fix is to require the user to do sim.run(0) if they want the initial weight values, particularly as we're planning to move towards having the majority of the connectors use on-chip expansion.

Christian-B:
sim.run(0) is NOT an option until at least the SDRAM branches come in.

If you turn on recording the code has to decide 
1. How much sdram to save for recording
2. How many auto pause resume steps to use.

This throws a divide by zero exception.

But even if we fixed that the first run would save enough space for 0 time steps and then the next run would try to do t in X runs of 0 steps and still go BOOM!

rowleya:
It should always be possible to call get on the connections before run, though using the result is invalid until after run.  I suspect this use was forgotten when the synapse expander was brought in; in this case, it should fill in the connection holder as part of the synapse expander algorithm.  This is easy enough to do but a bit of effort is required to do the implementation as the synaptic manager needs to be updated to be able to return a list of connection holders that need to be filled in post expansion.  Once this is available, it should just be a matter of running through them post expansion and getting the synapses as usual.

Issue #661, opened 2019-06-05
Import loop


dkfellows:
Observed while looking at [the logs of a documentation build](https://travis-ci.org/SpiNNakerManchester/sPyNNaker/jobs/541158339):

`DelayExtensionVertex` has an import loop with `spynnaker/pyNN/models/neuron/__init__.py`, `spynnaker/pyNN/models/neuron/abstract_population_vertex.py`, and `spynnaker/pyNN/models/neuron/synaptic_manager.py`…

```
/home/travis/build/SpiNNakerManchester/sPyNNaker/doc/source/spynnaker.pyNN.rst:23: WARNING: autodoc: failed to import module u'spynnaker.pyNN.abstract_spinnaker_common'; the following exception was raised:
Traceback (most recent call last):
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/sphinx/ext/autodoc.py", line 551, in import_object
    __import__(self.modname)
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/sPyNNaker-1!4.0.1-py2.7.egg/spynnaker/pyNN/abstract_spinnaker_common.py", line 13, in <module>
    from spynnaker.pyNN.models.utility_models import synapse_expander
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/sPyNNaker-1!4.0.1-py2.7.egg/spynnaker/pyNN/models/utility_models/__init__.py", line 3, in <module>
    from .delay_extension_vertex import DelayExtensionVertex
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/sPyNNaker-1!4.0.1-py2.7.egg/spynnaker/pyNN/models/utility_models/delay_extension_vertex.py", line 31, in <module>
    from spynnaker.pyNN.models.neuron.synapse_dynamics import (
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/sPyNNaker-1!4.0.1-py2.7.egg/spynnaker/pyNN/models/neuron/__init__.py", line 1, in <module>
    from .abstract_population_vertex import AbstractPopulationVertex
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/sPyNNaker-1!4.0.1-py2.7.egg/spynnaker/pyNN/models/neuron/abstract_population_vertex.py", line 24, in <module>
    from .synaptic_manager import SynapticManager
[...]

PR #681, opened 2019-07-17
Fix projection get before run

andrewgait:
Fixes https://github.com/SpiNNakerManchester/sPyNNaker/issues/613 - fills in connection data for projections/connectors that use the synapse expander.


andrewgait:
Latest commit fixes the doc-build errors observed at https://github.com/SpiNNakerManchester/sPyNNaker/issues/661

PR #690, opened 2019-08-02
cleans up matrix read

alan-stokes:
as requested as a seperate pr


alan-stokes:
dead pr. but the same issue with stand in the neuron recording. and its due to the mismatch between the format stored and the format to be read out. some need scales and these scales are not at the output, its from thew input. so they need to matched properly. 

killing pr anyhows. as suberceded by the neuron reocrding ones

PR #691, opened 2019-08-02
Match simulation API

andrewgait:
Don't know how this was missed or what uncovered it, but the call here didn't match the API correctly so it was failing to compile.


rowleya:
Ah yes, this would have come with the other branch but I'll merge this to fix master for now...

PR #695, opened 2019-08-12
Manage unigue labels in abstractSpynnaker base only

Christian-B:
Please wait for SpiNNakerManchester/sPyNNaker8#250 to verify that there are no monsters under the bed.

Recommended to do at same time as:
SpiNNakerManchester/SpiNNFrontEndCommon#470
SpiNNakerManchester/SpiNNakerGraphFrontEnd#119

### DIFF
--- a/unittests/mocks.py
+++ b/unittests/mocks.py
@@ -91,7 +91,7 @@ class MockSimulator(object):
     def add_population(self, pop):
         pass
 
-    def add_application_vertex(self, vertex):
+    def add_application_vertex(self, vertex, prefix):
         pass
 
     def verify_not_running(self):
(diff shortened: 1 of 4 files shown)


unittests/mocks.py
```diff
@@ -91,7 +91,7 @@ def get_random_distribution(self):
     def add_population(self, pop):
         pass
 
-    def add_application_vertex(self, vertex):
+    def add_application_vertex(self, vertex, prefix):
```
dkfellows: Needs the default value for `prefix`

dkfellows: Real default is `None` so that's easy

PR #698, opened 2019-09-02
convertor now finds logs with incorrect number of params so these lin…

Christian-B:
needed for https://github.com/SpiNNakerManchester/SpiNNUtils/pull/99

if you rather fix than comment the debugs even better


Christian-B:
NOT NEEDED

PR #702, opened 2019-09-05
Oops if multiple data for time received all where ignore leaving a hole

Christian-B:
This is minor as it touches code that should never be reached.

As @Alan is currently reworking that method he may want to checrry pick the fix into his branch and close this PR unmerged.


alan-stokes:
cherry picked

PR #704, opened 2019-09-19
Add spynnaker report

dkfellows:


rowleya:
@pabogdan has done something similar here:
https://github.com/pabogdan/pynn_object_serialisation/blob/master/pynn_object_serialisation/functions.py

The final version of this might want to include some of each e.g. run like this one, but writing json files with content like the other.

dkfellows:
While producing a JSON description of the PyNN-level graph might _also_ be a good idea, this report has its own value.

Christian-B:
Why is this a method in AbstractSpiNNakerCommon that is always run like it or not.
Instead of a report alogrithm that can be turned on or off with a cfg setting

Christian-B (changes requested):
I very much like the idea of a simple population report.

But it should be an algorithm and not a method

Christian-B:
> @pabogdan has done something similar here:
> https://github.com/pabogdan/pynn_object_serialisation/blob/master/pynn_object_serialisation/functions.py
> 
> The final version of this might want to include some of each e.g. run like this one, but writing json files with content like the other.

Peters too while good is a competently different level.

PR #711, opened 2019-10-02
Cherry pick changes for time steps

alan-stokes:
cherry picked changes from robert's mega branch that supports multiple vertices with different time steps

depends upon
spy
spy8
fec
pac


Christian-B:
Too big an API change to be considered worth the benefit.

Keeping branch for historical reasons

PR #721, opened 2019-10-26
Revert "Projections on popviews"

rowleya:
Reverts SpiNNakerManchester/sPyNNaker#718.

Some deduction has shown that this us causing some issues in master, so reverting until a fix is found...

Will create a PR to bring this back in a minute...


rowleya:
Need to get names to match!

Issue #732, opened 2019-11-13
Wrong variable in live poisson control code


dkfellows:
https://github.com/SpiNNakerManchester/sPyNNaker/blob/6d4b17dea981455fc49b19312c549bd87608e69e/spynnaker/pyNN/connections/spynnaker_poisson_control_connection.py#L113

Should that be `control_label` instead of `label`, which is calculated from `label` a few lines above?

Christian-B:
Quick fix with https://github.com/SpiNNakerManchester/sPyNNaker/pull/733

Christian-B:
I think however the correct fix is either of.
1. Check label first time it is used and go pop if wrong.
2. Check and fix the label every time it is used.
Similar to configparser use of the optionxform format.

PR #733, opened 2019-11-13
stop changing one label and using another.

Christian-B:
Quick fix for https://github.com/SpiNNakerManchester/sPyNNaker/issues/732

### DIFF
--- a/spynnaker/pyNN/connections/spynnaker_poisson_control_connection.py
+++ b/spynnaker/pyNN/connections/spynnaker_poisson_control_connection.py
@@ -104,9 +104,8 @@ class SpynnakerPoissonControlConnection(LiveEventConnection):
         :param label: The label of the Population to set the rates of
         :param neuron_id_rates: A list of tuples of (neuron ID, rate) to be set
         """
-        control_label = label
-        if not control_label.endswith(self.__control_label_extension):
-            control_label = self._control_label(label)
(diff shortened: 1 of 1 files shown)


rowleya:
The issue is that you want to send to a hidden population that the user doesn't know about (the connectivity goes Host -> Hidden Multicast Sender -> Poisson Source).  So the user only knows the label of the Poisson source, not the label of the population that is to actually receive the message and send the keys.  So there is some indirection here; the trick is to get it right!

What is more interesting is that this seems to work in master with the current code, according to the test: https://github.com/SpiNNakerManchester/sPyNNaker8/blob/master/p8_integration_tests/quick_test/test_spike_source/test_poisson_spike_source.py#L177

So what actually happens in master that makes it work?

andrewgait:
What happens in master that makes it work (at least for the test) is that the labels on the population, the connection, and the callback are all the same.  Try changing say the callback to something awkward like "input_control" and you'll see problems come up...

rowleya:
Yes, true that if you use _control in the postfix of the population label, it will likely go very wrong!  Maybe a check of this and a throw-up is useful, or else a label map rather than a simple check.

The issue I now see though is that the callback for the start function passes in the label.  Of course this passes the right label with the _control postfix, so it doesn't go wrong at that point...

PR #734, opened 2019-11-15
appears to work so using try except

Christian-B:
Remove tests that skip and other useless code.

Similar to other PRs with the same name but I think independent.

https://github.com/SpiNNakerManchester/PACMAN/pull/270


dkfellows:
The csa package is _so awful_…

PR #741, opened 2019-11-25
One conversion site

dkfellows:
It's complicated enough without spreading it all over the place. Let one class, `DataType`, be the sole expert on conversions.

See also SpiNNakerManchester/DataSpecification#77


andrewgait:
I've actually just remembered that there's an even worse case of this elsewhere that doesn't even refer to DataType.scale - it just specifies the number.   https://github.com/SpiNNakerManchester/sPyNNaker/blob/53132ef7d856521205d13fc4fe99004927013748/spynnaker/pyNN/models/spike_source/spike_source_poisson_vertex.py#L748 (the first one is an unsigned fract U032 and the second an accum S1615).

PR #746, opened 2019-11-28
Neuron recorder agdr

rowleya:
My changes to the neuron recorder.  This allows any types to be used in the recording; the C code just ensures that there is enough space for the recording to take place and provides some useful access functions for a few types.  This also currently allows "bitfield" recording of an arbitrary number of bitfield variables; these are treated separately in the C code for efficiency.

These changes make the neuron_do_timestep_update function mostly faster than it currently is by about 3 clock cycles per neuron, however it is slower by about 4 clock cycles per neuron when that neuron spikes.  This is probably caused by the fact that there are now more than one bitfield variable.  There are a couple of possible thoughts:
 1. Should there be more than one bitfield given that spikes are the only current user?  Putting this back to having just one might speed things up.
 1. Should neuron implementations take care of their own recording as Christian already did on the neuron_recorder_2 branch?  This would remove a number of loops, further speeding up the code.


alan-stokes:
the issue i have with both questions are as follows:

"Should there be more than one bitfield given that spikes are the only current user? Putting this back to having just one might speed things up."

if you lock it into 1, you are removing flexibility, what if i want a "this neuron received a packet" for debug at a later date? this would be currently supported by a second bitfield recorded var. I dont think we should remove flex just coz we dont currently see other uses. you know once the ability is given, users will find ways to use it. 

Given we can also add a new neuron.c that OUR models use which know we have 1 bitfield is acceptable. Its how @Christian-B got the extra speed up, and that feels like a better approach, given we probably will end up with a code generator eventually, if we're going to go for speed above all else.

"Should neuron implementations take care of their own recording as Christian already did on the neuron_recorder_2 branch? This would remove a number of loops, further speeding up the code."

there is 2 differing views here, replace neuron.c so that it records without the for loop. or augment the code so that the neuron IMPL does the recording, as then you remove the need for extra dtcm for buffers that are only used to wrap this interface. Heck, that probably would speed it up even more as it no longer needs to move data from 3 places. 

I'll review your impl of the neuron recorder once ive fixed the io protocol

rowleya:
So, so far this sounds like:
 1. Leave it so that bitfield recording allows multiple bitfields.
 1. Move recording in to the implementation.

I can have a go at these and see how they work and if they improve the speed of operation.

alan-stokes:
the issue with 
Move recording in to the implementation.

is that you now expose recording to the end user, which we discussed previously. So unless you've moved the goal posts on the view of speed..........

rowleya:
Yes, we just had a similar discussion; the end user is really exposed to some of the recording anyway, so I think a bit more exposure will be OK.  Lets see if it helps first though!

alan-stokes:
(cwl) i have no words

rowleya:
I have added the changes to make this faster by moving the records into the implementations.  This change is almost nothing to the implementer - instead of assigning a value to an array they have to call a function.

The tests are all passing now; critical to this is a change to JavaSpiNNaker: SpiNNakerManchester/JavaSpiNNaker#245.

alan-stokes:
i think the SPIKES_FUDGE_FACTOR = 1 that is in the Abstract_population_vertex.py has been made redundant by the changes of @rowleya . I cannot find 1 reference to it in any of the code at the moment

alan-stokes:
The SPIKES = "spikes" inside the Abstract_population_vertex.py is obsoleted by the NueornRecorder.Spikes class variable.

alan-stokes:
test_neuron_recorder.py has issues with the init params to the neuron recorder. aka the bitfield recoridngs are not v, gsyn. why the code doesnt test is a worry

alan-stokes:
get spikes should check that the variable sint in the matrix store

alan-stokes:
line 647 has magic number 2. should be named constant *VARIABLE_LENGTHS_ITEMS*

alan-stokes:
neuron_recorder.py line 746. that should go over the set(self.__sampling_rates) - set(self.__bitfield_variables) which avoids having to do the check about if its in the bit_fields in the first place.

note it might have been better to store the 2 groups seperately, so youd need neither of these tricks and jsut go ove rthe correct array.

alan-stokes:
ASB line 58 go bye bye, as its also in line 62

alan-stokes:
neuron_recording.c line 99 magic mask. named constant please.

alan-stokes:
timing_dependence_recurrent.py line 16 (import math) unused import. go bye bye

rowleya:
> neuron_recorder.py line 746. that should go over the set(self.__sampling_rates) - set(self.__bitfield_variables) which avoids having to do the check about if its in the bit_fields in the first place.
> 
> note it might have been better to store the 2 groups seperately, so youd need neither of these tricks and jsut go ove rthe correct array.

The subtraction doesn't work (this isn't something you can do with an OrderedDict); holding the values just for this seems overkill.

alan-stokes:
neuron_recording.h docs on the structs. recording_info_t, bitfield_info_t, recording_info, bitfield_info, recording_values, bitfield_values,

alan-stokes:
neuron_recording.c docs for recording_info, bitfield_info, recording_values, bitfield_values, bitfield_recording_data, neuron_recording_header, neuron_recording_initialise, neuron_recording_reset, 

fill in params in for:
bitfield_data_size

alan-stokes:
neuron_recorder.c building the bitfield_info.values size and n words uses a n_neurons_rec +1. whats the +1 for? why isnt it just n_neurons_rec? what boundry case is this trying to protect against?

alan-stokes:
this might be going over kill but meh

in neuron_recording.c 
        uint32_t alloc_size = recording_info[i].size +
                recording_info[i].element_size;

have a comment above stating the addon of the overflow component. just to be clear what its doing

alan-stokes:
neuron_recorder.py line 746. that should go over the set(self.__sampling_rates) - set(self.__bitfield_variables) which avoids having to do the check about if its in the bit_fields in the first place.

note it might have been better to store the 2 groups seperately, so youd need neither of these tricks and jsut go ove rthe correct array.

The subtraction doesn't work (this isn't something you can do with an OrderedDict); holding the values just for this seems overkill. 

fine. was just trying avoid a if with a continue. but if too much faff. then too much faff

alan-stokes:
neuron_recording.h line 205 "// recording_do_timestep_update(time);" go bye bye

alan-stokes:
just to check. if this is wrong. please tell me. but isnt 

neuron_recording.h neuron_recording_record_value
just a fault tolerance version of the accum, double, float, int32 versions? aka it handles all of those and possibly more? and that the others are only there for performance views?

if this is the case, it might be worth adding something into the docs of these to flag this up to users. coz at the moment the neuron_recording_record_value sticks out like a sore thumb. 

if im wrong. please reset my views

alan-stokes:
once all these little niggles are covered. im happy to rubber stamp this into master.

rowleya:
Hopefully the changes just pushed fix these things!

alan-stokes:
just a wee gotcha in docs of neuron_recording.h 

the "neuron_recording_record_value" only works on matrix based ones. it would blow up in bitfields. but the "the functions below" include bit_field one. so moving around may be required. or more docs to stop stupid. 

Also ya missed the same comment on the int32

rowleya:
Done.  Just running the Integration tests again to make sure nothing is now broken...

PR #775, opened 2020-02-04
Allocate the right size!

rowleya:


neural_modelling/src/synapse_expander/param_generator.c
```diff
@@ -132,7 +132,7 @@ param_generator_t param_generator_init(uint32_t hash, address_t *in_region) {
         if (hash == type->hash) {
             // Prepare a space for the data
             param_generator_t generator =
-                    spin1_malloc(sizeof(param_generator_t));
+                    spin1_malloc(sizeof(struct param_generator));
```
andrewgait: Shouldn't this be 

`struct param_generator *generator = spin1_malloc(sizeof(struct param_generator));`

rowleya: It's the same thing, but yes this makes it consistent with the other allocations

PR #777, opened 2020-02-05
Generate the inverse CDF lut as a numpy array

dkfellows:
This lets us write it in the DSG more compactly


dkfellows:
Some experimentation (with `timeit`) leads me to conclude that this ought to be quite a bit faster… for the few times it is needed at all, which will depend on the number of STDP synapses with recurrent timing dependence.

PR #779, opened 2020-02-10
Filter edges correctly when using PopulationViews

andrewgait:
Fixes so that OneToOneConnector works correctly when filtering ProjectionMachineEdges while using PopulationViews.

Tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/347 (with new test).


spynnaker/pyNN/models/neural_projections/projection_machine_edge.py
```diff
@@ -56,8 +56,38 @@ def filter_edge(self, graph_mapper):
...
-                    n_filtered += 1
+                # Filter edge if both are views and outside limits
+                if ((synapse_info.prepop_is_view) and
+                        (synapse_info.postpop_is_view)):
+                    prepop_lo = synapse_info.pre_population._indexes[0]
+                    prepop_hi = synapse_info.pre_population._indexes[-1]
+                    postpop_lo = synapse_info.post_population._indexes[0]
+                    postpop_hi = synapse_info.post_population._indexes[-1]
+                    if ((prepop_hi < pre_lo or prepop_lo > pre_hi) or
+                            (postpop_hi < post_lo or postpop_lo > post_hi)):
```
rowleya: Not sure this is right - this seems to check that the population view is in range of the population.  Shouldn't it check pre_hi/lo against post_hi/lo?

spynnaker/pyNN/models/neural_projections/projection_machine_edge.py
```diff
@@ -56,8 +56,38 @@ def filter_edge(self, graph_mapper):
...
+                        n_filtered += 1
+                # Filter edge if pre-pop is outside limit and post_lo is bigger
+                # than n_pre_neurons
+                elif synapse_info.prepop_is_view:
+                    prepop_lo = synapse_info.pre_population._indexes[0]
+                    prepop_hi = synapse_info.pre_population._indexes[-1]
+                    prepop_size = synapse_info.n_pre_neurons
+                    if ((prepop_hi < pre_lo or prepop_lo > pre_hi) or
+                            (prepop_size < post_lo)):
```
rowleya: As above pre vs. post

spynnaker/pyNN/models/neural_projections/projection_machine_edge.py
```diff
@@ -56,8 +56,38 @@ def filter_edge(self, graph_mapper):
...
+                        n_filtered += 1
+                # Filter edge if post-pop is outside limit and pre_lo is bigger
+                # than n_post_neurons
+                elif synapse_info.postpop_is_view:
+                    postpop_lo = synapse_info.post_population._indexes[0]
+                    postpop_hi = synapse_info.post_population._indexes[-1]
+                    postpop_size = synapse_info.n_post_neurons
+                    if ((postpop_hi < post_lo or postpop_lo > post_hi) or
+                            (postpop_size < pre_lo)):
```
rowleya: As above pre vs. post.

spynnaker/pyNN/models/neural_projections/projection_machine_edge.py
andrewgait: This is checking whether the values from the indices of a PopulationView are outside the current slice which the View could be a part of.  So it needs to check prepop_view_hi/lo vs. pre_slice_lo/hi (and likewise for post).

spynnaker/pyNN/models/neural_projections/projection_machine_edge.py
andrewgait: This is a bit more confusing; the first part is the same as the above case, but it's also necessary to check the size of the pre-population against the low value of the post slice (the post-pop in this case is not a view, so the OneToOne will start from index 0 and continue for as long as there are values from the pre-pop View available to it).

spynnaker/pyNN/models/neural_projections/projection_machine_edge.py
andrewgait: Same answer as above, but swap pre- for post-

spynnaker/pyNN/models/neural_projections/projection_machine_edge.py
rowleya: OK, yes, I think I see this now.  Although not wrong, would this filter out everything?  For example, if I have two populations of 6 neurons with 2 neurons per core, and connect a[0:4] to b[2:6], as follows:
```
a: |01|23|--|
b: |--|01|23|
```
There will now be an edge between machine vertices a[2:3] and b[2:3].  Both of these have something in the pre and post slices that are in range, but there will be no synapses on this edge I think.

A possible thought I had was to subtract ```prepop_lo``` from ```pre_lo``` and ```pre_hi``` and correspondingly ```postpop_lo``` from ```post_lo``` and ```post_hi```.  You can then compare the results of these to check for the cross over.  I think you still need the above check as well though.

spynnaker/pyNN/models/neural_projections/projection_machine_edge.py
rowleya: I think this suffers from the same issue (again not wrong) e.g.:
```
a: |--|01|23|
b: |01|23|
```
Again a[2:3] -> b[2:3] should be filtered but I don't think it is.

spynnaker/pyNN/models/neural_projections/projection_machine_edge.py
rowleya: Same again with a and b swapped above.

spynnaker/pyNN/models/neural_projections/projection_machine_edge.py
andrewgait: I think I've got this working correctly now.

andrewgait:
I probably still need to add a test to check the filterer is working properly; will add it when I've thought of the best way to do it.

rowleya:
Yes, it looks OK, but a test would be good!

andrewgait:
It's not so easy to do this test as a unittest - (a) filter_edge is currently called after run(), and (b) OneToOneConnectors with PopulationViews are by default now done on machine.  So I'm inclined just to do it as an integration test...

PR #780, opened 2020-02-11
Recalculate ring buffer shifts if they're all zero

andrewgait:
Issue found by trying to add a projection in between two runs, between two already-existing populations.  The issue is that the ring_buffer_shifts were not recalculated after the new projection was added.

New test added in https://github.com/SpiNNakerManchester/sPyNNaker8/pull/350


rowleya:
It seems that this will only work when the ring buffer shifts are 0 at the start; if you already have projections and add another, you probably want to recalculate this again there too.

rowleya:
This looks more like it now, and the tests all pass, so I will merge it.

PR #786, opened 2020-02-19
Kernel uint32 fix

andrewgait:
Make KernelConnector work correctly for neuron index > UINT16_MAX


dkfellows:
Divmod by repeated subtraction? That's only really acceptable when the expected likely range of the dividend is between 0 and around 3 times the divisor.

dkfellows:
My concerns are addressed in #788

PR #787, opened 2020-02-21
overrides is now ONLY in spinutils!

Christian-B:
Must be done at or before merging https://github.com/SpiNNakerManchester/PACMAN/pull/281


alan-stokes:
are we keeping this branch for any reason?

Christian-B:
no

PR #790, opened 2020-02-24
10 randon tries is enough after 2 years

Christian-B:
@rowleya we have been testing the same 1000 random number every time for 2 years now.

Lets just test ten each time from now on!

### DIFF
--- a/unittests/model_tests/test_connectors.py
+++ b/unittests/model_tests/test_connectors.py
@@ -17,6 +17,7 @@ from __future__ import print_function
 import functools
 import numpy
 import pytest
+import random
 from pacman.model.graphs.common import Slice
 from spynnaker.pyNN.models.neural_projections.connectors import (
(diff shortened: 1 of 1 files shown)


rowleya:
The point of the 1000 is to protect against changes... I guess if you think 10 will be enough, I am happy to change it.

A better change would be to fix the code to guarantee that rows are never longer than estimated i.e. cut off rows that are too long.  Note that the synaptic expander does this, so it wouldn't be a bad thing for the python code to match up anyway...

Christian-B:
Fully agree with the better fix but out of scope here.

dkfellows:
Two years is a very long timeout.

rowleya:
OK, I thought that this was going to sometimes go wrong due to the randomness, but I now see that it is picking out 10 of the original 1000 random seeds, so it should still work consistently...

Issue #766, opened 2020-01-14
reset doesnt work fully alas


alan-stokes:
Having had a explore of this stuff for a competely different side issue. I noticed that our reset (aka set time = 0) in the c code will not have wiped the ring buffer and input buffer. so if tehre are delayed spikes to be processed by a core. bad things will happen. aka precognition

so we need to augment our resume callback to know when going to 0 to clear said buffers.

andrewgait:
Looking at this now; confirmed that this is an issue, example below shows it

```
import spynnaker8 as sim
from pyNN.utility.plotting import Figure, Panel
import matplotlib.pyplot as plt

sim.setup(timestep=1.0, min_delay=1.0, max_delay=144.0)

runtime = 50

inp = sim.Population(1, sim.SpikeSourceArray(spike_times=[5,20]), label='input')
out = sim.Population(1, sim.IF_curr_exp(), label='test')

sim.Projection(inp, out, sim.OneToOneConnector(),
               sim.StaticSynapse(weight=5.0, delay=7))

inp.record(["spikes"])
out.record(["v", "spikes"])

sim.run(runtime/2)

sim.reset()

# out.set(v_thresh=-58.0)

sim.run(runtime/2)

spikes_in = inp.get_data("spikes")
v1 = out.get_data("v")
spikes1 = out.get_data("spikes")

Figure(
    # membrane potential of the postsynaptic neuron
    Panel(v1.segments[0].filter(name='v')[0],
          xlabel="Time (ms)", xticks=True,
          ylabel="Membrane potential (mV)",
          data_labels=[out.label], yticks=True, xlim=(0, runtime)),
    # raster plot of the postsynaptic neuron spike times
    Panel(spikes1.segments[0].spiketrains,
          xlabel="Time (ms)", xticks=True,
          yticks=True, markersize=1.5, xlim=(0, runtime)),
    # raster plot of the presynaptic neuron spike times
    Panel(spikes_in.segments[0].spiketrains,
          xlabel="Time (ms)", xticks=True,
          yticks=True, markersize=1.5, xlim=(0, runtime)),
    # plots of the second segment
    Panel(v1.segments[1].filter(name='v')[0],
          xlabel="Time (ms)", xticks=True,
          ylabel="Membrane potential (mV)",
          data_labels=[out.label], yticks=True, xlim=(0, runtime)),
    # raster plot of the postsynaptic neuron spike times
    Panel(spikes1.segments[1].spiketrains,
[...]

PR #793, opened 2020-02-27
Deal with reset case with left over delayed spike

andrewgait:
Fixes https://github.com/SpiNNakerManchester/sPyNNaker/issues/766

This is probably overkill, but the alternative is to ask the machine to report back that there are (delayed) spikes left over that have not yet been processed, and I'm not sure that's any better.

### DIFF
--- a/neural_modelling/src/neuron/synapses.c
+++ b/neural_modelling/src/neuron/synapses.c
@@ -383,7 +383,7 @@ uint32_t synapses_get_pre_synaptic_events(void) {
             synapse_dynamics_get_plastic_pre_synaptic_events());
 }
 
-void synapses_flush_ring_buffers() {
+void synapses_flush_ring_buffers(void) {
     uint32_t n_neurons_power_2 = n_neurons;
     uint32_t log_n_neurons = 1;
     if (n_neurons != 1) {
--- a/neural_modelling/src/neuron/synapses.c
+++ b/neural_modelling/src/neuron/synapses.c
@@ -41,6 +41,9 @@ static uint32_t n_synapse_types;
 // Ring buffers to handle delays between synapses and neurons
 static weight_t *ring_buffers;
 
+// Ring buffer size
+static uint32_t ring_buffer_size;
+
 // Amount to left shift the ring buffer by to make it an input
 static uint32_t *ring_buffer_to_input_left_shifts;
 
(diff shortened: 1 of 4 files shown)


rowleya:
I wondered if it was possible on the core to detect that time is equal to 0 just before the run (or -1 I guess), and then in that circumstance to clear all the spike buffers?

andrewgait:
Turns out that it was relatively straightforward to do this; all that was needed was to flush the ring_buffers at time=0.

neural_modelling/src/neuron/synapses.c
```diff
@@ -382,3 +382,28 @@ uint32_t synapses_get_pre_synaptic_events(void) {
     return (num_fixed_pre_synaptic_events +
             synapse_dynamics_get_plastic_pre_synaptic_events());
 }
+
+void synapses_flush_ring_buffers() {
```
dkfellows: That should be `void synapses_flush_ring_buffers(void) {` because of legacy C weirdness. (You'd be OK in C++.)

andrewgait: Good catch, somehow missed it here but not in the header file...

neural_modelling/src/neuron/synapses.c
```diff
@@ -382,3 +382,28 @@ uint32_t synapses_get_pre_synaptic_events(void) {
...
+
+    uint32_t n_synapse_types_power_2 = n_synapse_types;
+    if (!is_power_of_2(n_synapse_types)) {
+        n_synapse_types_power_2 = next_power_of_2(n_synapse_types);
+    }
+    uint32_t log_n_synapse_types = ilog_2(n_synapse_types_power_2);
+
+    uint32_t n_ring_buffer_bits =
+            log_n_neurons + log_n_synapse_types + SYNAPSE_DELAY_BITS;
+    uint32_t ring_buffer_size = 1 << (n_ring_buffer_bits);
```
rowleya: This does a lot of work to get the ring buffer size!  There are two easy options to avoid this:
 1. Store ```ring_buffer_size``` where it is calculated previously in a static global variable.
 1. See lines 304-326 above for how to loop over the neuron and synapse indices and use that to get the ring buffer index.

I suspect the first is the easier...

andrewgait: Done - stored as static global.

andrewgait:
> Is this enough by itself, or should we also be flushing the spike input buffer as well? I guess that might have flushed itself by this point...

Testing suggests that this is enough by itself and that the input buffer has been flushed by this point.

PR #796, opened 2020-03-03
Json graphs

Christian-B:
Moved EdgeToNKeysMapper earlier which allows the json to output earlier.

Not critical but as we may write the n keys mapping out to json do sooner

Not enforced by tokens as if it happens later and everything works nothing altered.

Part of https://github.com/SpiNNakerManchester/PACMAN/pull/287 but can be merged before or after.


spynnaker/pyNN/spynnaker.cfg
```diff
@@ -30,7 +30,7 @@ one_to_one_connection_dtcm_max_bytes = 2048
 # MallocBasedRoutingInfoAllocator, GraphEdgeFilter, EdgeToNKeysMapper
 
 application_to_machine_graph_algorithms = PartitionAndPlacePartitioner
-machine_graph_to_machine_algorithms = GraphEdgeFilter,OneToOnePlacer,NerRoute,BasicTagAllocator,EdgeToNKeysMapper,ProcessPartitionConstraints,MallocBasedRoutingInfoAllocator,BasicRoutingTableGenerator
+machine_graph_to_machine_algorithms = GraphEdgeFilter,EdgeToNKeysMapper,OneToOnePlacer,NerRoute,BasicTagAllocator,EdgeToNKeysMapper,ProcessPartitionConstraints,MallocBasedRoutingInfoAllocator,BasicRoutingTableGenerator
```
rowleya: This *will* run EdgeToNKeysMapper twice...

Christian-B: yes that is a mistake. Will fix.

rowleya:
This reordering breaks the live packet gather functionality.  This is because some algorithms are missing tokens to enforce order.  The specific issue in this case is that the generation of the NKeysMap happens before the edges of the live packet gatherers are inserted into the graph, so those edges don't get any keys.  Possible solutions include:
1) Forcing NKeysMap generation to depend on tokens generated by things that modify the graph.  This would be easy enough with tokens.
1) Change the bit that inserts the edges into the graph to also update the NKeysMap.  This would mean that the NKeysMap wouldn't include these edges in the report however...

alan-stokes:
i vote for 1. as that was also required for ...... something in i think bitfield compressors. so it would fix multiple issues in 1 go

Christian-B:
The order in the algorithm list should be a what would be convenient order as in this case it would be convenient if we run EdgeToNKeysMapper before OneToOnePlacer so that we can get the json files out before the placer. So we could run the placer independently.

If however algorithms depend on an order to work  this should be enforced by required inputs or tokens.

Christian-B:
files changed = 0 so closed.

Issue #798, opened 2020-03-05
Multiple KernelConnector projections between the same two populations give SynapticConfigurationException


ej159:
I'm trying to do convolution with the KernelConnector between two populations. The convolution has a number of channels of depth (typically >10). When running with FromList connectors this does not typically happen. I've tried this with master and pop_based_master_pop_table branches and encounter the same thing.

``` python 
import numpy as np
import spynnaker8 as p

x_res = 100.
y_res = 100.

shape_pre = np.asarray([x_res, y_res])
shape_kernel = np.asarray([5, 5])
pre_sample_steps = (1,1)
start_location = np.asarray((shape_kernel/2).astype(int))
shape_post = np.asarray([int((x_res - shape_kernel[0]) / pre_sample_steps[0]) + 1, int((y_res - shape_kernel[1]) / pre_sample_steps[1]) + 1])


kernels = [np.clip(np.random.rand(*shape_kernel), a_min=0, a_max=None) for i in range(32)]


print("shape_pre", shape_pre)
print("shape_post", shape_post)
print("shape_kernel", shape_kernel)
print("pre_sample_steps", pre_sample_steps)
print("start_location", start_location)

p.setup(timestep=1)
p.set_number_of_neurons_per_core(p.IF_curr_exp, 16)
p.set_number_of_neurons_per_core(p.SpikeSourceArray, 16)


spike_times = [[i] for i in range(int(x_res*y_res))]

pre_pop = p.Population(x_res*y_res, p.SpikeSourceArray(spike_times=spike_times))

post_pop = p.Population(shape_post[0]*shape_post[1], p.IF_curr_exp())

def make_projection(kernel):
    return p.Projection(pre_pop, post_pop, p.KernelConnector(shape_pre=shape_pre,
                                                           shape_post=shape_post,
                                                           shape_kernel=shape_kernel,
                                                           weight_kernel=kernel,
                                                           post_sample_steps_in_pre=pre_sample_steps,
                                                           post_start_coords_in_pre=start_location),
                                                           receptor_type='excitatory')


projections = [make_projection(kernel) for kernel in kernels]

runtime = x_res*y_res
p.run(runtime)

```

rowleya:
What is the error that you encounter?

ej159:
```spynnaker.pyNN.exceptions.SynapticConfigurationException: Address 0x89c9710 is out of range for this population table!```

andrewgait:
[...]

PR #799, opened 2020-03-05
Do the correct calculation for partitioning

andrewgait:
Fixes https://github.com/SpiNNakerManchester/sPyNNaker/issues/798

Tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/377


rowleya:
Looks OK; does it help with the issue?

andrewgait:
It doesn't fail at the partitioning stage any more - I admit that I did leave that script running after I left the office yesterday and got a different error, but I don't think it's necessarily related to this.  Perhaps we should wait for @ej159 to get back to us, as it's his script after all...

ej159:
This fixes that problem for up to about 32 kernels in the test script. Above this I get a SynapticConfigurationException `SynapticConfigurationException: Address 0x802c810 is out of range for this population table!`. Presumably that's just too many kernel to expand?

rowleya:
So that could mean that the size is still vastly overestimating...

andrewgait:
Will look at this again - could be that the estimate of the max number of connections to each post-neuron is also overestimated...

andrewgait:
Hi @ej159, I've been experimenting a bit more with your script, and the real issue actually seems to be the choice to run at a low number of neurons per core.  I'm not sure in terms of what you're trying to do whether this is necessary or not - I've found that increasing the number of neurons per core (for both populations) up to say 64 or even 128 seems to help a lot in terms of running this script (as an example, it currently takes ~20 minutes from start to finish to run 96 kernels at 128 neurons per core, which is much faster than anything I've seen when running at 16 neurons per core, even for a very low number of kernels).  I'm guessing this is because the data is more spread out across the machine and therefore simply just takes longer to be (sequentially) copied everywhere, but that's something of a guess.  Note here too that using Java helps with this script as the network is spread across 3 boards and therefore the data loading time vs. Python is reduced.

(I believe I have now improved the partitioning calculations on this branch so that you won't see pop-table address errors until you get to somewhere over 100 kernels; if you need more than that then there may be more we can do).

andrewgait:
I think the (necessary) fixes on this branch are ready to go in now.  

In the actual cases @ej159 and @such-a-git are trying to run, the problems are in general due to the size of the source populations; the suggested workaround in this case is to split up single source or target populations in to multiple separate populations as appropriate.

To give some more information, the memory size for a synaptic matrix is memory_size = n_atoms x row_size x n_kernels, even if row_size is small, will cause problems when n_atoms is big. In one case I have been looking at, the pre-pop is the whole "image" simulated as an SSA, so 304 x 240 = 72960 atoms. The post-pop is the same size as the number of filters/kernels applied - so something like row_size = (n_kernels + 3) * 4 (the 3 here is the fixed header size of a synaptic row). So 

memory_size = 72960 * n_kernels * [(n_kernels + 3) * 4].  

Limiting this to the SDRAM available using 128MB gives: 

n_kernels^2 + 3*n_kernels = ~438 

and solving this gives ~20 kernels in this case.  As you can see, the ways to reduce this are to reduce the number of kernels and/or the sizes of the populations.

What this doesn't explain entirely is why the "same" FromListConnectors are apparently working - more investigation of this (by @ej159 and @such-a-git) would be welcome...

alan-stokes:
could it be that the calcs for max row length for the from list is more accurate than that of the kernel conenctor?

andrewgait:
> could it be that the calcs for max row length for the from list is more accurate than that of the kernel conenctor?

We've discussed this previously and we don't think so - if the FromList is creating the same connections as the KernelConnector then the calculations should be the same.

alan-stokes:
same connections yes. but if the max row length is off. different sizes of padding's. which in turn makes different max size. Basically, as a possible scope to find a bug. is the padding the same?

andrewgait:
> same connections yes. but if the max row length is off. different sizes of padding's. which in turn makes different max size. Basically, as a possible scope to find a bug. is the padding the same?

I'm sure that the max row length should be the same in each case.  Whether it is or not is dependent on whether the function to create the FromList is actually reproducing what the Kernel is actually doing (I suspect not, hence the discrepancy).

andrewgait:
This has been tested by @such-a-git with the "iCub simulator" and seems to be working, so I think it's ready to be merged.

PR #803, opened 2020-03-10
SDRAM bug fix

Christian-B:
Fixes a few small issues in the sdram models.

Does not include the MultiRegionSDRAM that will come in a later PR.

### DIFF
--- a/spynnaker/pyNN/models/neuron/abstract_population_vertex.py
+++ b/spynnaker/pyNN/models/neuron/abstract_population_vertex.py
@@ -265,14 +265,8 @@ class AbstractPopulationVertex(
     def _get_sdram_usage_for_atoms(
             self, vertex_slice, graph, machine_time_step):
 
-        old_overhead = SYSTEM_BYTES_REQUIREMENT
-        #DATA_SPECABLE_BASIC_SETUP_INFO_N_BYTES + SIMULATION_N_BYTES
-        #APP_PTR_TABLE_BYTE_SIZE + SARK_PER_MALLOC_SDRAM_USAGE + SIMULATION_N_BYTES
-        #APP_PTR_TABLE_HEADER_BYTE_SIZE + MAX_MEM_REGIONS * 4 + SARK_PER_MALLOC_SDRAM_USAGE + SIMULATION_N_BYTES
-        #8 + 16 * 4 + (2 * 4) + (3 * 4) = 92
-
-        new_overhead = APP_PTR_TABLE_HEADER_BYTE_SIZE + MAX_MEM_REGIONS * 4
-        #SYSTEM = 0
+        # These values can be cross checked with the data_spec_text_files
+        # SYSTEM = 0
         region_0 = SIMULATION_N_BYTES
         # NEURON_PARAMS = 1
         region_1 = self._get_sdram_usage_for_neuron_params(vertex_slice)
@@ -292,21 +286,13 @@ class AbstractPopulationVertex(
         region_8 = profile_utils.get_profile_region_size(
             self.__n_profile_samples)
 
-        region_sum = region_0 + region_1 + sm_regions + region_6 + region_7 + region_8
-        malloced = new_overhead + region_sum
-        sdram_cost = malloced + SARK_PER_MALLOC_SDRAM_USAGE
+        overhead = APP_PTR_TABLE_HEADER_BYTE_SIZE + MAX_MEM_REGIONS * 4
 
-        sdram_requirement = (
-                SYSTEM_BYTES_REQUIREMENT +
-                self._get_sdram_usage_for_neuron_params(vertex_slice) +
-                self._neuron_recorder.get_data_spec_sdram_usage(vertex_slice) +
-                PopulationMachineVertex.get_provenance_data_size(
-                PopulationMachineVertex.N_ADDITIONAL_PROVENANCE_DATA_ITEMS) +
-                self.__synapse_manager.get_sdram_usage_in_bytes(
-                vertex_slice, machine_time_step, graph, self) +
-                profile_utils.get_profile_region_size(self.__n_profile_samples))
-
-        return sdram_requirement
+        # Should match HostExecuteDataSpecification.__malloc_region_storage
+        malloced =  overhead + region_0 + region_1 + sm_regions + region_6 + \
+                    region_7 + region_8
+        # Don't forget the cost of doing a malloc
+        return (malloced + SARK_PER_MALLOC_SDRAM_USAGE)
 
     def _reserve_memory_regions(self, spec, vertex_slice, vertex):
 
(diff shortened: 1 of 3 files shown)


spynnaker/pyNN/models/neuron/synaptic_manager.py
```diff
@@ -261,7 +261,8 @@ def _get_size_of_generator_information(self, in_edges):
         """ Get the size of the synaptic expander parameters
         """
         gen_on_machine = False
-        size = 0
+        # Overhead added by _write_synaptic_matrix_and_master_population_table
+        size = 1 * BYTES_PER_WORD
```
rowleya: I think this actually comes from the direct matrix region (region 10).

spynnaker/pyNN/models/neuron/synaptic_manager.py
```diff
@@ -322,14 +323,20 @@ def get_sdram_usage_in_bytes(
...
+        # SYNAPSE_DYNAMICS = 5
+        region_5 = self._get_synapse_dynamics_parameter_size(
+            vertex_slice, application_graph, app_vertex)
+        # SYNAPTIC_MATRIX = 4
+        region_4 = self._get_synaptic_blocks_size(
+            vertex_slice, in_edges, machine_time_step)
+        # POPULATION_TABLE = 3
+        region_3 = self.__poptable_type.get_master_population_table_size(in_edges)
+        #  DIRECT_MATRIX = 10
+        region_10 = self._get_size_of_generator_information(in_edges)
```
rowleya: The generator information goes in region 9.

spynnaker/pyNN/models/neuron/synaptic_manager.py
```diff
@@ -322,14 +323,20 @@ def get_sdram_usage_in_bytes(
...
+        return region_2 + region_5 + region_4 + region_3 + region_10
```
rowleya: It looks like the SDRAM calculation misses the size of the direct matrix; this looks like it could be a tricky calculation, and should technically be subtracted from the synaptic blocks size (or else this calculation of this should remove the direct synapses).  I think that the direct and normal synaptic matrices were once a single region, so the calculation was valid, but when they were split up, this error crept in.

To be complete, the regions required here are:
 - [x] SYNAPSE_PARAMS (2)
 - [x] POPULATION_TABLE (3)
 - [x] SYNAPTIC_MATRIX (4)
 - [x] SYNAPSE_DYNAMICS (5)
 - [x] CONNECTION_BUILDER (9)
 - [ ] DIRECT_MATRIX (10)

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
```diff
@@ -262,17 +264,36 @@ def _get_sdram_usage_for_neuron_params(self, vertex_slice):
...
+        # SYNAPTIC_MATRIX = 4
+        # SYNAPSE_DYNAMICS = 5
+        # DIRECT_MATRIX = 10
+        sm_regions = self.__synapse_manager.get_sdram_usage_in_bytes(
+                vertex_slice, machine_time_step, graph, self)
+        # PROFILING = 8
+        region_8 = profile_utils.get_profile_region_size(
+            self.__n_profile_samples)
+
+        overhead = APP_PTR_TABLE_HEADER_BYTE_SIZE + MAX_MEM_REGIONS * 4
```
rowleya: Just saw this too; there is a constant for this in FEC called DATA_SPECABLE_BASIC_SETUP_INFO_N_BYTES.  This also adds a sark allocation, so you might not need this below if this constant is used.

Christian-B: I intentional did not use DATA_SPECABLE_BASIC_SETUP_INFO_N_BYTES because it had the spark allocation included.  Removing it allowed me to come up with a number that could be compared with HostExecuteDataSpecification.__malloc_region_storage methods size param.

rowleya: Ah OK - in that case the equivalent constant is APP_PTR_TABLE_BYTE_SIZE (which is from data_specification.constants)

spynnaker/pyNN/models/neuron/synaptic_manager.py
Christian-B: Absolutely the 1 * BYTES_PER_WORD is applied to the Direct matrix region.

It is the method _write_synaptic_matrix_and_master_population_table that add is there.

My assumption was that the method _get_size_of_generator_information was used to calc the size of the Direct matrix region

spynnaker/pyNN/models/neuron/synaptic_manager.py
Christian-B: thanks.

spynnaker/pyNN/models/neuron/synaptic_manager.py
Christian-B: It was exactly my intention in this rewrite to align the cost methods with the regions so that we could detect the differences.

If you say that it is easier or more accurate to estimate the size of regions SYNAPTIC_MATRIX and DIRECT_MATRIX in a single method that is fine.  It only really makes sense to do a subtraction if someone want to check the estimate less the direct matrix estimate against the actual.

spynnaker/pyNN/models/neuron/synaptic_manager.py
Christian-B: moved.

spynnaker/pyNN/models/neuron/synaptic_manager.py
Christian-B: Ok changed the name to region 9 and move the +1 word to what is now the region4_10 method

Christian-B:
In a soon to follow on PR I will make it so these cost function can return variable sdram.  

This may make it easier to produce more accurate estimates.

spynnaker/pyNN/models/neuron/synaptic_manager.py
rowleya: Unfortunately, the issue is that the space will be estimated only once (both the DIRECT_MATRIX and SYNAPTIC_MATRIX together) and then the memory will be reserved for the SYNAPTIC_MATRIX using this value, and then even more memory will be allocated for the DIRECT_MATRIX.  This is again hidden because it will then estimate use for the CONNECTION_BUILDER which it won't then use.

As an example, the following should show the error:
```
import spynnaker8 as p

p.setup(1.0)
pre = p.Population(1, p.SpikeSourcePoisson(rate=1))
post = p.Population(1, p.IF_curr_exp())
post.record("all")


def no_gen(self, weights, delays):
    return False


p.OneToOneConnector.generate_on_machine = no_gen

proj = p.Projection(pre, post, p.OneToOneConnector(), p.StaticSynapse())

p.run(100)
```

Note that this overrides the generate_on_machine of the OneToOneConnector to stop the extra memory being allocated there from hiding the issue here.

So what is needed is a way to get a size for the SYNAPTIC_MATRIX and DIRECT_MATRIX separately so that this can be used in both estimation and region creation.

Christian-B:
Now part of https://github.com/SpiNNakerManchester/sPyNNaker/pull/805

Issue #575, opened 2018-12-11
count_rewires in c_main.c


alan-stokes:
this is used in log_debugs, but looks like a provenance data item. It should be pushed into the provenance data stuff and thus make it more reportable in the main tools.

pabogdan:
Very well spotted! Recording is something that I should have done a long time ago ...

And just to clarify do you mean recording (something that the user can request) or is this just something that would be added in the `reports` folder in some file somewhere?

alan-stokes:
"And just to clarify do you mean recording" 
i didnt say recording, you did. 

provenance is anything that can be used to either report errors/warnings to the end user, or anything that can report how things behaved outside of the recorded variables of spikes, v, gsyn. Thus allowing a second run to deduce if it was functionally equivalent to the first. 

all provenance data items are put into a core level report inside reports/provenance_items. But anything that can be assessed to produce a error/ warning, aka dropped packets. are interpreted by the host during the provenance translation phase. Look at abstractPopulationMachineVertex for deeper understanding. 

basically stuff that's just being reported, doesnt have a report and message param in provenanceDataItem.

PR #808, opened 2020-03-18
Report the final value of count_rewires back as provenance

andrewgait:
Fixes https://github.com/SpiNNakerManchester/sPyNNaker/issues/575 (I think) - is this a sensible value to be reporting back?


andrewgait:
I think I've done what was suggested in the latest commit, but the numbers coming back for the structural plasticity scripts in PyNN8Examples seem lower than what I'd expect given your explanation... ?  Could probably do with someone checking them...

pabogdan:
where can I find this provenance item?

rowleya:
It should go in with the rest of the provenance items for a core, like timer tick overrunning etc.

pabogdan:
Did something change with the default reporting? I'm not getting xmls out anymore in the docker I'm running this from.

andrewgait:
The place I've been looking at the numbers is in the provenance database file (provenance.sqlite3 in the reports/../run_1/provenance_data directory).

pabogdan:
Right. What is the recommended tool to look inside the database?

rowleya:
The SQL is default; if you want to change back to XML, you would have to add ```provenance_format = xml``` to the ```[reports]``` section of the cfg file.

andrewgait:
> Right. What is the recommended tool to look inside the database?

I've been using DB Browser for Sqlite (https://sqlitebrowser.org/) - other people may have other preferences, though.

andrewgait:
> An integration test for this could be to start off with no connections and only enable rewiring with a formation rule. The network should run until all synapses are most likely created. This way you'd know if the reported number is correct.

That's a good idea - I'll get round to writing one at some point soon, in that case.

andrewgait:
(I'll put it on an equivalent branch to this in sPyNNaker8 so that we can then merge them at the same time, and we get a further integration test check of this as a bonus).

pabogdan:
> That's a good idea - I'll get round to writing one at some point soon, in that case.
Conversely, one could start with all-to-all connectivity, a removal only rule, and run until there are no connections left to remove. The number of rewires should be the same in both cases(forming up, removing down).

neural_modelling/src/neuron/c_main.c
```diff
@@ -95,8 +96,8 @@ int32_t rewiring_period = 0;
 //! Flag representing whether rewiring is enabled
 bool rewiring = false;
 
-// FOR DEBUGGING!
-uint32_t count_rewires = 0;
+//! DEBGGING: for counting the number of rewiring attempts
```
alan-stokes, line 99: if this is going into prov data. its no longer for debugging. its now for prov.

alan-stokes:
with the comment changed. im fine with the changes

neural_modelling/src/neuron/c_main.c
andrewgait: Technically speaking this value is counting the number of attempts, whereas what goes into the provenance is the number of successful rewires.  But I have updated the comment anyway as there's no real need to say it's for debugging either.

PR #813, opened 2020-03-25
Remove utterly broken test

dkfellows:
The thing it was testing no longer had an even vaguely matching API (‘1000’ is not an IP address and never has been), so updating the test isn't worthwhile.


dkfellows:
Linked to https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/574 but formally independent

PR #814, opened 2020-03-26
change to remove negative in front of eta. Moved to -= cumulative del…

oliverrhodes:
This should get the code compiling on all machines - but please test for correct operation


andrewgait:
This change was incorporated in the pause-and-resume changes, so this PR is not needed any more.

alan-stokes:
branch needed?

PR #815, opened 2020-04-06
Fix pause-and-resume for eprop neuron model

andrewgait:
Fix pause-and-resume for eprop neuron; basically, the issue was that the global parameters were being regularised again on the second run after pausing.  It's possibly a bit confusing, but the current state of pause-and-resume (binary gets paused, and then starts again from the beginning of c_main()) means that initialise() gets called again at the start of the second run, so anything that should only happen on the first run needs to be guarded against the second time round.

### DIFF
--- a/neural_modelling/src/neuron/implementations/neuron_impl_eprop_adaptive.h
+++ b/neural_modelling/src/neuron/implementations/neuron_impl_eprop_adaptive.h
@@ -72,6 +72,9 @@ global_neuron_params_pointer_t global_parameters;
 // The synapse shaping parameters
 static synapse_param_t *neuron_synapse_shaping_params;
 
+// Bool to regularise on the first run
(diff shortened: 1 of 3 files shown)


andrewgait:
Note as well that I included the other PR change for eta in here too (https://github.com/SpiNNakerManchester/sPyNNaker/pull/814) because it was causing issues with my compiler...

andrewgait:
I also should have said that I have added code to update_values(...) in the python neuron_model_eprop, but this does not affect the pause-and-resume directly, as this function is only called when doing e.g. sim.run(..), pop.set(...), sim.run(...) or e.g. sim.run(...), sim.reset(...), sim.run(...).  What I've put there should work but needs testing with an example and if it needs fixing then probably a different branch and PR.

PR #821, opened 2020-05-15
Fix the connection estimates for FixedNumberPostConnector

andrewgait:
The connection estimates were incorrect for the FixedNumberPostConnector; this fixes it so that scripts that should work now do, rather than failing because of an overestimate of the synaptic row length required.

Additional test added as a check for this in the future: https://github.com/SpiNNakerManchester/sPyNNaker8/pull/406

### DIFF
--- a/spynnaker/pyNN/protocols/munich_io_spinnaker_link_protocol.py
+++ b/spynnaker/pyNN/protocols/munich_io_spinnaker_link_protocol.py
@@ -43,16 +43,16 @@ PUSH_BOT_MOTOR_WITHOUT_UART_MASK = 0x7C0
 PUSH_BOT_MOTOR_UART_SHIFT = 0 + _OFFSET_TO_I
 
 
-def munich_key(I, F, D):
-    return (I << _OFFSET_TO_I) | (F << _OFFSET_TO_F) | (D << _OFFSET_TO_D)
+def munich_key(Instr, F, D):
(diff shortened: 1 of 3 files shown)


spynnaker/pyNN/protocols/munich_io_spinnaker_link_protocol.py
```diff
@@ -43,16 +43,16 @@
 PUSH_BOT_MOTOR_UART_SHIFT = 0 + _OFFSET_TO_I
 
 
-def munich_key(I, F, D):
-    return (I << _OFFSET_TO_I) | (F << _OFFSET_TO_F) | (D << _OFFSET_TO_D)
+def munich_key(Instr, F, D):
+    return (Instr << _OFFSET_TO_I) | (F << _OFFSET_TO_F) | (D << _OFFSET_TO_D)
```
rowleya, line 47: Did you mean to make these changes?

andrewgait: flake8 has a new (I think) error for "ambiguous variable name", so it's not allowing the use of "I" here, for example.  That's why this has changed, though if preferable we could work out how to skip it instead.

rowleya: Ah OK, no problem, just wondered!

PR #825, opened 2020-05-27
Reload neuron parameters after setting initial value

andrewgait:
Most likely an oversight in the recent-ish multi-run / reset changes, as this is not a standard PyNN function.  This fixes the issue at https://github.com/SpiNNakerManchester/sPyNNaker8/issues/331.

Have added the script from @ej159 as a test in https://github.com/SpiNNakerManchester/sPyNNaker8/pull/409.

### DIFF
--- a/spynnaker/pyNN/models/neuron/abstract_population_vertex.py
+++ b/spynnaker/pyNN/models/neuron/abstract_population_vertex.py
@@ -627,9 +627,9 @@ class AbstractPopulationVertex(
                 " parameter {}".format(variable))
 
         parameter = self._get_parameter(variable)
-
         ranged_list = self._state_variables[parameter]
         ranged_list.set_value_by_selector(selector, value)
+        self.__change_requires_neuron_parameters_reload = True
 
     @property
     def conductance_based(self):


ej159:
I can confirm that this works with population.set_initial_value(), many thanks.
If running with population.set() I still get `InvalidParameterType` and `initialize can only be called before the first call to run` exceptions because it tries to call initialize(). Wouldn't it make more sense for set_initial() to be called at that point?

andrewgait:
The rules as set by PyNN are reasonably clear in that they don't want users to initialize state variables between runs without first calling reset() (which is why we have a separate non-PyNN function that does it).  As set() is a PyNN function, we take the view that we ought to make it behave in the way PyNN want it to work.

ej159:
> The rules as set by PyNN are reasonably clear in that they don't want users to initialize state variables between runs without first calling reset() (which is why we have a separate non-PyNN function that does it). As set() is a PyNN function, we take the view that we ought to make it behave in the way PyNN want it to work.

That makes sense. I'll use set_initial_value() then. Thanks.

PR #829, opened 2020-06-03
Try to switch Travis to Focal Fossa (Ubuntu 20.04LTS)

dkfellows:
This should let us pick up gcc-arm-none-eabi 9.2 or later, which produces tighter code...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ python:
   - 2.7
   - 3.6
   - 3.7
-dist: xenial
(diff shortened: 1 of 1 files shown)


dkfellows:
Note that the Travis docs do not mention support for Focal, but a few people online seem to have already switched.

andrewgait:
This was covered by https://github.com/SpiNNakerManchester/sPyNNaker/pull/826

PR #831, opened 2020-06-11
One dict

Christian-B:
Adds support for finding dict where release should put it.
Part of https://github.com/SpiNNakerManchester/SpiNNUtils/pull/114


alan-stokes:
approved once fec's issues solved

PR #701, opened 2019-09-04
Graph mapping sanity

dkfellows:
Downstream of SpiNNakerManchester/PACMAN#256

PR #836, opened 2020-06-30
Post graph-mapper-sanity corrections

dkfellows:
Equivalent of https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/617 but for sPyNNaker.

Note that these changes are formally entirely independent of that PR. Deliberately.


dkfellows:
This has now been recast so that it holds the corrections that I want to apply to `master` now that #701 is in.

The changes are _mostly_ about improving documentation and tidying up import order so that it follows general PEP8 rules and our own project order. But also:

* Adds a utility function, `ceildiv()`, for doing an efficient divide-and-round-up. Better than the constructs it replaces.
* Reworks the reporting code for on-chip bitfield compression to factor out the code to actually read things from SpiNNaker. This greatly simplifies the rest of the report generation in that file.

dkfellows:
Too awkward to fix

PR #840, opened 2020-07-14
changes to speed up matrix extraction massively.

alan-stokes:
this moves te setting of the system routes and the timeouts from a machine vertex level to the projection level. MASSIVELY speeds up matrix extraction

JENKINS: https://github.com/SpiNNakerManchester/sPyNNaker8/pull/419

### DIFF
--- a/spynnaker/pyNN/models/neuron/synaptic_manager.py
+++ b/spynnaker/pyNN/models/neuron/synaptic_manager.py
@@ -1313,7 +1313,7 @@ class SynapticManager(object):
             transceiver, placement, master_pop_table, indirect_synapses,
             direct_synapses, key, pre_vertex_slice.n_atoms, index,
             using_extra_monitor_cores, placements, monitor_api,
-            fixed_routes=fixed_routes)
+            extra_monitor, fixed_routes=fixed_routes)
 
         # Get the block for the connections from the delayed pre_vertex
         delayed_data = None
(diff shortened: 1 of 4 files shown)


spynnaker/pyNN/models/neuron/synaptic_manager.py
```diff
@@ -1322,7 +1313,7 @@ def get_connections_from_machine(
             transceiver, placement, master_pop_table, indirect_synapses,
             direct_synapses, key, pre_vertex_slice.n_atoms, index,
             using_extra_monitor_cores, placements, monitor_api,
-            extra_monitor, monitor_cores, fixed_routes=fixed_routes)
```
andrewgait, line 1325: extra_monitor should still be in the argument list here

alan-stokes: confused. everywhere else it was removing 2 elements. why would this one only need to remove 1? also wheres the handle_time_out_config? ug. 

but the tests pass.......... very very confused

andrewgait: The integration tests don't pass - check the equivalent spynnaker8 branch...

alan-stokes: doh. tis good to renember theres a 8 at the end of that repo. ;-)

alan-stokes: good catch by the way on the bug

PR #842, opened 2020-07-15
Determism

alan-stokes:
ork to make all networks more repeatable.


neural_modelling/makefiles/neuron/Makefile
```diff
@@ -38,9 +38,6 @@ MODELS = IF_curr_exp \
          IF_cond_exp_structural_last_neuron_distance_weight \
          IF_curr_exp_structural_last_neuron_distance_weight \
          IF_curr_exp_structural_random_distance_weight \
-         IF_cond_exp_stdp_mad_pair_additive_structural_last_neuron_distance_weight \
-         IF_cond_exp_stdp_mad_pair_additive_structural_random_distance_weight \
-         IF_curr_exp_stdp_mad_pair_additive_structural_random_distance_weight \
```
andrewgait, line 45: Just a warning that if you're going to get rid of all structural plasticity + STDP, then you'll have to do something about the examples and integration tests that currently use it as well.

alan-stokes: will do. will make them skipped. coz we cant keep just going up compilers. and given this problem will be properly solveable with self partitioning. coz i bet the first thing we should do is make all structural ones do the compartmental stuff.

dkfellows (approved):
> ork to make all networks more repeatable

Doesn't seem to be their style.

image

(Looks pretty good to me provided it passes integration tests.)

neural_modelling/src/neuron/c_main.c
```diff
@@ -272,6 +286,9 @@ void resume_callback(void) {
 //!            executed since start of simulation
 //! \param[in] unused: unused parameter kept for API consistency
 void timer_callback(uint timer_count, UNUSED uint unused) {
+    use(unused);
```
rowleya, line 289: I think the marker above means you don't need this

neural_modelling/src/neuron/spike_processing.c
```diff
@@ -257,32 +289,46 @@ static inline void setup_synaptic_dma_write(
...
+            log_warning("Could not trigger user event\n");
+        }
     }
 }
 
 //! \brief Called when a DMA completes
 //! \param unused: unused
 //! \param[in] tag: What sort of DMA has finished?
 static void dma_complete_callback(UNUSED uint unused, uint tag) {
+    use(unused);
```
rowleya, line 330: Not critical but not needed

rowleya (approved):
Very minor issues; change 'em if you wanna!

neural_modelling/src/neuron/c_main.c
alan-stokes: marker above? this has been needed for a long time for warning killing in the compiler

neural_modelling/src/neuron/spike_processing.c
alan-stokes: same question. has something changed that allows us to not have that?

neural_modelling/src/neuron/c_main.c
rowleya: Note the UNUSED in the line above

neural_modelling/src/neuron/spike_processing.c
rowleya: Note the UNUSED in the line above

alan-stokes: fixed

neural_modelling/src/neuron/c_main.c
alan-stokes: fixed

PR #843, opened 2020-07-20
bool flag for turning off iobuf for rerun expander

alan-stokes:
title

syn 8 link
https://github.com/SpiNNakerManchester/sPyNNaker8/pull/423

### DIFF
--- a/spynnaker/pyNN/spynnaker.cfg
+++ b/spynnaker/pyNN/spynnaker.cfg
@@ -9,6 +9,7 @@ read_bif_field_iobuf = True
 generate_bit_field_report = False
 generate_bit_field_summary_report = False
 write_router_compressor_with_bitfield_iobuf = True
+write_rerun_expander_iobuf = False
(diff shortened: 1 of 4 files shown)


spynnaker/pyNN/spynnaker.cfg
```diff
@@ -9,6 +9,7 @@ read_bif_field_iobuf = True
 generate_bit_field_report = False
 generate_bit_field_summary_report = False
 write_router_compressor_with_bitfield_iobuf = True
+write_rerun_expander_iobuf = False
```
rowleya, line 12: Seems odd to only control the IOBUF extraction of this part without also having IOBUF in the original synapse expander also use this.  This would then give a simpler variable name.

alan-stokes: you might want to only have the iobuf of the rerun. but if you want to accum them. itll have to happen when the first pr is merged. coz eitehr can do it.

PR #844, opened 2020-07-20
cleans up synaptic expander

alan-stokes:
gets the synaptic expander to use the system_control_logic.

syn8 tests: https://github.com/SpiNNakerManchester/sPyNNaker8/pull/424


spynnaker/pyNN/models/utility_models/synapse_expander/synapse_expander.py
```diff
@@ -33,7 +36,7 @@
 
 def synapse_expander(
         app_graph, placements, transceiver, provenance_file_path,
-        executable_finder):
+        executable_finder, extract_iobuf, no_sync_changes):
```
rowleya, line 39: This seems to take the number of sync changes just so it can be ignored later on.

spynnaker/pyNN/models/utility_models/synapse_expander/synapse_expander.py
```diff
@@ -61,34 +65,14 @@ def synapse_expander(
...
-
-        if not finished:
-            raise SpynnakerException(
-                "The synapse expander failed to complete")
+    run_system_application(
+        expander_cores, expander_app_id, transceiver, provenance_file_path,
+        executable_finder, extract_iobuf, functools.partial(
+            _fill_in_connection_data, placements=placements,
+            expanded_pop_vertices=expanded_pop_vertices),
+        None, [CPUState.FINISHED], False, no_sync_changes,
```
rowleya, line 74: Since needs_sync_barrier is False, no_sync_changes can be ignored.

rowleya:
Just realised that there might be an issue here: the new parameter hasn't been added to the XML (or the config file).

spynnaker/pyNN/models/utility_models/synapse_expander/synapse_expander.py
alan-stokes: so ignoring the fact your griping about a param from a FEC call. which is beyond this pr (and this pr is bust)

just so we're clear, your saying no_sync_changes should default to None in the FEC call, so that it can handle the False better?

rowleya: Given the bug report, it shouldn't be there at all...  From the perspective of this PR, there should be no "no_sync_changes" argument to the function at all, as it will be ignored inside anyway.

alan-stokes: > Just realised that there might be an issue here: the new parameter hasn't been added to the XML (or the config file).

did say the pr was broken. i noticed that and was about to fix it till other things distracted me. 

teach me for rushing this coz i WAS VERY HUNGRY!

alan-stokes:
> Just realised that there might be an issue here: the new parameter hasn't been added to the XML (or the config file).


did say the pr was broken. i noticed that and was about to fix it till other things distracted me. 

teach me for rushing this coz i WAS VERY HUNGRY!

PR #852, opened 2020-07-30
Weight scale

andrewgait:
Updates to use minimum weight values for scaling rather than calculating scales from ring buffers.

Merge at same time as
https://github.com/SpiNNakerManchester/DataSpecification/pull/87 and
https://github.com/SpiNNakerManchester/sPyNNaker8/pull/399


andrewgait:
@pabogdan and @hopper333 - @rowleya said you were interested in looking at this work, so here it is.  It's close to ready, I think, but any (further) ideas you have on the subject would be welcome, and if you have any examples that differ from the main codebase examples to test it on then even better.

dkfellows (approved):
Looks OK to me at a quick glance…

neural_modelling/src/neuron/synapse_row.h
```diff
@@ -232,20 +232,33 @@ static inline weight_t synapse_row_sparse_weight(uint32_t x) {
     return x >> (32 - SYNAPSE_WEIGHT_BITS);
 }
 
+////! \brief Converts a weight stored in a synapse row to an input
+////! \param[in] weight: the weight to convert in synapse-row form
+////! \param[in] left_shift: the shift to use when decoding
+////! \return the actual input weight for the model
+//static inline input_t synapse_row_convert_weight_to_input(
```
Christian-B, line 239: Are there future plans to use this new and commented out code

andrewgait: I think this is *old* commented out code rather than new since it has a left-shift in it.  I'm not sure how it stayed in but perhaps there was some confusion during a merge and it was kept somehow.

andrewgait: Thinking about it further I think this "conversion" was moved from here during the split neuron-synapse modelling, so it was probably when that was merged into this branch that the confusion happened.

Christian-B:
I am uncomfortable with getting min_weights from cfg

The way that the number of weight to must equal the number of synapse types in each population/ neuron_impl

This suggests the notion of a global  min_weights does not make sense.

andrewgait:
> I am uncomfortable with getting min_weights from cfg
> 
> The way that the number of weight to must equal the number of synapse types in each population/ neuron_impl
> 
> This suggests the notion of a global min_weights does not make sense.
> 
> max_stdp_spike_delta weight_random_sigma
> 
> May make sense if these are not expected to change often.

I think you're probably right on balance - and I think ideally the other values are not necessarily global either.  These should probably be made into Population / vertex-level functions rather than coming from cfg.

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
```diff
@@ -828,204 +890,174 @@ def reset_to_first_timestep(self):
...
+                weight_min = DataType.S1615.decode_from_int(1)
+            weight_min *= weight_scale
+            if not numpy.isnan(weight_min):
+                if min_weights[synapse_type] != sys.maxsize:
+                    weight_min = float_gcd(
+                        min_weights[synapse_type], weight_min)
+                min_weights[synapse_type] = min(
+                    min_weights[synapse_type], weight_min)
+
+            if isinstance(synapse_dynamics, SynapseDynamicsSTDP):
```
Christian-B, line 945: It would be cleaner if this code could be pushed into the synapse_dynamics rather than doing different stuff here based on the isinstance.

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
```diff
@@ -828,204 +890,174 @@ def reset_to_first_timestep(self):
...
+                max_weight = self.__get_closest_weight(weight_max)
+                min_weights[synapse_type] = max_weight / weight_scale_limit
 
-        :param int ring_buffer_to_input_left_shift:
-        :rtype: float
+        self.__check_weights(min_weights, weight_scale, incoming_projections)
+
+        return min_weights
+
+    def __check_weights(
```
Christian-B, line 994: This code does not appear to run on user supplied min_weights

Would it not be better if it did.

spynnaker/pyNN/models/neuron/plasticity/stdp/common.py
```diff
@@ -51,3 +51,41 @@ def get_exp_lut_array(time_step, time_constant, shift=0):
...
+    """ Get the smallest non-zero value of an exponential lookup array,\
+        or None if no such value
+
+    :param numpy.ndarray exp_lut_array: The lookup array
+    :param float time_step: The time step in milliseconds
+    :param float max_stdp_spike_delta: The maximum expected difference between
+        spike times in milliseconds
+    :rtype: float
+    """
+    if not len(exp_lut_array):
```
Christian-B, line 77: Just do:
if not exp_lut_array:

spynnaker/pyNN/models/neuron/plasticity/stdp/weight_dependence/weight_dependence_additive_triplet.py
```diff
@@ -145,14 +146,24 @@ def write_parameters(
...
         return self.__w_max
 
+    @property
+    @overrides(AbstractWeightDependence.weight_minimum)
+    def weight_minimum(self):
+        return self.__w_min
+
+    @overrides(AbstractWeightDependence.weight_change_minimum)
+    def weight_change_minimum(self, min_delta):
+        pot, dep = min_delta
```
Christian-B, line 160: (a2_plus, a3_plus), (a2_minus, a3_minus) = min_delta

spynnaker/pyNN/models/neuron/synapse_dynamics/abstract_synapse_dynamics.py
```diff
@@ -150,7 +150,7 @@ def get_delay_variance(self, connector, delays, synapse_info):
         """ Get the variance in delay for the synapses
 
         :param AbstractConnector connector:
-        :param ~numpy.ndarray delays:
```
Christian-B, line 153: Delays param still used so why remove docs?

spynnaker_integration_tests/test_various/test_alltoone_with_large_weight.py
```diff
@@ -40,7 +41,10 @@ def do_run(self):
         sim.end()
 
         weight_sum = sum(weight[2] for weight in weight_list)
-        self.assertEqual(weight_sum, sources * weights)
+        # 50.0 is not exactly representable so work out the actual value
+        weight_used = 1 / (
```
Christian-B, line 45: Would AlmostEqual not work even without this adjustment.

I would prefer that as then you are testing without the same adjustment as in the code.

andrewgait: It would work - I'd just need to work out the correct tolerance for it...

spynnaker/pyNN/models/neuron/plasticity/stdp/common.py
andrewgait: I'm not sure this is possible - but also not sure that this check is necessary.  This function is only ever called if exp_lut_array exists, so I'm not sure what this was checking.

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
andrewgait: Fixed this - now runs on user supplied values (via population additional_parameters).

spynnaker/pyNN/models/neuron/plasticity/stdp/common.py
andrewgait: I don't think it can be empty provided that STDP_FIXED_POINT_ONE is a sensible value... but I suppose it could be possible that it's empty.  But we don't have any examples where it is empty, and I suspect if it does return None then it'll lead to a crash elsewhere because it'll be trying to multiply by the returned None value.

spynnaker/pyNN/models/neuron/plasticity/stdp/weight_dependence/weight_dependence_additive_triplet.py
andrewgait: Done.

spynnaker/pyNN/models/neuron/synapse_dynamics/abstract_synapse_dynamics.py
andrewgait: Docs added back in.

spynnaker/pyNN/models/neuron/plasticity/stdp/common.py
rowleya: If None fails elsewhere, better to fail early and pick up the real error then

Christian-B: Agreed. if not exp_lut_array:  fails   
Had not seen it was a numpy array! Sorry

A numpy purest would say do:
if not exp_lut_array.size
but len(exp_lut_array) works too

I also 100% agree don't return None if calling code can not handle None!

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
andrewgait: Code moved into relevant synapse dynamics.

spynnaker_integration_tests/test_various/test_alltoone_with_large_weight.py
andrewgait: Done now.

PR #853, opened 2020-08-03
Remove gnu functions

rowleya:
Removes type conversion and division making the code size much smaller.  Ideally the rule that encodes distance might use a look-up-table but because the size of this is unknown it might have to live in SDRAM, making it less ideal.  Hence the division is now replaced by reciprocal multiplication.

### DIFF
--- a/spynnaker/pyNN/models/neuron/structural_plasticity/synaptogenesis/formation/distance_dependent_formation.py
+++ b/spynnaker/pyNN/models/neuron/structural_plasticity/synaptogenesis/formation/distance_dependent_formation.py
@@ -16,6 +16,7 @@
 import numpy
 from spinn_utilities.overrides import overrides
 from .abstract_formation import AbstractFormation
+from data_specification.enums.data_type import DataType
 
 
 class DistanceDependentFormation(AbstractFormation):
@@ -146,6 +147,8 @@ class DistanceDependentFormation(AbstractFormation):
     @overrides(AbstractFormation.write_parameters)
     def write_parameters(self, spec):
         spec.write_array(self.__grid)
+        spec.write_array(DataType.S031.encode_as_numpy_int_array(
+            1 / self._grid))
         spec.write_value(len(self.__ff_distance_probabilities))
         spec.write_value(len(self.__lat_distance_probabilities))
         spec.write_array(self.__ff_distance_probabilities.view("<u4"))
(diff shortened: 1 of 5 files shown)


rowleya:
I have run the structural plasticity tests on this and they all pass.  Integration testing is probably therefore a bit overkill since this only affects those cases.

alan-stokes:
nay. the number of times murph has seen that kinda statement and engineered to hit you. whats the harm in running them to defeat the murph?

rowleya:
Fine - tested in SpiNNakerManchester/sPyNNaker8#431

rowleya:
Tests appear to pass now...

spynnaker/pyNN/models/neuron/structural_plasticity/synaptogenesis/formation/distance_dependent_formation.py
```diff
@@ -16,6 +16,11 @@
 import numpy
 from spinn_utilities.overrides import overrides
 from .abstract_formation import AbstractFormation
+from data_specification.enums.data_type import DataType
+from spinn_front_end_common.utilities.constants import BYTES_PER_WORD
+
+# Number of words in the parameter struct excluding VLAs
+_WORDS_IN_STRUCT = 6
```
alan-stokes, line 23: habit here is to add to the comment what those elements are. and usually its bytes and you do the mul here in the define and not on line 78

spynnaker/pyNN/models/neuron/structural_plasticity/synaptogenesis/formation/distance_dependent_formation.py
```diff
@@ -146,6 +152,11 @@ def distance(self, x0, x1, metric):
     @overrides(AbstractFormation.write_parameters)
     def write_parameters(self, spec):
         spec.write_array(self.__grid)
+        # Work out the reciprocal, but zero them if >= 1 as they are not
+        # representable as S031 in that case, and not used anyway
+        recip = 1 / self.__grid
+        recip[recip >= 1.0] = 0
+        spec.write_array(DataType.S031.encode_as_numpy_int_array(recip))
```
alan-stokes, line 159: id argue that recip is a param that you should do at init. so its just a write_array here.

rowleya: Not sure I agree; this allows grid to be changed if desired (yes I know it doesn't just now but I could see a future change where it becomes settable).  I don't think this is called more than once per run so not a big deal to do it here.

spynnaker/pyNN/models/neuron/structural_plasticity/synaptogenesis/formation/distance_dependent_formation.py
rowleya: OK, fixed

spynnaker/pyNN/models/neuron/structural_plasticity/synaptogenesis/formation/distance_dependent_formation.py
alan-stokes: i had the same debate after posting it. so fine to leave it be.

PR #855, opened 2020-08-04
rejig to make external device lif control better for tdma

alan-stokes:
a rejig

https://github.com/SpiNNakerManchester/sPyNNaker8/pull/432 jenkins


spynnaker/pyNN/external_devices_models/external_device_lif_control.py
```diff
@@ -20,12 +20,13 @@
 from spynnaker.pyNN.models.defaults import default_initial_values,\
     default_parameters
 from spynnaker.pyNN.models.neuron.input_types import InputTypeCurrent
-from spynnaker.pyNN.models.neuron.neuron_models import (
-    NeuronModelLeakyIntegrateAndFire)
 from spynnaker.pyNN.models.neuron.synapse_types import SynapseTypeExponential
 from .external_device_lif_control_vertex import ExternalDeviceLifControlVertex
 from .threshold_type_multicast_device_control import (
     ThresholdTypeMulticastDeviceControl)
+from ..models.neuron.neuron_models.\
```
rowleya, line 27: 1) Don't use relative imports like this.  Only import from the current folder, or else go for the full thing.
2) Why do you need this?  You are using standard LIF in the C code...

spynnaker/pyNN/models/neuron/neuron_models/neuron_model_external_device_control.py
```diff
@@ -0,0 +1,230 @@
+# Copyright (c) 2017-2019 The University of Manchester
```
rowleya, line 1: I'm confused about the need for this entire file.  It looks like a clone of LIF.  Why?

neural_modelling/src/neuron/implementations/neuron_impl_external_devices.h
```diff
@@ -0,0 +1,572 @@
...
+                additional_inputs, soma_voltage);
+
+        // update neuron parameters
+        state_t result = neuron_model_state_update(
+                NUM_EXCITATORY_RECEPTORS, exc_input_values,
+                NUM_INHIBITORY_RECEPTORS, inh_input_values,
+                external_bias, this_neuron);
+
+        // determine if a spike should occur
+        has_spiked = threshold_type_is_above_threshold(the_threshold_type);
```
rowleya, line 427: I suspect this is misnamed; it isn't really a threshold type any more, and it isn't a spike.

alan-stokes: inherited from before, but will change.

spynnaker/pyNN/external_devices_models/external_device_lif_control.py
alan-stokes: coz i couldnt get the compiler to play ball without it.

rowleya: The python compiler?  You  should be able to use just the LIF model I believe

spynnaker/pyNN/models/neuron/neuron_models/neuron_model_external_device_control.py
alan-stokes: was when thinking a different route. killing

spynnaker/pyNN/external_devices_models/external_device_lif_control.py
alan-stokes: 1. was pycharm doing pycharmness. am correcting when replacing with old lif

alan-stokes: fixed

neural_modelling/src/neuron/implementations/neuron_impl_external_devices.h
alan-stokes: fixed

PR #856, opened 2020-08-04
There can be only one...

dkfellows:
... use of a random number source per selected random partner.

Note that this works because the original selected neuron ID just needs to be converted into a subpop/local ID pair. We simply don't need to draw another random number to achieve the choice of the second part; there was already sufficient random information in the first random number to provide an adequate selection. We just need to stop throwing the information away!

Note that this _does_ change the sequence of selections.


pabogdan:
I clearly haven't looked at this code in a while so I need to double check this: the change only affects the way a presynaptic partner is selected, not the way that the postsynaptic target is selected, 

In this case, this is fine. I don't know why there was an additional RNG there. 

I do feel it necessary to point out that the RNG used to select a postsynaptic neuron should still be used with a separate seed and only once per rewiring attempt per core.

dkfellows:
Thank you.

PR #859, opened 2020-08-12
Minor bugfixes

dkfellows:
Miscellaneous bits extracted from #857, mostly documentation tweaking to ensure that we always use the same active verb tense for brief descriptions of C functions.

### DIFF
--- a/neural_modelling/Doxyfile
+++ b/neural_modelling/Doxyfile
@@ -907,7 +907,7 @@ USE_MDFILE_AS_MAINPAGE =
 # also VERBATIM_HEADERS is set to NO.
 # The default value is: NO.
 
-SOURCE_BROWSER         = NO
+SOURCE_BROWSER         = YES
 
 # Setting the INLINE_SOURCES tag to YES will include the body of functions,
 # classes and enums directly into the documentation.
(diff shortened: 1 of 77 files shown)


neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c
```diff
@@ -270,7 +270,7 @@ bool synapse_dynamics_initialise(
 
     // Load timing dependence data
     address_t weight_region_address = timing_initialise(address);
-    if (address == NULL) {
+    if (weight_region_address == NULL) {
```
dkfellows, line 273: This change was an absolute howler. Fortunately, I think the compiler was proving in all cases that the branch couldn't be taken, but it's still absurd.

neural_modelling/Doxyfile
```diff
@@ -907,7 +907,7 @@ USE_MDFILE_AS_MAINPAGE =
 # also VERBATIM_HEADERS is set to NO.
 # The default value is: NO.
 
-SOURCE_BROWSER         = NO
+SOURCE_BROWSER         = YES
```
dkfellows, line 910: Generates a link to the (pretty-printed) source for all files, not just headers.

neural_modelling/src/common/neuron-typedefs.h
```diff
@@ -51,56 +47,46 @@ union _spike_t {
...
+//! The type of a spike
+typedef uint32_t spike_t;
+#endif /*SPIKES_WITH_PAYLOADS*/
 
-//! \brief helper method to retrieve the key from a spike
+//! \brief Retrieve the key from a spike.
 //! \param[in] s: the spike to get the key from
-//! \return key_t: the key from the spike
+//! \return the key from the spike
 static inline key_t spike_key(spike_t s) {
```
dkfellows, line 58: The key change I'm doing here is I'm trying to make it so that `spike_key` and `spike_payload` are only defined once; the differences are in the types they talk about and the body of the function. This makes the documentation process much nicer.

neural_modelling/src/common/neuron-typedefs.h
```diff
@@ -25,22 +30,13 @@
...
 #include "maths-util.h"
 
-// Determine the type of a spike
-/*
- * defines a spike with either a pay load or not and implements the
- * functionality to extract the key and pay load in both cases. If the
- * spike is compiled as not having a pay load, the pay load will always be
- * returned as 0
- */
-#ifndef __SPIKE_T__
```
dkfellows, line 35: Don't need this. The file already has a guard and only this file defines the type of spikes.

PR #861, opened 2020-08-13
Changes requested by Alan, done here first in case he doesn't like it

rowleya:
*** This is a PR on pop_based_master_pop_table, not master ***

This is the master pop table adding split in 2 for you to examine and decide if you like it.

### DIFF
--- a/spynnaker/pyNN/models/neuron/master_pop_table.py
+++ b/spynnaker/pyNN/models/neuron/master_pop_table.py
@@ -392,10 +392,27 @@ class MasterPopTableAsBinarySearch(object):
         self.__entries = dict()
         self.__n_addresses = 0
 
-    def update_master_population_table(
-            self, block_start_addr, row_length, key_and_mask, core_mask=0,
-            core_shift=0, n_neurons=0, is_single=False):
-        """ Add an entry in the binary search to deal with the synaptic matrix
+    def add_machine_entry(
+            self, block_start_addr, row_length, key_and_mask, is_single=False):
+        """ Add an entry for a machine-edge to the population table
+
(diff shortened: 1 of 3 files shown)


spynnaker/pyNN/models/neuron/master_pop_table.py
```diff
@@ -424,6 +441,31 @@ def update_master_population_table(
                 "The core mask of {} is too big (maximum {})".format(
                     core_mask, _MAX_CORE_MASK))
 
+        self.__update_master_population_table(
```
alan-stokes, line 444: make this the machine entry

spynnaker/pyNN/models/neuron/master_pop_table.py
```diff
@@ -424,6 +441,31 @@ def update_master_population_table(
...
+            The number of neurons in each machine vertex (bar the last)
+        :param bool is_single:
+            Flag that states if the entry is a direct entry for a single row.
+        :return: The index of the entry, to be used to retrieve it
+        :rtype: int
+        :raises SynapticConfigurationException: If a bad address is used.
+        """
+
         # pylint: disable=too-many-arguments, arguments-differ
         if key_and_mask.key not in self.__entries:
```
alan-stokes, line 470: move these checks into the app entry. and then you have what i was thinkign of

alan-stokes:
so when i looked on pycharm (as i was going to rewrite it to reflect my views, i found it was as i wanted it). i don't see why this is worse than what ya had. there's no cloned code, and now the context is correct, and no odd defaults in the method, coz they encapsulated within the context method. so id rather have this version than  whats in pop table currently.

Issue #850, opened 2020-07-22
orphan DelayExtensionVertex


Christian-B:
it is possible that all the edges into a DelayExtensionVertex are filtered out.

to repliacte:
import spynnaker8 as sim
sim.setup(timestep=1.0)
sim.set_number_of_neurons_per_core(sim.IF_curr_exp, 2)
sim.set_number_of_neurons_per_core(sim.SpikeSourceArray, 2)
pop_1 = sim.Population(4, sim.IF_curr_exp(), label="pop_1")
input = sim.Population(10, sim.SpikeSourceArray(spike_times=[0]), label="input")
input_proj = sim.Projection(input, pop_1, sim.OneToOneConnector(),
synapse_type=sim.StaticSynapse(weight=5, delay=21))
sim.run(40)

In https://github.com/SpiNNakerManchester/sPyNNaker/pull/847 there is a workaround which avoids errors by writing zeros for the unkown values.
See: https://github.com/SpiNNakerManchester/sPyNNaker/blob/0cd1c37d69ea54e77ba9e54367249b96bfd18a48/spynnaker/pyNN/models/utility_models/delays/delay_extension_vertex.py#L254

Christian-B:
I dont suggest trying to fix this until well into the self paritioning work

alan-stokes:
isnt the easy fix to this to do what EVERY other vertex does and has has key and key. adn then delay just checks the bool before firing?

alan-stokes:
pr that i think fixes it in a better way. coz saying key 0 is odd.....plus i would still fire with key 0. and i have no idea what that would do. likely lots of dropped packets.

Christian-B:
This needs to be fixed by avoiding that the DelayExtensionVertex is ever created in the first place.

Something that  can be done once delay extension are added on the fly rather than when projections are created,

PR #863, opened 2020-08-19
kfixes the key issue for filtered delays

alan-stokes:
fixes https://github.com/SpiNNakerManchester/sPyNNaker/issues/850

### DIFF
--- a/neural_modelling/src/delay_extension/delay_extension.c
+++ b/neural_modelling/src/delay_extension/delay_extension.c
@@ -54,6 +54,8 @@ struct delay_extension_provenance {
 };
 
 // Globals
+//! bool in int form for if there is a key
+static bool has_key;
 //! Base multicast key for sending messages
 static uint32_t key = 0;
 //! Key for receiving messages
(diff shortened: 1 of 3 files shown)


Christian-B:
When https://github.com/SpiNNakerManchester/sPyNNaker/issues/850 is fixed correctly this PR is useless.

The whole delay extension vertex should either never have existed or no longer exist.

IE. There should be no core wasted, no DSG/E time wasted for a vertex that has ZERO use!

Suggest closing this NOT MERGED!

alan-stokes:
no. this is needed to ensure that current MASTER delay extensions would never fire IF they got packets. and brings them in line with the rest of the tools with has key api

neural_modelling/src/delay_extension/delay_extension.c
```diff
@@ -156,6 +158,7 @@ static inline uint32_t round_to_next_pot(uint32_t v) {
 static bool read_parameters(struct delay_parameters *params) {
     log_debug("read_parameters: starting");
 
+    has_key = params->has_key;
```
rowleya, line 161: I guess this might happen if the filtering removes enough...

rowleya (approved):
Looks fine to me

rowleya:
I see some previous controversy.  The issue is that we can't currently remove a vertex after edge filtering has happened.  Having a has_key check doesn't take much effort so adding here is fine for now.  We could review has_key in general over the whole application after edge merging is done.  An issue has been added to review this later, but this needs merging to fix master now.

PR #864, opened 2020-08-20
unexpected fiddles

alan-stokes:
these were the only fiddles needed to allow the spillter object to travel down to basically app vertex level without issues. but given all in abstracts we control. not bad


spynnaker/pyNN/external_devices_models/external_device_lif_control.py
```diff
@@ -100,7 +100,7 @@ def __init__(
     def create_vertex(
             self, n_neurons, label, constraints, spikes_per_second,
             ring_buffer_sigma, incoming_spike_buffer_size,
-            n_steps_per_timestep):
+            n_steps_per_timestep, splitter_object):
```
Christian-B, line 103: I think splitter_object should be optional with a default to None.

While I agree that users are going to want the abilty to add it from the script that should not long term be the recommended way.

alan-stokes: well its optiomal in the sense that the pop level defaults say its None, from abstract_pynn_neuron_model.py

alan-stokes: so i dont see the point making it a default=none here. over the top

Christian-B: I dislike have required params where the docs then say they can be None.  Even worse where you break existing API by forcing all uses to just add ", None"   UGLY.

In this case it is even worse because it implies the prefer usage is to create the Spiltter Object before the Vertex Object when Long term that is the least desired way.   The Spiltter should be choose based on the information in the Vertex and not the other way araound!

alan-stokes: think after discussion, we're happy?

alan-stokes:
think this is dead now. the work was readded manually into the cherry pick already

Issue #865, opened 2020-08-20
additional params should be ingored. not force a blow up


alan-stokes:
the additional params on populations are meant to be for backends to have whatever they need. if the end user has added one we don't recognise (aka say a nest param) our code will currently blow about incorrect params for create vertex etc. 

the reason it was exposed was draft work for splitter objects
https://github.com/SpiNNakerManchester/sPyNNaker/pull/864

we should be able to just ignore them fully though. making the unexpected stuff redundant.

alan-stokes:
fixed by
https://github.com/SpiNNakerManchester/sPyNNaker/pull/867
https://github.com/SpiNNakerManchester/sPyNNaker8/pull/437

PR #867, opened 2020-08-20
fixes https://github.com/SpiNNakerManchester/sPyNNaker/issues/865

alan-stokes:
title
https://github.com/SpiNNakerManchester/sPyNNaker8/pull/437


alan-stokes:
> This looks like someone is trying to win a prize in how complcated can they write their code.
> 
> Just do:
> def _process_additional_params(...
> for key in additional_parameters:
> if key in population_parameters:
> population_parameters[key] = additional_parameters[key]
> else:
> logger....

i am shocked! me! do bells and whistles. the set logic was the first thought that came to mind. doing the multiple loops just seemed more complex than what this code says.......... and i note that i was avoiding all that 1. modifying end user data structs and the static dict adding merlarky. but up to you if ya want to change it.

alan-stokes:
fixed for ya. coz felt nice

PR #870, opened 2020-08-26
Dynamic cost models fix for DSG

Christian-B:
tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/440

Fix to make sure cost models use actual number of incoming machine_vertexes rather than an estimate based on max atom per core when reserving memory regions.

Question.
In https://github.com/SpiNNakerManchester/sPyNNaker/blob/9b6c0b700cc9834870a378bfb94c52baed2e816f/spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py#L226
method __get_structural_edges
The check if for multiple  AbstractSynapseDynamicsStructural on each EDGE but the error message is
 "Only one Projection between each pair of Populations can use structural plasticity"
So what is the actual limits.
1. One Dynamic per destination Population
2. One between source /destination pairs
3. One per edge
@pabogdan Do you know?

Also:
https://github.com/SpiNNakerManchester/sPyNNaker/blob/9b6c0b700cc9834870a378bfb94c52baed2e816f/spynnaker/pyNN/models/neuron/synapse_dynamics/abstract_synapse_dynamics_structural.py#L25
method get_structural_parameters_sdram_usage_in_bytes  has a parameter n_synapse_types which none of the known implementations use.  Is this parameter still needed and why?

### DIFF
--- a/spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py
+++ b/spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py
@@ -239,8 +239,8 @@ class SynapseDynamicsStructuralCommon(object):
                                   AbstractSynapseDynamicsStructural):
                         if found:
                             raise SynapticConfigurationException(
-                                "Only one SynapseInfo on an edge "
-                                "can use structural plasticity")
+                                "Only one Projection between each pair of "
+                                "Populations can use structural plasticity")
                         found = True
                         structural_edges.append((app_edge, synapse_info))
         return structural_edges
@@ -448,8 +448,8 @@ class SynapseDynamicsStructuralCommon(object):
                         if found_synapse_info and \
                                 found_synapse_info != synapse_info:
                             raise SynapticConfigurationException(
-                                "Only one SynapseInfo on each edge "
-                                "can use structural plasticity")
+                                "Only one Projection between each pair of "
+                                "Populations can use structural plasticity")
                         found_synapse_info = synapse_info
                 if found_synapse_info:
                     n_infos += 1
--- a/spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py
+++ b/spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py
@@ -20,7 +20,8 @@ from data_specification.enums.data_type import DataType
 from spinn_front_end_common.utilities.constants import (
     MICRO_TO_MILLISECOND_CONVERSION, MICRO_TO_SECOND_CONVERSION,
     BYTES_PER_WORD, BYTES_PER_SHORT)
-from spynnaker.pyNN.models.neural_projections import ProjectionApplicationEdge
+from spynnaker.pyNN.models.neural_projections import (
+    ProjectionApplicationEdge, ProjectionMachineEdge)
 from .abstract_synapse_dynamics_structural import (
     AbstractSynapseDynamicsStructural)
 from spynnaker.pyNN.exceptions import SynapticConfigurationException
@@ -238,8 +239,8 @@ class SynapseDynamicsStructuralCommon(object):
                                   AbstractSynapseDynamicsStructural):
                         if found:
                             raise SynapticConfigurationException(
-                                "Only one Projection between each pair of "
-                                "Populations can use structural plasticity")
+                                "Only one SynapseInfo on an edge "
+                                "can use structural plasticity")
                         found = True
                         structural_edges.append((app_edge, synapse_info))
         return structural_edges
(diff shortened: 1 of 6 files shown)


andrewgait:
First question, this works on application edges so it's what it says, only one structural plasticity Projection is allowed between a pair of Populations.

Second question, n_synapse_types looks like it was accidentally left over from a previous version of this function before the functionality was correctly separated out so that each part that could have parameters did their calculations separately.  So it can most likely safely be removed at this point.

Christian-B:
Question 1. One between source /destination pairs VS One per edge are actually the same thing as multiple projections between pairs of vertexes are merged into a single edge
.

pabogdan:
Re: sPyNNaker/spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py

The error message was put in place by @rowleya 13 months ago. It's unclear immediately why this is the case. We'd need to look at the C equivalent of that function. It is possible that some tables needed for structural plasticity are generated on pre-synaptic machine vertices, rather than for each structurally plastic projection. If this is the case, I think it should be changed. **I don't see a reason why you can't have Structural Plasticity on 2 different projections going from populations A -> B.** In fact, this is currently affecting research done by Gi. 

Structural plasticity adds some additional data to each core so that rewiring can happen. Some data is global (how often to rewire, formation and removal rules), but some data is projection specific. Without expanding what is currently stored in the projection specific data tables, multiple SP between the same pair of Populations should be supported if they have the same **global** data (I guess using .is_same_as).

I have an example that should work here:
2 populations A & B
A -> B via inh + struct. pl + STDP
A -> B via exc + structl. pl + STDP

should work if the global parameters for the structural plasticity rules are the same and if the global parameters for STDP are the same.


Re: question 2 
The signature for the method is `def get_parameters_sdram_usage_in_bytes(self, application_graph, app_vertex, n_neurons):` so I guess we don't use n_synapse_types.

pabogdan:
I've had a re-think and look at the underlying C. 

The data structure that holds the connectivity for each neuron (PostToPreTable or something) uses the (routing) key of the pre-synaptic population to identify it. Thus, 2 projections from the same pre-synaptic vertex would make entries in that table ambiguous.

Having the same key for 2 different entries in the PopulationInfo table ( https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py#L303-L363 ) would probably bias the formation rule if latest spike method is used. 

In conclusion, I think it's correct for the current implementation to disallow 2 different structurally plastic connections between the same pair of populations.

spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py
```diff
@@ -418,7 +419,53 @@ def __write_post_to_pre_table(
             numpy.concatenate(padded_rows).T, formats="u1, u1, u2").view("u4")
         spec.write_array(post_to_pre)
 
-    def get_parameters_sdram_usage_in_bytes(
+    def get_actual_parameters_sdram_usage_in_bytes(
```
pabogdan, line 422: As far as I can tell, this is the same as get_parameters_sdram_usage_in_bytes(). If this is called after partitioning is complete I would hope this would indeed return the actual SDRAM required. Have you run any tests for this?

pabogdan (approved):
I looked mostly at the Structural Plasticity changes. They seem alright to me. If testing passes this should be fine.

Christian-B:
> I've had a re-think and look at the underlying C.
> 
> The data structure that holds the connectivity for each neuron (PostToPreTable or something) uses the (routing) key of the pre-synaptic population to identify it. Thus, 2 projections from the same pre-synaptic vertex would make entries in that table ambiguous.
> 
> Having the same key for 2 different entries in the PopulationInfo table ( https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py#L303-L363 ) would probably bias the formation rule if latest spike method is used.
> 
> In conclusion, I think it's correct for the current implementation to disallow 2 different structurally plastic connections between the same pair of populations.

Thank you @pabogdan that is very useful.  The Projection code actually merges the connections between the same pair of populations into a single AppEedge and goes boon before this code is even called.

That 2 different source vertexes are allowed shows that the this pr is actually incorrect.

spynnaker/pyNN/models/neuron/synapse_dynamics/synapse_dynamics_structural_common.py
Christian-B: This was tested with https://github.com/SpiNNakerManchester/sPyNNaker8/pull/440
However I don't think any of these include the case of two  structurally plastic connections (each from a different source) 
into the same vertex
@pabogdan do you have an example script that has such a script.
Also if you have other scripts that you are keen to include in the test please add then either in the above spy8 pr or in a separate pr.

PR #873, opened 2020-09-01
fix for bitfields

alan-stokes:
this fixes the situation where the bitfield array (before any bitfields) cant be malloced (lack of dtcm, likely due to the master pop table being so big). This stops it from dabting, and allows the flow to keep operating. 

testing via : https://github.com/SpiNNakerManchester/sPyNNaker8/pull/443

### DIFF
--- a/neural_modelling/src/neuron/population_table/population_table_binary_search_impl.c
+++ b/neural_modelling/src/neuron/population_table/population_table_binary_search_impl.c
@@ -90,7 +90,7 @@ static uint16_t items_to_go = 0;
 static uint32_t bit_field_filtered_packets = 0;
 
 //! The bitfield map
-bit_field_t *connectivity_bit_field;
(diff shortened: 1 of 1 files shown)


neural_modelling/src/neuron/population_table/population_table_binary_search_impl.c
```diff
@@ -90,7 +90,7 @@ static uint16_t items_to_go = 0;
 static uint32_t bit_field_filtered_packets = 0;
 
 //! The bitfield map
-bit_field_t *connectivity_bit_field;
+bit_field_t *connectivity_bit_field = NULL;
```
rowleya, line 93: Not strictly required but OK

rowleya (approved):
Yep, looks like a good idea!

alan-stokes:
think the tests are just failing due to the dead chip issue for the sync signal. So I'd argue this is actually passing.

PR #882, opened 2020-09-15
Fix sub timestep STDP issues

rowleya:
This refactors a bit of code to make it work.  I'm not sure how it changed logically, but it makes more sense to me now, and it seems to do the right thing...


rowleya:
Tested by SpiNNakerManchester/sPyNNaker8#455.  A New test is added, but to make this fail(!) it depends on SpiNNakerManchester/SpiNNFrontEndCommon#660.

andrewgait:
This makes sense to me, so I'm happy for this to be merged (or to move into determinism branches, whichever is best - probably a good idea for it to be tested there).

andrewgait (approved):

rowleya:
Closing as now part of #842

PR #883, opened 2020-09-15
move get_ring_buffer_shifts to machine level

Christian-B:
More work moving stuff to machine level

1. Instead of going through the Application Edges (from the App Vertex) it now goes through the unique synapse_info objects found by checking the machine edges.
   - A bit of extra work to cover the normal cases but allows the flexibility that the Application Level synapse_information could be split to different (groups of) machine vertexes with DIFFERENT SynapticManager objects.
 - The result is still cached so the calculation is only done once per SynapticManager

2. Instead of hard coding that ONLY SpikeSourcePoisson will ever spike more than once per timestep, adds an AbstractMaxSpikes interface

Tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/456


Christian-B:
@alan-stokes thinks there is a better way so please hold off merging.

rowleya:
To add to the suggestions, note that the ring buffer shift can currently be calculated once per application vertex; there are a few such things in the code like this, so it is worth thinking about how to avoid repeating work.

Christian-B:
In fact the assumption in SynapticManager (in current master) is that there is one shift per SynapticManager instance.
So once calculated it is cached and reused.

rowleya:
Indeed; there is one Synaptic Manager per AbstractPopulationVertex though.  So you could just pass that object around I guess...

alan-stokes:
wasnt even looking at merging as of yet. coz of conflicts. ;-)

Christian-B:
With the compartmentalised work we are trying to break away as much as possible from the per Application Vertex model.  

While the one Synamptic manager for all the MachineVertext that come from a single App vertex will probably hold in nearly all case we want to allow the splitter to divide the SynapseManager. For example in Lucas stuff the part associated with the Poisson source  will go info neuron vertexes (if needed) and the other to the synapse vertexes,  And he may even split the ihn synapse from the exh synapse too.

rowleya:
Yes, this is true; I think the SynapticManager is an application vertex in Luca's code.  There is still a need for some caching at the application level though here I think, or else extra work will be required.  It might be possible of course to move the computation to be done just before the vertex is split, since I think it really only depends on application-vertex-level things.  Then the data could be passed down to the machine level vertex.

Christian-B:
@rowleya says " There is still a need for some caching at the application level though here I think"  
Reply the caching is already done by _get_ring_buffer_shifts which is the ONLY method that calls _get_ring_buffer_to_input_left_shifts

andrewgait:
Another consideration here (I'm really only commenting to remind myself that this is likely to need work when merging) is that _get_ring_buffer_shifts(...) is possibly going to disappear if the weight_scale work that uses min_weights rather than shifting looks promising enough to be merged.  Not something for you to worry about right now, though, particularly if this moving is easy enough to do now, then go ahead and do it.

rowleya:
I'm still not sure this is the right path.  When this is merged with the current master, the cache will work as expected.  However if you move synaptic manager to be based on the machine vertex, there will be a cache per machine vertex, so the work will be repeated.  The synapse_info is stored on the application edge, and will always have to be since that is where the information is available, so you can do this before splitting the vertex once you get to self-partitioning, and then pass the appropriate information down to all the machine vertices.  So keeping it at the application level is more likely to work.  An additional point is that I don't think you will have ProjectionMachineEdge once you implement self partitioning as I understand it, so this is likely to break there rather than help...

Christian-B:
This work like master does the calculation at the synapse info level. The only difference is that the synapse infos are obtained from the machine level.

rowleya:
Have talked this over on Skype... conclusion is that the SynapticManager will be shared between MachineVertex instances with the same synapse type.  This might be all of them if they all share synapse types of course, but as long as this is enforced elsewhere this is fine.

PR #885, opened 2020-09-21
Cache slices at MachineEdge creation time

Christian-B:
This code caches all slices in the connector if required.

It then asserts that they are the same as what is currently passed in by create_synaptic_block
 and gen_connector_params

Then once the test in .... pass

The slices params will be removed from create_synaptic_block
 and gen_connector_params


Christian-B:
WRONG! PyNN allows conncetors to be used multiple times so saving slices in Connectors is a NON NON!

PR #890, opened 2020-09-25
Switch scaling to using u3232 instead of s1615

andrewgait:
Found an example (thanks to @such-a-git) that doesn't work when using S1615 for scaling, so this switches to using U3232 instead.

Tested by (& merge with) https://github.com/SpiNNakerManchester/sPyNNaker8/pull/466


rowleya (approved):
Looks good and tests pass

PR #891, opened 2020-09-29
Run synchronized

rowleya:
Makes use of the calls in SpiNNakerManchester/SpiNNFrontEndCommon#666 and also adds the ability to call a run with the synchronization in place.


andrewgait (approved):
Happy with this once FEC PR is ready.

PR #898, opened 2020-10-13
Pre vertexes by app and partition name

Christian-B:
needed as https://github.com/SpiNNakerManchester/PACMAN/pull/308 now checks machine vertices have app vertices


Christian-B:
replaced

PR #899, opened 2020-10-14
all or nothing for machinevertex.appvertex

Christian-B:
With https://github.com/SpiNNakerManchester/PACMAN/pull/312
It now matters if the app grpah not passed to the machine graph


Christian-B:
needed because of down stream merge

PR #905, opened 2020-10-26
refactored get vertices methods so they are the same name as slice ones

Christian-B:
part of https://github.com/SpiNNakerManchester/PACMAN/pull/317


alan-stokes:
but the process outgoing edge partition fec function will check the machine vertex first

PR #917, opened 2020-11-03
avoid making a invalid error message

alan-stokes:


alan-stokes:
tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/488

Issue #927, opened 2020-11-16
AbstractReadParametersBeforeSet still on App vertex


Christian-B:
For example in SpikeSourcePoissonVertex

alan-stokes:
fixed by https://github.com/SpiNNakerManchester/sPyNNaker/pull/928
tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/497

PR #928, opened 2020-11-16
makes AbstractReadParametersBeforeSet apply to only machine verts

alan-stokes:
tested by 
https://github.com/SpiNNakerManchester/sPyNNaker8/pull/497

fixes https://github.com/SpiNNakerManchester/sPyNNaker/issues/927


spynnaker/pyNN/models/spike_source/spike_source_poisson_vertex.py
```diff
@@ -541,71 +521,14 @@ def kiss_seed(self, vertex_slice):
...
-            self.__data["rates"].set_value_by_id(
-                i,
-                spikes_per_tick *
-                (MICRO_TO_SECOND_CONVERSION / float(machine_time_step)))
-
-            # Store the updated time until next spike so that it can be
-            # rewritten when the parameters are loaded
-            self.__data["time_to_spike"].set_value_by_id(
-                i, time_to_next_spike)
+    def update_kiss_seed(self, vertex_slice, seed):
```
rowleya, line 524: Any reason that this information needs to be stored at the app vertex level rather than at the machine vertex level?  The machine vertex is by definition "by vertex slice" so storing there instead should work...

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
```diff
@@ -97,7 +95,7 @@ class AbstractPopulationVertex(
     # 5 elements before the start of global parameters
     # 1. has key, 2. key, 3. n atoms,
     # 4. n synapse types, 5. incoming spike buffer size.
-    _BYTES_TILL_START_OF_GLOBAL_PARAMETERS = 5 * BYTES_PER_WORD
+    BYTES_TILL_START_OF_GLOBAL_PARAMETERS = 5 * BYTES_PER_WORD
```
rowleya, line 98: I think this should be in the Machine vertex, as it is a property of that really.

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
```diff
@@ -317,7 +315,7 @@ def get_sdram_usage_for_neuron_params(self, vertex_slice):
         :return: The SDRAM required for the neuron region
         """
         return (
-            self._BYTES_TILL_START_OF_GLOBAL_PARAMETERS +
+            self.BYTES_TILL_START_OF_GLOBAL_PARAMETERS +
```
rowleya, line 318: I could see that this whole method would make a little bit more sense as part of the machine vertex.  That said, I can see the issue with having to ask the neuron impl for its sdram usage, making it harder.  The property could still come from the machine vertex though I think...

spynnaker/pyNN/models/neuron/population_machine_vertex.py
```diff
@@ -627,3 +628,35 @@ def gen_on_machine(self):
...
+        # locate SDRAM address to where the neuron parameters are stored
+        neuron_region_sdram_address = self.neuron_region_sdram_address(
+            placement, transceiver)
+
+        # shift past the extra stuff before neuron parameters that we don't
+        # need to read
+        neuron_parameters_sdram_address = (
+            neuron_region_sdram_address +
+            self._app_vertex.tdma_sdram_size_in_bytes +
+            self._app_vertex.BYTES_TILL_START_OF_GLOBAL_PARAMETERS)
```
rowleya, line 645: It seems odd to go to the trouble of moving this here, only to ask the app vertex anyway!  I have already suggested moving one of these properties above; would it make sense to move the other as well?

spynnaker/pyNN/models/neuron/population_machine_vertex.py
```diff
@@ -627,3 +628,35 @@ def gen_on_machine(self):
...
+
+        # get size of neuron params
+        size_of_region = self._app_vertex.get_sdram_usage_for_neuron_params(
+            vertex_slice)
```
rowleya, line 649: Like the above, this seems odd, especially as this machine vertex should know this, being the thing that represents the vertex_slice.

spynnaker/pyNN/models/neuron/population_machine_vertex.py
```diff
@@ -627,3 +628,35 @@ def gen_on_machine(self):
...
+        size_of_region -= (
+            self._app_vertex.BYTES_TILL_START_OF_GLOBAL_PARAMETERS +
+            self._app_vertex.tdma_sdram_size_in_bytes)
```
rowleya, line 652: Similar to comment above; marking to keep track

spynnaker/pyNN/models/neuron/population_machine_vertex.py
```diff
@@ -627,3 +628,35 @@ def gen_on_machine(self):
...
+
+        # get data from the machine
+        byte_array = transceiver.read_memory(
+            placement.x, placement.y, neuron_parameters_sdram_address,
+            size_of_region)
+
+        # update python neuron parameters with the data
+        self._app_vertex.neuron_impl.read_data(
+            byte_array, 0, vertex_slice, self._app_vertex.parameters,
+            self._app_vertex.state_variables)
```
rowleya, line 662: This is the ultimate call of all the app vertex stuff... any way this could come down here too?

rowleya (changes requested):
I've had a look at this.  My main issue is that all the machine vertex calls do is call the app vertex lots.  Is this by choice, or is this by definition something that is happening as a midway point?

spynnaker/pyNN/models/neuron/population_machine_vertex.py
alan-stokes: > I've had a look at this. My main issue is that all the machine vertex calls do is call the app vertex lots. Is this by choice, or is this by definition something that is happening as a midway point?

def a midway point. coz we dont know how olivers stuff is going to impact it. The current thinking is that the APp vertex is a DATA holder, and so neuron impl is safe staying up there. Given the different machine verts impl will still need the neuron params at some point. it seems sensible to keep that at the app level to avoid cloning code. Esp as the call now takes the dsg region as a input param, to allow the flex to the neuron core to put it whereever it wants to

spynnaker/pyNN/models/spike_source/spike_source_poisson_vertex.py
alan-stokes: highly possible, but out of scope of THIS pr. there's possibly lots of stuff that could move from app to machine vertex once all the stuff has been moved. but to avoid feature creep. it was not done in this pr

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
alan-stokes: its a property of the neuron impl, which is still inside the app vertex, as its data related, it was deemed to stay there so that olivers stuff could use it without clones

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
alan-stokes: goes against previous logic

spynnaker/pyNN/models/spike_source/spike_source_poisson_vertex.py
rowleya: OK, fair enough

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
rowleya: OK fair enough

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
rowleya: OK, fair enough

spynnaker/pyNN/models/neuron/population_machine_vertex.py
alan-stokes: yet again, the neuron impl is a data holder and so can live in the app vertex, asking the app vertex for data is considered perfectly fine. moving the BYTES_TILL_START_OF_GLOBAL_PARAMETERS doesnt on its own help, as the neuron impl is the one enforcing the tdma now,and as the app vertex is the one impling that api right now, you cant push the call for tdma size as its directly related to the app vertex.

spynnaker/pyNN/models/neuron/population_machine_vertex.py
rowleya: Yes, I think I was just hopeful.  I guess there isn't really an issue with the sharing, so long as the route to the data is consistent.

spynnaker/pyNN/models/neuron/population_machine_vertex.py
rowleya: Resolved above

spynnaker/pyNN/models/neuron/population_machine_vertex.py
alan-stokes: alas not, unless we moved the neuron impl INTO the machine vertex, and we have yet to do this. If we even want to go there

spynnaker/pyNN/models/neuron/population_machine_vertex.py
alan-stokes: refer to main view difference in impls.

spynnaker/pyNN/models/neuron/population_machine_vertex.py
rowleya: OK

spynnaker/pyNN/models/neuron/population_machine_vertex.py
rowleya: OK

rowleya (approved):

rowleya:
OK, on clarification that this isn't moving everything, this is fine.

PR #933, opened 2020-11-19
Enable specific documenting of __call__() methods

dkfellows:
That something is callable is an important thing in a lot of our code, so we should document it explicitly.


dkfellows:
Might interfere with other PRs. If so, those have higher priority; get me to fix this one. 😁

Christian-B (approved):
Ok to merge

PR #934, opened 2020-11-23
Update README.md to use new badgers

dkfellows:


dkfellows:
Only touches the README, nothing else!

Issue #886, opened 2020-09-21
verify FixedNumberPreConnector setting seed correctly


Christian-B:
FixedNumberPreConnector.gen_connector_params method

            self.__pre_connector_seed[key] = [
                int(i * 0xFFFFFFFF) for i in self._rng.next(n=4)]

@alan-stokes worries this could create a key outside of the acceptable range.

Same in FixedNumberPostConnector

alan-stokes:
ya also got it here:

https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/spynnaker/pyNN/models/neural_projections/connectors/abstract_generate_connector_on_machine.py#L97-L108

alan-stokes:
and
https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/spynnaker/pyNN/models/neural_projections/connectors/abstract_generate_connector_on_machine.py#L110-L128

andrewgait:
Do you mean a key outside the acceptable range, or a seed outside the acceptable range?  I'm not sure how the code that's there could do either of those...

alan-stokes:
this comes from the not using:
https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/spynnaker/pyNN/utilities/utility_calls.py#L285-L313

where the logic for when seed 1 = 0, or when seed[3] is > 698769069

alan-stokes:
also: https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/spynnaker/pyNN/models/neural_projections/connectors/abstract_connector.py#L351-L353

creates 1 int with the same * mask, but this is being used by a numpyRNG. what does the * 0x7FFFFFFF do in that scope? is this something that really needs to be mapped to a mar64 seed later on?

andrewgait:
> this comes from the not using:
> https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/spynnaker/pyNN/utilities/utility_calls.py#L285-L313
> 
> where the logic for when seed 1 = 0, or when seed[3] is > 698769069

I have to confess that I thought this was done in the C code for the expander (there's an equivalent function for this in spinn_common/random.c) but it would appear not.  So this probably does need fixing to validate the seeds, though I'm surprised that this hasn't caused problems before now...

alan-stokes:
guessing that if youve got the bad numbers the seed just inst very random. how many end users check our randomness?

PR #935, opened 2020-11-24
Validate mars kiss seeds before use

andrewgait:
Fixes https://github.com/SpiNNakerManchester/sPyNNaker/issues/886 (and a few more places where this issue was also discovered).

Tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/504 (which also needs merging as the tests using fixed seeds needed to be updated).


alan-stokes:
can approve this, but it begs a question. is the seed default better if its set to None in the utility function? 
also, if there is no rng, dont we want to report the seed of the rng somehow?

alan-stokes (approved):

andrewgait:
> can approve this, but it begs a question. is the seed default better if its set to None in the utility function?
> also, if there is no rng, dont we want to report the seed of the rng somehow?

I think the short answer to the first question is "Yes", but we should probably allow someone to set a seed when using this function if they really want to.

I'm not sure I understand your second question; numpy appears to cope fine with numpy.random.RandomState(None)...

alan-stokes:
> I think the short answer to the first question is "Yes", but we should probably allow someone to send in a seed when using this function if they really want to.

in this i meant. that in utility functiuons. change seed to seed=None. as it seems thats the default.

alan-stokes:
> I'm not sure I understand your second question; numpy appears to cope fine with numpy.random.RandomState(None)...

it occured to me. if you have no rng, and no seed. your getting a random seed. thus should the prov data report that seed so that future users could acutally reproduce the work?

andrewgait:
> > I'm not sure I understand your second question; numpy appears to cope fine with numpy.random.RandomState(None)...
> 
> it occured to me. if you have no rng, and no seed. your getting a random seed. thus should the prov data report that seed so that future users could acutally reproduce the work?

With no rng and no seed, you get a whole new rng (with seed=None).  The future user would then have to bypass this to directly set kiss_seed, wouldn't they?

andrewgait:
> > I think the short answer to the first question is "Yes", but we should probably allow someone to send in a seed when using this function if they really want to.
> 
> in this i meant. that in utility functiuons. change seed to seed=None. as it seems thats the default.

Done.

alan-stokes:
> > > I'm not sure I understand your second question; numpy appears to cope fine with numpy.random.RandomState(None)...
> > 
> > 
> > it occured to me. if you have no rng, and no seed. your getting a random seed. thus should the prov data report that seed so that future users could acutally reproduce the work?
> 
> With no rng and no seed, you get a whole new rng (with seed=None). The future user would then have to bypass this to directly set kiss_seed, wouldn't they?

you'd get a new rng. but i presume there's gotta be some seed for said rng, isnt None just a misnomer for random seed. So by putting said new seed in prov, the user can then get the information if they do desire a repeatable run by seeding their rngs the second time around.

alan-stokes:
> numpy.random.RandomState(None).

Random seed used to initialize the pseudo-random number generator. Can be any integer between 0 and 2**32 - 1 inclusive, an array (or other sequence) of such integers, or None (the default). If seed is None, then RandomState will try to read data from /dev/urandom (or the Windows analogue) if available or seed from the clock otherwise.
https://numpy.org/doc/1.13/reference/generated/numpy.random.RandomState.html

which says theres some seed, just where it gets said seed, is from some psuedo random location, and is os specific.

alan-stokes:
im not saying it needs to be done here. it was that here was what sparked the question.

andrewgait:
> > numpy.random.RandomState(None).
> 
> Random seed used to initialize the pseudo-random number generator. Can be any integer between 0 and 2**32 - 1 inclusive, an array (or other sequence) of such integers, or None (the default). If seed is None, then RandomState will try to read data from /dev/urandom (or the Windows analogue) if available or seed from the clock otherwise.
> https://numpy.org/doc/1.13/reference/generated/numpy.random.RandomState.html
> 
> which says theres some seed, just where it gets said seed, is from some psuedo random location, and is os specific.

Ah, okay.  Looking into this a bit further, it is possible to reproduce - you'd need to use the RandomState.get_state() function, put that into provenance somewhere, and then a user can get it back by building their own rng and using the set_state() function.

PR #936, opened 2020-11-24
Use shared actions

dkfellows:
This moves useful snippets into actions defined in our support repository.

I'm not interested in publishing them in the github marketplace.


dkfellows:
> Is it possible to move the `if: matrix.python-version == 3.6` to the SupportScripts action.

I'm currently experimenting with what exactly I can do. It's very much _not everything_ from what I can see…

dkfellows:
See also SpiNNakerManchester/SupportScripts#25

Issue #897, opened 2020-10-13
synaptic manager dsg regions being set in hardcoded manner


alan-stokes:
for future support of slef parittioning, we should ensure these are parameters of the manager, not hardocded. as a synapse core ant going to want dsg regions .....6 onwards

alan-stokes:
fixed by https://github.com/SpiNNakerManchester/sPyNNaker/pull/937

dkfellows:
Is there any evidence that we will ever actually have varying region IDs? Or is this just prepping for a future that never comes?

alan-stokes:
well theres 2 avenues. a conways app graph using a genral synaptic manager for edge data. should not be limtied to the pynn region ids. But further more, the synapse cores of the luca modle will highly likely not have the neuron params and thus would want different region ids. if we follow convention of going from 0 upwards, they would be lower numbers for them

alan-stokes:
the ear vertex's which recievied from other neuron modles, had to contort themselves to supprot those fixed region ids. this way they no longer would need to contort

Christian-B:
Will not fix as in discussion of https://github.com/SpiNNakerManchester/sPyNNaker/pull/937

PR #937, opened 2020-11-24
removes hard coded region ids

alan-stokes:
fixes hard codes

tested by:
https://github.com/SpiNNakerManchester/sPyNNaker8/pull/505

fixes: https://github.com/SpiNNakerManchester/sPyNNaker/issues/897


rowleya (changes requested):
This seems very heavy - the synaptic manager has variables already for the regions, but these are "hard coded" in the constructor.  Wouldn't it be simpler to pass the values in there?

alan-stokes:
> This seems very heavy - the synaptic manager has variables already for the regions, but these are "hard coded" in the constructor. Wouldn't it be simpler to pass the values in there?

nope. coz the constuctor is a app level thing. And the regions can be different for different machine verts.

spynnaker/pyNN/models/neuron/synaptic_manager.py
```diff
@@ -178,35 +159,49 @@ def __get_synaptic_matrices(self, post_vertex_slice):
         matrices = SynapticMatrices(
             post_vertex_slice, self.__n_synapse_types,
             self.__all_single_syn_sz, self.__synapse_io,
-            self._synaptic_matrix_region, self._direct_matrix_region,
-            self._pop_table_region)
+            synaptic_matrix_region, direct_matrix_region,
+            pop_table_region)
```
rowleya, line 163: It seems quite dangerous to have this cache along with this change.  The SynapticMatrices object is actually a per-machine-vertex level object already.  Would it instead make sense to store this in the appropriate machine vertex?

spynnaker/pyNN/models/neuron/synaptic_manager.py
```diff
@@ -736,23 +768,33 @@ def get_connections_from_machine(
         for post_vertex in progress.over(post_vertices):
             post_slice = post_vertex.vertex_slice
             placement = placements.get_placement_of_vertex(post_vertex)
-            matrix = self.__get_synaptic_matrices(post_slice)
+            matrix = self.__get_synaptic_matrices(
+                post_slice, pop_table_region, synaptic_matrix_region,
+                direct_matrix_region)
```
rowleya, line 773: An example of how dangerous this could be.  If you pass in different values here than in the previous call to __get_synaptic_matrices, you will not get the thing you think you will.  The region arguments are actively ignored if the object exists.

spynnaker/pyNN/models/neuron/synaptic_manager.py
alan-stokes: fair cop. will remove the creating of the cache during generate sdram (which is the second of dsg, and get sdram which do the caching). I think we need to cache them so that clear_connection_cache can work correctly.

spynnaker/pyNN/models/neuron/synaptic_manager.py
alan-stokes: if the sdram hasnt cached, this problem goes away

alan-stokes:
we could also remove the cache completely from the synaptioc_matrix's i THINK. just checking now

rowleya:
I'm not sure you want to do that; they store things like the positions in memory to read back from.

alan-stokes:
no you miss my point. the 2/3 region ids that are STORED in these synaptic_matrices. they used for finish master pop and get connections from machine. both of which, go via the synaptic manager first. so the caching of those region ids might be redundant. at which point if that wasnt cached, wouldnt it solve the problem?

alan-stokes:
the other solution. is to remove the calls to the PMV and make them function calls. which then means you cant put in the wrong number

alan-stokes:
which was what i was planning to do once this came into slef partitioning. but ic an push it here first

rowleya:
Ah Ok, you mean that you then need to add more inputting of region ids... I guess that would work, although you may find there are a lot of places to pass them around in there.  My comment about these being machine-vertex based in any case would mean that the actual machine vertex that needs to use them could store them.  This could then give it the region ids it needs at initialisation, on the basis that you know they are then not going to change from the machine vertex point.  I think a lot of methods are directly passed through from the synaptic manager anyway at this point, so that might make things neater overall.

rowleya:
A (temporary) alternative here would be to store them by machine vertex rather than by slice, to avoid issues with the slices being the same in multiple machine vertices (very much a potential issue here).  You could then *only* access the cache when you think it should already exist e.g. when reading connections back.

alan-stokes:
> My comment about these being machine-vertex based in any case would mean that the actual machine vertex that needs to use them could store them.

that breaks the idea of the synaptic manager beign the store of these things. which given these are all synaptic manager stuff. would be odd for the mac vertex to store them, and again, if they were not cached during sdram calc. then they going to be built during dsg, which def is machine vertex calls, and then its just reconsiling the other public methods. like get connections. which could be solved with a api of give me your master pop region id, etc etc. otherwise id remove the sdram cache and say GIGO

alan-stokes:
let me do some changes. on a path i think youd like

rowleya:
The SynapticMatrices objects are currently stored in the synaptic manager but are really machine-vertex-level objects, as attested to by the fact that they have the vertex slice in them.  There is one of these for each (current) machine vertex.  In the future with multiple synapse cores, I would expect to have one of these objects for each synapse machine vertex if that helps the thinking.

It is true that caching them during SDRAM calculations is a bit odd, especially if they are going to change later on.  This is again a feature of allowing a changing number of neurons per core during partitioning that makes this a pain; you want to know what size it would have for a given slice, but not really to store it at that point in case you don't end up using that slice.  A possible idea there is to make it not create the object, but ask at the class level for the potential size (i.e. in a static method), but I don't know how much that would break other things.

alan-stokes:
ive got a better idea. you dont need to put those region ids in at the sdram point at all. you can hand them to the matrixe's at dsg time. moving them to the machine vertex im not prepared to do right now. but it could be done if needed at a later date. but i think i can ensure you dont have the option to NOT put GIGO. 

as we dont have a clue about how luca is going to use this. we're flying blind. and i remind you baby steps

alan-stokes:
im building a api to get the data from the machine vertex, but give me about 1 hour to get it tied in. basically im going to do the work i was planning to do in slef partitioning merge here for ya

alan-stokes:
how about this @rowleya ???

rowleya (changes requested):
This is ok as an intermediate change.  I would like to see some way in the future to have the regions stored in objects.  I believe this is possible, though it will require some thinking about what can be represented at the machine level and what needs the application level, and how the two communicate.  This may mean that some things would move out of the synaptic manager and become application properties, such as the weight scaling properties, allowing the synaptic manager to be a machine level object in itself.

I have put this as "request changes" because I believe that Donal put quite a bit of work into not requiring all doc lines to end in "\" and this seems to break that again (a lot of the changes herein are doc changes that shouldn't be needed).

alan-stokes:
"shouldn't" be needed, but was needed to get 3.6 to pass. I yet again claim that it only checks files that are changed, and is demanding those \'s. You can see the errors it reported in the cycle of commits. I am yet unable to find any other way to make them go away par the \'s. and as you can see, there's doc changes on methods i haven't even touched, as voiced in the skype channel for days.

alan-stokes:
> I would like to see some way in the future to have the regions stored in objects. I believe this is possible, though it will require some thinking about what can be represented at the machine level and what needs the application level, and how the two communicate. This may mean that some things would move out of the synaptic manager and become application properties, such as the weight scaling properties, allowing the synaptic manager to be a machine level object in itself.

as much as i want to agree with you. we agreed to not do MASSIVE changes in these master related prs. coz the self partitioning changes makes a ton of stuff way way easier. Its the main reason i want them IN ASAP

PR #940, opened 2020-11-30
SynapseIO is all static!

rowleya:
I believe that all the methods in SynapseIO are actually static.  This means they can just be functions.  This will help to move things from App to Machine vertices in a future PR.


rowleya:
Note: Although there are lots of changes in synapse_io.py, all the changes should be removal of self

rowleya:
Obsoleted by #949.

PR #941, opened 2020-11-30
Get outgoing partition constraints2

Christian-B:
tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/509


Christian-B:
This PR is not correct.
The issue is that ExternalDeviceLifControlVertex which extends AbstractPopulationVertex overwrites the get_outgoing_partition_constraints method.

If you move get_outgoing_partition_constraints to PopulationMachineVertex then the method only gets called at the machine level and the overridden function is never called.

PR #943, opened 2020-12-01
add_application_entry should return a value

andrewgait:
Found when running a structural plasticity example (working on adding examples to cover the remaining binaries in https://github.com/SpiNNakerManchester/sPyNNaker8/issues/484) that for some reason catches that this function isn't returning the index when it should be (I don't know why the current tests/examples aren't catching this...).

Tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/513


rowleya (approved):
Oops!

PR #947, opened 2020-12-03
fiddled. not the cleanest, but heck

alan-stokes:
for ya @andrewgait


Christian-B:
Do we still need/want this?

I say nuke it

rowleya:
Looks to me to be a bit out-of-date anyway

andrewgait:
It was an idea that ended up not being used, so yes probably best to nuke it.

Christian-B:
nice idea but no longer needed

PR #949, opened 2020-12-04
Better synapse machine handling

rowleya:
This is a replacement for #937.  This allows things to have stored ids, but use them in the correct way.

A possible improvement would be to again combine the synaptic bits into a single place rather than have them all in the population machine vertex.  I would advise that this waits until there is also a synapse machine vertex for synapse cores, as this will then create an object that both things can share, and will indicate what the shared parts need to be in much more clarity.


rowleya:
Integration testing:
http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/better_synapse_machine_handling/1/pipeline/

rowleya:
Obsoleted by #954

PR #954, opened 2020-12-14
Synapse neuron splitting

rowleya:
Builds on (and obsoletes #949).

Splits up the synapse and neuron splitting in a way that is likely to be useful for future actual core splitting.  In particular:
 - Creates C header files for the neuron and synapse parts, and adds a "common" part too.
 - Moves some C functionality from the neuron to the synapse part to facilitate splitting.  This does mean the same thing is read in two different places (the number of neurons), but this is a minor addition for a major increase in functionality.
 - Fixes some oddities in the way the structural plasticity was done, allowing more autonomy in the implementation (i.e. timer updates for structural plasticity are now part of the structural plasticity module).
 - Creates new Python mix-ins that mirror the changes to the C code, with neuron, synapse and common parts.
 - Makes some things that were really static actually static (even more than in the previous branches).
 - Moves some common calculations back out of the splitter and into the ApplicationVertex, but without making it possible to do other things too.
 - Moves some things out of the NeuronRecorder to make them more generic, and adds a synapse_recorder property to the machine vertex, which is just another NeuronRecorder (but then keeps the concerns separate).
 - Fixes C code recording setup to allow the general recording setup to be done separately from the neuron recording setup.

Depends on SpiNNakerManchester/SpiNNFrontEndCommon#725

Tested by http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/synapse_neuron_splitting/3/pipeline/


spynnaker/pyNN/models/neuron/population_machine_common.py
```diff
@@ -0,0 +1,193 @@
...
+from spinn_front_end_common.utilities.constants import SIMULATION_N_BYTES
+
+
+# Identifiers for common regions
+CommonRegions = namedtuple(
+    "CommonRegions",
+    ["system", "provenance", "profile", "recording"])
+
+
+class PopulationMachineCommon(
```
alan-stokes, line 47: this cant be init'ed itself. thus its an abstract.

spynnaker/pyNN/models/neuron/population_machine_neurons.py
```diff
@@ -0,0 +1,230 @@
...
+    N_ITEMS = len(_fields_)
+
+
+# Identifiers for neuron regions
+NeuronRegions = namedtuple(
+    "NeuronRegions",
+    ["neuron_params", "neuron_recording"])
+
+
+class PopulationMachineNeurons(AbstractReadParametersBeforeSet):
```
alan-stokes, line 54: id think this is a abstract as well. it makes abstract properities afterall. thus cant be built itself

spynnaker/pyNN/models/neuron/population_machine_synapses.py
```diff
@@ -0,0 +1,479 @@
...
+
+# Identifiers for synapse regions
+SynapseRegions = namedtuple(
+    "SynapseRegions",
+    ["synapse_params", "direct_matrix", "pop_table", "synaptic_matrix",
+     "synapse_dynamics", "structural_dynamics", "bitfield_builder",
+     "bitfield_key_map", "bitfield_filter", "connection_builder"])
+
+
+class PopulationMachineSynapses(
```
alan-stokes, line 88: same problem.

spynnaker/pyNN/models/neuron/population_machine_common.py
rowleya: I think if you extend an abstract you don't have to put the abstract here too (it might even break things)

spynnaker/pyNN/models/neuron/population_machine_synapses.py
```diff
@@ -0,0 +1,479 @@
...
+from spynnaker.pyNN.models.neuron.synapse_dynamics import (
+    AbstractSynapseDynamicsStructural)
+from spynnaker.pyNN.utilities.utility_calls import get_n_bits
+from spynnaker.pyNN.utilities import bit_field_utilities
+from spynnaker.pyNN.models.abstract_models import AbstractSynapseExpandable
+
+from .synaptic_matrices import SynapticMatrices
+
+
+get_placement_details = \
```
alan-stokes, line 39: VOM! this is a static method. why not just make it public and import properly?!

spynnaker/pyNN/models/neuron/population_machine_neurons.py
rowleya: As above - inherits from abstract so abstraction will be checked

spynnaker/pyNN/models/neuron/population_machine_synapses.py
rowleya: Same thing - imports from abstract already

spynnaker/pyNN/models/neuron/population_machine_synapses.py
rowleya: The methods should definitely be made public here

spynnaker/pyNN/models/neuron/population_machine_synapses.py
```diff
@@ -0,0 +1,479 @@
...
+    ProvidesProvenanceDataFromMachineImpl._get_placement_details
+add_name = ProvidesProvenanceDataFromMachineImpl._add_name
```
alan-stokes, line 41: again, static method. just make it public

spynnaker/pyNN/models/neuron/population_machine_synapses.py
```diff
@@ -0,0 +1,479 @@
...
+    def connection_generator_region(self):
+        return self._synapse_regions.connection_builder
+
+    def get_connections_from_machine(
+            self, transceiver, placement, app_edge, synapse_info):
+        """ Get the connections from the machine for this vertex.
+
+        :param ~spinnman.transceiver.Transceiver transceiver:
+            How to read the connection data
+        :param ~pacman.model.placement.Placement placements:
```
alan-stokes, line 466: placement. not placements

alan-stokes:
get_placement_details = \
    ProvidesProvenanceDataFromMachineImpl._get_placement_details
add_name = ProvidesProvenanceDataFromMachineImpl._add_name

this is found in numerious places. should just make them public, as they static methods afterall

spynnaker/pyNN/models/neuron/master_pop_table.py
```diff
@@ -315,57 +316,59 @@ class MasterPopTableAsBinarySearch(object):
...
+        for proj in incoming_projections:
+            in_edge = proj._projection_edge
+
+            # If we haven't seen this edge before, add it in
+            if in_edge not in seen_edges:
+                seen_edges.add(in_edge)
+                vertex = in_edge.pre_vertex
+
+                # If the pre-vertex not been partitioned, estimate
+                if not vertex.machine_vertices:
```
alan-stokes, line 348: that is not a great check for not being partitioned. and givne the vertex.splitter.get_out_going_slices() returns 2 elements. 1 the slices, but 2 a BOOL THAT SAYS ITS AN ESTIMATE OR NOT! 
so line 349 is now a length of a tuple. but also, means you could use that bool. or heck, you could just use the line 348 for either.

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
```diff
@@ -43,6 +69,14 @@
...
+# 1 for number of neurons
+# 1 for number of synapse types
+# 1 for number of neuron bits
+# 1 for number of synapse type bits
+# 1 for drop late packets,
+# 1 for incoming spike buffer size
+_SYNAPSES_BASE_SDRAM_USAGE_IN_BYTES = 6 * BYTES_PER_WORD
+
 
 class AbstractPopulationVertex(
```
alan-stokes, line 81: i note that we cann this abstract, but we dont put the annotation for abstract

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
```diff
@@ -571,8 +755,557 @@ def reset_to_first_timestep(self):
...
+        if weight_std_dev > 0:
+            # pylint: disable=no-member
+            lngamma = special.gammaln(1 + upper_bound)
+            gammai = special.gammaincc(
+                1 + upper_bound, average_spikes_per_timestep)
+
+            big_ratio = (math.log(average_spikes_per_timestep) * upper_bound -
+                         lngamma)
+
+            if -701.0 < big_ratio < 701.0 and big_ratio != 0.0:
```
alan-stokes, line 824: magic numbers. and not even the baconised ones

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
```diff
@@ -571,8 +755,557 @@ def reset_to_first_timestep(self):
...
+    def __get_weight_scale(ring_buffer_to_input_left_shift):
+        """ Return the amount to scale the weights by to convert them from \
+            floating point values to 16-bit fixed point numbers which can be \
+            shifted left by ring_buffer_to_input_left_shift to produce an\
+            s1615 fixed point number
+
+        :param int ring_buffer_to_input_left_shift:
+        :rtype: float
+        """
+        return float(math.pow(2, 16 - (ring_buffer_to_input_left_shift + 1)))
```
alan-stokes, line 960: mmmmm 16 rashers of bacon?

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
```diff
@@ -571,8 +755,557 @@ def reset_to_first_timestep(self):
...
+
+    @property
+    def synapse_recordables(self):
+        """ Get the names of variables that can be recorded by the synapses
+
+        :rtype: list(str)
+        """
+        return self.__synapse_recorder.get_recordable_variables()
+
+    def get_common_constant_sdram(self, n_record, n_provenance):
```
alan-stokes, line 1173: everything below this is really NOT meant to be in a APP vertex. But in the given splitter. coz the APP vertex has no authority to state that the synapse core's sdram usage is x. thats up to the splitter whom is making those machine verts.

rowleya: This is a utility method to avoid repetition in all the splitters.  I understand the concern, but there is nothing that makes a splitter call this if it wants to do something different.

andrewgait:
Obsoleted by https://github.com/SpiNNakerManchester/sPyNNaker/pull/1082, which is now merged, so closing.

PR #958, opened 2021-01-05
Merge sPyNNaker8

dkfellows:
We've talked about doing this for a while.

1. The code in the `spynnaker8` directory has been moved over wholesale from the sPyNNaker8 repo with no changes.
2. Some minor changes have been done in the `spynnaker` directory (relating to code version)
3. The code in the `unittests` directory has been merged, I hope, and simply moved wherever possible.
4. The integration tests have been left exactly where they are.
5. I believe I've correctly merged the common files.

Next:
- [x] testing this!
- [x] doing an equivalent PR on the sPy8 side to remove the moved stuff.
- [ ] removing the trivial subclasses.
- [ ] get rid of the weird hacks put in because the code was separated.


Christian-B:
I would recommend not doing the " doing an equivalent PR on the sPy8 side to remove the moved stuff." until

1. This has been merged in.
2. All PR in spy8 are migrated or marked as never to merge

dkfellows:
I'll need the branch to help sort out testing issues. 😁

dkfellows:
Need to fix the docs… but otherwise things are now working here.

dkfellows:
The sPy8 side is “done” in a PR.

Christian-B:
Is the plan in later prs to start moving the code out of the spynakker8 directory into the spynakker directory?

Most likely leaving just thin wrappers behind for reverse compatibility?

If not I am not sure what this gains!

dkfellows:
I think the first step will be to get rid of some of the existing thin wrappers.

Christian-B:
> I think the first step will be to get rid of some of the existing thin wrappers.

The first step is get this already to big PR into master.

The thin wrappers are a challenge as we have two conflicting interests.
1. Avoid breaking other users stuff.
2. Cleaning up the mess causes by the need to support two pynn versions at once

A compromise may be to move the code / remove the wrappers but leave behind files which when imported print out instructions on how to update the code and then die.

Christian-B:
https://github.com/SpiNNakerManchester/sPyNNaker/tree/merge-spy8/p8_integration_tests reviewed. 
This is a straight copy and paste with a tiny number of changes due to keeping up to date with later merges and the moving of PopulationView and FromFileConnector classes.

Christian-B (approved):
While there are changes that should happen ASAP. File name class name rematching and other which should be done soon I believe this PR should go in as soon as tested by https://github.com/SpiNNakerManchester/IntegrationTests

PR #960, opened 2021-01-07
Warn if user supplies spike time for SSA that's too low

andrewgait:
Fixes https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/529

An integration test can probably be written to check this works if desired?

### DIFF
--- a/spynnaker/pyNN/models/spike_source/spike_source_array_vertex.py
+++ b/spynnaker/pyNN/models/spike_source/spike_source_array_vertex.py
@@ -98,6 +98,15 @@ class SpikeSourceArrayVertex(
 
         """
         time_step = self.get_spikes_sampling_interval()
+        current_time = globals_variables.get_simulator().get_current_time()
+        # warn the user if they are asking for a spike time out of range
+        for n in range(len(spike_times)):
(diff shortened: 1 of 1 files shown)


rowleya (approved):
The code looks fine.  I guess some sort of test would be good to ensure the message formats correctly.  Even a virtual unit test should be possible here I think...

Christian-B:
> The code looks fine. I guess some sort of test would be good to ensure the message formats correctly. Even a virtual unit test should be possible here I think...

I think this could be done using and extending the MockSimulator. This will probably come up in many places so will be work that I do as part of the coverage increasing.  Shall I start that for this?

andrewgait:
> > The code looks fine. I guess some sort of test would be good to ensure the message formats correctly. Even a virtual unit test should be possible here I think...
> 
> I think this could be done using and extending the MockSimulator. This will probably come up in many places so will be work that I do as part of the coverage increasing. Shall I start that for this?

Sure - go ahead.  The other way of doing this would be something similar to the test_hello_world integration test in GFE that uses log capture to make sure the expected output is being logged.

PR #961, opened 2021-01-11
input_type is not always const in this function (see example in NMT)

andrewgait:
More cleaning up of compiler warnings - in this case, an example in the NewModelTemplate requires this struct parameter to be non-const as it allows it to be modified.

Merge alongside https://github.com/SpiNNakerManchester/sPyNNaker8NewModelTemplate/pull/64
Tested by https://github.com/SpiNNakerManchester/sPyNNaker8/pull/525


rowleya (approved):
Seems reasonable; not sure what the issue with is with const vs. not const in this context, but it looks OK.

PR #963, opened 2021-01-13
remove MundyOnChipRouterCompression in commented out reference

Christian-B:


rowleya (approved):
Seems fine - not sure if we want to change the default to pair?

Christian-B:
> Seems fine - not sure if we want to change the default to pair?

Changing to pair was not part of this PR but a good idea.

PR #972, opened 2021-01-25
unmutable models

alan-stokes:
fixes models to be unmutable and fixes docs accordingly

### DIFF
--- a/spynnaker/pyNN/models/neuron/additional_inputs/additional_input_ca2_adaptive.py
+++ b/spynnaker/pyNN/models/neuron/additional_inputs/additional_input_ca2_adaptive.py
@@ -37,9 +37,15 @@ class AdditionalInputCa2Adaptive(AbstractAdditionalInput):
 
     def __init__(self,  tau_ca2, i_ca2, i_alpha):
         r"""
-        :param float tau_ca2: :math:`\tau_{\mathrm{Ca}^{+2}}`
-        :param float i_ca2: :math:`I_{\mathrm{Ca}^{+2}}`
-        :param float i_alpha: :math:`I_{\alpha}`
+        :param tau_ca2: :math:`\tau_{\mathrm{Ca}^{+2}}`
+        :type tau_ca2: Float, iterable of Floats, RandomDistribution or \
+                       function
+        :param i_ca2: :math:`I_{\mathrm{Ca}^{+2}}`
+        :type i_ca2: Float, iterable of Floats, RandomDistribution or \
+                     function
+        :param i_alpha: :math:`I_{\alpha}`
+        :type i_alpha: Float, iterable of Floats, RandomDistribution or \
+                       function
         """
         super(AdditionalInputCa2Adaptive, self).__init__([
             DataType.S1615,   # e^(-ts / tau_ca2)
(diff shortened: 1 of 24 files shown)


spynnaker/pyNN/models/neuron/additional_inputs/additional_input_ca2_adaptive.py
```diff
@@ -37,9 +37,15 @@ class AdditionalInputCa2Adaptive(AbstractAdditionalInput):
 
     def __init__(self,  tau_ca2, i_ca2, i_alpha):
         r"""
-        :param float tau_ca2: :math:`\tau_{\mathrm{Ca}^{+2}}`
-        :param float i_ca2: :math:`I_{\mathrm{Ca}^{+2}}`
-        :param float i_alpha: :math:`I_{\alpha}`
+        :param tau_ca2: :math:`\tau_{\mathrm{Ca}^{+2}}`
+        :type tau_ca2: Float, iterable of Floats, RandomDistribution or \
```
andrewgait, line 41: I think our standard way of writing something like this would be something like "float, iterable(float), ~pyNN.random.RandomDistribution or (mapping) function" given how other documentation currently looks.  (I think we'd put pyNN.random.RandomDistribution because our RandomDistribution class is a simple wrapper to the PyNN class).

spynnaker/pyNN/models/neuron/synapse_types/synapse_type_alpha.py
```diff
@@ -53,12 +53,24 @@ class SynapseTypeAlpha(AbstractSynapseType):
...
+                         function
+        :param inh_response: :math:`response^\mathrm{linear}_i`
+        :type inh_response: Float, iterable of Floats, RandomDistribution or \
+                            function
+        :param inh_exp_response: :math:`response^\mathrm{exponential}_i`
+        :type inh_exp_response: Float, iterable of Floats, \
+                                RandomDistribution or function
+        :param tau_syn_I: :math:`\tau^{syn}_i`
+        :type tau_syn_I: Float, iterable of Floats, RandomDistribution or \
+                         function
```
andrewgait, line 73: Is there a reason why you've added the params/types like this for the components (i.e. synapse_type, input_type, threshold_type), but not for the neuron models?  I think it's just neuron_model_izh.py and neuron_model_leaky_integrate_and_fire.py where you've not done it...

andrewgait (changes requested):
One suggestion to make this more consistent with other docs, then I think you've missed changing the docs on the neuron_model components... ?

spynnaker/pyNN/models/neuron/additional_inputs/additional_input_ca2_adaptive.py
alan-stokes: fixed

spynnaker/pyNN/models/neuron/synapse_types/synapse_type_alpha.py
alan-stokes: ermm........... no idea what ya on about ;-) definitely wasn't that i missed those completely until you pointed them out to me and fixed. definitely not that. fixed now.

andrewgait (approved):

PR #973, opened 2021-01-25
Remove Python 2.7 support

dkfellows:
PyPI and pip are forcing the issue; we can't keep going much longer.
SpiNNakerManchester/SpiNNUtils#132

Note that the code is not yet adapted to not use six and other compatibility hacks due to SpiNNakerManchester/sPyNNaker#958; a further PR will be developed to address that later.


spynnaker/pyNN/models/populations/idmixin.py
```diff
@@ -72,38 +68,36 @@ def record(self, variables, to_file=None, sampling_interval=None):
...
                 selector=self.__id, parameter_names=name)[0]
-        except Exception:
-            ei = sys.exc_info()
+        except Exception as e:
             try:
                 # try initialisable variable
                 return self.__population._get_initial_value(
                     selector=self.__id, variable=name)[0]
-            except Exception:
+            except Exception:  # pylint: disable=broad-except
```
alan-stokes, line 81: why dont we acutally get he right exceptions caught instead?

alan-stokes (approved):

PR #978, opened 2021-02-03
Update defaults

rowleya:
Updates the default algorithms to use:
 - PairOnChipRoutingCompression (loading algorithm)
 - SpreaderPlacer (mapping)
 - NerRouteTrafficAware (mapping)
 - ZonedRoutingInfoAllocator (mapping)

These are what we believe to be our "best" algorithms in general, and seem to help a bit with the cortical microcircuit model.

This also modifies the Munich motor device to provide n keys rather than use a fixed mask constraint, as this works with the ZonedRoutingInfoAllocator and has the same effect.

Tested by:
http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/update_defaults/11/pipeline/


Christian-B (commented):
Merge when integration tests green!

PR #983, opened 2021-02-04
file name fixing for merge

Christian-B:


spynnaker8/models/connectors/all_to_all_connector.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2017-2019 The University of Manchester
+# Copyright (c) 2017-202119 The University of Manchester
```
alan-stokes, line 1: now thats a doozy of a year

alan-stokes:
once the year is fixed. fine

PR #984, opened 2021-02-04
Initialize fix

Christian-B:
fixes https://github.com/SpiNNakerManchester/sPyNNaker8/issues/535

Non pynn mthods that where only on a population so views and idmixin could call then have been changed to protected. (underscore)
With a stub left behind to explain the change.

pop/view/idmixin initialize now.
1. Immediately changes the values returned by initial_values property
2. works after run (before reset) but warns that only valid until reset

reset() now 
Immediately changes the values returned by initial_values property
including any initialize before run
Excluding any after run

idmixin now supports initialize and initial_values while keeping the get_initial_value and set_initial_value the PyNN's version had.

See p8_integration_tests/quick_test/test_change_neuron_parameters_between_runs/test_initialize.py for examples


spynnaker/pyNN/models/populations/population.py
```diff
@@ -569,8 +569,10 @@ def initial_values(self):
                 "Population does not support the initialisation")
         return self._vertex.initial_values
 
-    # NON-PYNN API CALL
     def get_initial_value(self, variable, selector=None):
+        raise NotImplementedError("use initial_values property")
```
alan-stokes, line 573: shouldnt we add a depreciated thingy here as well, so next release we can kill it fully

spynnaker/pyNN/models/populations/population.py
```diff
@@ -587,30 +589,10 @@ def get_initial_value(self, variable, selector=None):
...
-        if not self.__vertex_population_initializable:
-            raise KeyError(
-                "Population does not support the initialisation of {}".format(
-                    variable))
-        if get_not_running_simulator().has_ran \
-                and not self.__vertex_changeable_after_run:
-            raise Exception("Population does not support changes after run")
-        self._read_parameters_before_set()
-        self._vertex.set_initial_value(variable, value, selector)
+        raise NotImplementedError("use initialize method")
```
alan-stokes, line 593: same

spynnaker/pyNN/models/populations/idmixin.py
```diff
@@ -172,15 +176,21 @@ def get_initial_value(self, variable):
...
     def set_initial_value(self, variable, value):
         """ Set the initial value of a state variable of the cell.
-
         :param str variable: The name of the variable
         :param float value: The value of the variable
         """
-        self.__population.set_initial_value(variable, value, self.__id)
+        self.__population._initialize(variable, value, self.__id)
+
+    def initialize(self, **initial_values):
```
alan-stokes, line 188: thank you.

spynnaker/pyNN/models/populations/population.py
Christian-B: Yes could do.
What format do you suggest the takes.

p8_integration_tests/quick_test/test_change_neuron_parameters_between_runs/test_initialize.py
```diff
@@ -0,0 +1,92 @@
...
+        # v on not changed is now the current state not initital value
+        self.assertNotEqual(-60, pop.initial_values["v"][0])
+        self.assertNotEqual(-62, pop.initial_values["v"][0])
+        self.assertEquals(-60, pop.initial_values["v"][1])
+        self.assertNotEqual(-60, pop.initial_values["v"][2])
+        self.assertNotEqual(-62, pop.initial_values["v"][2])
+        p.run(runtime3)
+
+        p.reset()
+        neo = pop.get_data('v')
```
alan-stokes, line 57: this neo inst used. is there a reason for it being here? line 73 overrides it afterall

alan-stokes: if its used to set up initial state, shouldnt it be _ = pop_get_data('v')

Christian-B: this neo inst used. = Agreed removed.

shouldnt it be _ : NO! get_data is the Standard PyNN call to get the value (of in this case v) for each timestep not just the initial state.

PR #989, opened 2021-02-08
Put doc builds in parallel with other validation

dkfellows:
Documentation building is taking a long time (due to details of how `sphinx-apidoc` works). Split it into its own workflow so that we more clearly get the flake8/pylint results back in the GitHub UI.

### DIFF
--- a/.github/workflows/python_actions.yml
+++ b/.github/workflows/python_actions.yml
@@ -122,6 +122,41 @@ jobs:
       uses: ./support/actions/validate-xml
       with:
         base-path: spynnaker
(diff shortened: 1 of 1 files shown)


dkfellows:
It's a shame we can't make the Sphinx build go faster, but it's an actually tricky problem due to the limitations of Python's `fnmatch` (which Sphinx apidoc is committed to using). Their standard workaround is to tell you to only run apidoc when the structure of the project changes, but that's one more thing for us to forget to do so I'm loth to consider it…

PR #990, opened 2021-02-08
dont mix nparray and empty list

Christian-B:
fixes the broken 
p8_integration_tests/quick_test/test_spike_source/test_spike_source_array.py

Also removed a broken test that was better tested with:
p8_integration_tests/quick_test/test_change_neuron_parameters_between_runs/test_initialize.py

Also stopped population from reading back the state variables if reset had been called.

Lastly an emergency fudge until https://github.com/SpiNNakerManchester/sPyNNaker/pull/991 is solved

tested by https://github.com/SpiNNakerManchester/IntegrationTests/pull/12


spynnaker/pyNN/models/spike_source/spike_source_array_vertex.py
```diff
@@ -40,10 +40,7 @@ def _send_buffer_times(spike_times, time_step):
     if len(spike_times) and hasattr(spike_times[0], "__len__"):
         data = []
```
rowleya, line 41: Note this can be made an empty numpy array too if that helps

Christian-B: NO because in this case data is a list of nparrays.

The issue was the mix between nparrays and empty lists confusing numpy

p8_integration_tests/quick_test/test_spike_source/test_spike_source_array.py
```diff
@@ -132,7 +132,7 @@ def test_recording_numerous_element_with_zero(self):
     def recording_with_empty_lists(self):
         p.setup(timestep=1.0)
         p.set_number_of_neurons_per_core(p.SpikeSourceArray, 2)
-        spike_times = [[1], [], [], [], [4], [3]]
+        spike_times = [[], [1], [], [], [4], [3]]
```
alan-stokes, line 135: why did you do this change? is there something special about the first array?

Christian-B: yes. The check if spike_times are a single list or multiple looks at the first array. So wanted to make sure it worked even if first array is empty.

spynnaker/pyNN/models/spike_source/spike_source_array_vertex.py
rowleya: Oh yes, sorry!

p8_integration_tests/quick_test/test_spike_source/test_spike_source_array.py
alan-stokes: do we have a second test that does the initial version, in case we break it the other way?

rowleya: It would be potentially easy to do that as part of this test too i.e. have one one way and one the other.

PR #1003, opened 2021-02-16
Restore doing the (very slow!) doc builds in parallel with testing

dkfellows:
This PR halves the effective build time relative to its parent branch. This is why the jobs were parallel on `master` in the first place.

Timings (with 3.8):

| Phase | Timing (s) | Notes |
| - | - | - |
| Installation | 100 (~70 just for doc) | Shorter for doc-only because we can omit installing some things |
| Testing | 179 | Quicker without coverage, varies a lot between Python versions |
| Validation | 93 | Mostly (75%) due to pylint |
| Documentation | 350 | It's the running of `sphinx-apidoc` that's costly |

We can put the Documentation build in parallel (i.e., as a separate build job) to the Testing and Validation, and it will reduce our total elapsed wall-clock build times by around 4&half; minutes (estimated 179+93=272 seconds). Once I figure out how to make Sphinx go faster, we can stop this, but that's tricky (it's dependent on how we process the code to determine what exactly goes in the docs) so we should do this for now.

### DIFF
--- a/.github/workflows/python_actions.yml
+++ b/.github/workflows/python_actions.yml
@@ -13,8 +13,12 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-# This workflow will install Python dependencies, run tests, lint and rat with a variety of Python versions
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+# This workflow will install Python dependencies, run tests, lint and rat with
+# a variety of Python versions. For more information see:
+# https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
+
+# Note that documentation building is done in parallel (with testing) because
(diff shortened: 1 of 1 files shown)


Christian-B:
While I understand the idea and you beat me to testing it the results say that doing it in parallel is actually slower.
Have a look on the actions page.

Restore doing the (very slow!) doc builds in parallel with testing    7m 21s
added 3.9 and rat 3.8 (more like other repr)  5m 59s

Christian-B (changes requested):
I liked the idea.
To bad that github actions does not.

These tests actually take longer to complete,

dkfellows:
You weren't running the documentation build at all. The action internally checks for the version of Python being used, and was only doing the doc build for 3.6 (it's the `blessed-version` parameter to the action; I updated that to `3.8` this morning).

dkfellows:
Now that that's turned on, I reran the combined build; the time taken there for doc building was 9m 12s (552s).

PR #1005, opened 2021-02-18
Lost spike issues

rowleya:
Makes some changes to the execution of the code, which should ensure more guaranteed operation in practice.  This is done by having the timer tick at priority 0 in the API; this means that it is at the same level as the user interrupt and the DMA interrupt, but the order that the API puts these things into the VIC means that the timer tick will actually interrupt these things.  So the order is now MC (-1), timer (0), DMA (0), user (0), SDP (1), Background (1).  

The "background" task is then a new thing which works as follows.  When the timer goes off, it turns off interrupts (so stops new MC packet reception) and then does some critical tasks like clearing the spike input buffer and stopping DMAs in progress (one thing that it can't do is stop the processing of a DMA that has finished before this is called, though that will at least be interrupted here).  It then checks whether this is the end of processing; if not, it then puts a background task on the queue to do neuron processing.  This task goes into the Spin1API queues, and so can be interrupted by the DMA and user interrupts as usual.

Note that the delay extension code has been refactored to do a similar thing.  In the current code, the delay extension code actually processes spikes with interrupts off.  This was considered dangerous and likely to result in lost packets.  This was done due to a lack of understanding of which queues will be examined during any timer tick.  After more analysis, the correct queues to clear have now been identified and the code has been fixed overall.

Note that due to changes in how TDMA is handled as well, this depends on SpiNNakerManchester/SpiNNFrontEndCommon#756.


neural_modelling/src/delay_extension/delay_extension.c
```diff
@@ -393,21 +437,106 @@ static inline void spike_process(void) {
...
+static void background_callback(uint local_time, UNUSED uint timer_count) {
+    // reset the tdma for this next cycle.
+    tdma_processing_reset_phase();
+
+    // Loop through delay stages
+    for (uint32_t d = 0; d < num_delay_stages; d++) {
+        // If any neurons emit spikes after this delay stage
+        bit_field_t delay_stage_config = neuron_delay_stage_config[d];
+        if (nonempty_bit_field(delay_stage_config, neuron_bit_field_words)) {
+            // Get key mask for this delay stage and it's time slot
```
andrewgait, line 460: "its" rather than "it's"

neural_modelling/src/delay_extension/delay_extension.c
```diff
@@ -393,21 +437,106 @@ static inline void spike_process(void) {
...
+    spike_processing = false;
     spin1_mode_restore(state);
 }
 
+static void user_callback(UNUSED uint unused0, UNUSED uint unused1) {
+    spike_process();
+}
+
+static void background_callback(uint local_time, UNUSED uint timer_count) {
+    // reset the tdma for this next cycle.
```
andrewgait, line 452: Do we want to capitalise TDMA in comments?  Or have something like tDMA instead?

neural_modelling/src/delay_extension/delay_extension.c
```diff
@@ -393,21 +437,106 @@ static inline void spike_process(void) {
...
+
+                    // If this neuron emits a spike after this stage
+                    if (bit_field_test(delay_stage_config, n)) {
+
+                        // Calculate key all spikes coming from this neuron will be
+                        // sent with
+                        uint32_t neuron_index = ((d * num_neurons) + n);
+                        uint32_t spike_key = neuron_index + key;
+
+                        if (delay_stage_spike_counters[n] > 0) {
```
andrewgait, line 482: Is it better to put this inside a #if LOG_LEVEL >= LOG_DEBUG rather than having an if (...) that will always be checked but then only to print something using log_debug(..) ?

neural_modelling/src/neuron/c_main.c
```diff
@@ -83,11 +83,15 @@ struct neuron_provenance {
...
+    //! Maximum backgrounds queued
+    uint32_t max_backgrounds_queued;
+    //! Background queue overloads
+    uint32_t n_background_queue_overloads;
 };
 
 //! values for the priority for each callback
 typedef enum callback_priorities {
-    MC = -1, DMA = 0, USER = 0, SDP = 1, TIMER = 2
+    MC = -1, DMA = 0, USER = 0, SDP = 1, TIMER = 0, BACKGROUND = 1
```
andrewgait, line 94: Is there a reason why the order is different here compared to the delay extension code?  (TIMER is before USER in the enum there...)

p8_integration_tests/quick_test/test_auto_pause_and_resume_tests/test_long_run/spynnaker.cfg
```diff
@@ -5,22 +5,6 @@ max_sdram_allowed_per_chip = 1529836
 [Buffers]
 use_auto_pause_and_resume = True
 # Lower than defaults
-# This cause less partitioning and therefor more auto pause runs
+# This cause less partitioning and therefore more auto pause runs
```
andrewgait, line 8: "This causes", and probably "auto-pause and resume" instead of just "auto pause".

andrewgait (changes requested):
Looks good to me; a good idea to split into timer_callback and background_callback.  Just a few minor things that need looking at.

neural_modelling/src/delay_extension/delay_extension.c
rowleya: Fixed

neural_modelling/src/delay_extension/delay_extension.c
rowleya: I have made them all TDMA in this file

neural_modelling/src/delay_extension/delay_extension.c
rowleya: Yes, definitely - done

neural_modelling/src/neuron/c_main.c
rowleya: The order isn't critical but I have changed it to be more consistent anyway

p8_integration_tests/quick_test/test_auto_pause_and_resume_tests/test_long_run/spynnaker.cfg
rowleya: Fixed

andrewgait (approved):

PR #1008, opened 2021-02-19
cap numpy

Christian-B:
This is an alternative to:
https://github.com/SpiNNakerManchester/SupportScripts/pull/27

However if this approach is used the same fix has to be applied to 
https://github.com/SpiNNakerManchester/SpiNNGym


Christian-B:
Cleaner solution implemented by Donal

PR #1009, opened 2021-02-19
Test of matplotlib installation action

dkfellows:
Test of SpiNNakerManchester/SupportScripts#27


dkfellows:
OK it works, but I've put it in the wrong place in the workflow here so this PR shouldn't be merged.

PR #1010, opened 2021-02-19
Much faster documentation building

dkfellows:
The problem with the slow documentation building was because the apidoc preparation code was finding lots of irrelevant files (notably in the `.git` directory!) that was hugely extending the list of things to filter out unnecessarily.

Once some other minor issues are sorted, this will enable merging the doc building into the main workflows.


dkfellows:
This cut the build time for the sphinx phase (measured in GHA) to 47 seconds.

dkfellows:
This now requires turning the requirement for `doc-check` off in order to commit.

PR #1011, opened 2021-02-19
We need to use today's new special matplotlib trickery

dkfellows:
Because “`Beginning with Matplotlib 3.4, Python 3.7 or above is required.`” and it is a transitive dependency of pynn, which doesn't version its requirement there.


andrewgait (approved):
Looks good to me... is this still a draft for a reason?

dkfellows:
At the time I started to make it, I thought I still had another bug fix to apply. But I did that one in the interim before actually pressing the create button. 😁

PR #1013, opened 2021-02-22
cap matplotlib and remove duplicate quantites

Christian-B:
replaces previous Pr that tried this.

Matplotlib not added to setup.py as that has caused more issues than it fixed in the past.


andrewgait (approved):
Looks fine to me.

PR #1016, opened 2021-02-23
More exposed via global variables

Christian-B:
Depends on https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/762

Read configs preview before sim.setup

Make use of the extra info available through global variables.

### DIFF
--- a/p8_integration_tests/base_test_case.py
+++ b/p8_integration_tests/base_test_case.py
@@ -147,7 +147,7 @@ class BaseTestCase(unittest.TestCase):
             report_file.write(message)
 
     def get_provenance(self, _main_name, detail_name):
-        provenance_file_path = globals_variables.provenance_file_path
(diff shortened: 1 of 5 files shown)


rowleya (approved):
Agree that this makes things neater and avoids accessing _ variables!

rowleya:
Note a branch name mismatch makes this fail

Christian-B:
> Note a branch name mismatch makes this fail

yes I say that.

But as a sub Pr of a much bigger PR it does not really matter.
Just wanted to make sure these changes where see sperately

rowleya:
OK - I think I merged the GFE one for some reason without merging these.  Do you want them merged?

Christian-B:
> OK - I think I merged the GFE one for some reason without merging these. Do you want them merged?

create a FEC branch to match name so green.

Yes may as well merge as makes follow up PR cleaner. Thanks

PR #1017, opened 2021-02-23
readthedocs fixing

dkfellows:


Christian-B:
looks good

dkfellows:
The reason for all the mocks is that readthedocs doesn't install the project that it is documenting. It's a mess, but now it is at least a mess that is producing largely correct docs.

PR #1019, opened 2021-02-25
These globals are no longer used in this impl

andrewgait:
A couple of global parameters are no longer used in the external devices impl (caught by the compiler), so this PR removes them.

Tested by https://github.com/SpiNNakerManchester/IntegrationTests/pull/24


rowleya (approved):
Always happy to remove unused variables!

Issue #487, opened 2018-03-18
[Structural Plasticity] Add ability to record


pabogdan:
Currently, the only ways of observing the effect of structural plasticity are either to enable appropriate log_infos in relevant C files, recompiling and text mining through the resulting reports, or to occasionally extract connectivity information during the simulation. This is clearly not ideal.

A solution that would tie in more smoothly to PyNN would be nice. For example, ``structurally_plastic_projection.record("rewiring")`` would probably be more useful than our current approaches, not to mention faster and more accurate.

andrewgait:
I agree this is a good idea, but what kind of variables or information would you suggest to be "recorded" here?  Clearly it's somewhat prohibitive in larger networks to record the connectivity after every timestep, but are you thinking of other variables (successful creations/eliminations, for example, or maybe just a "simple" count of the number of current connections at each timestep) as well?

pabogdan:
> Clearly it's somewhat prohibitive in larger networks to record the connectivity after every timestep
Quite right. Users can already do this if they really want to, but I don't think anyone needs/wants this.
> but are you thinking of other variables (successful creations/eliminations, for example, or maybe just a "simple" count of the number of current connections at each timestep) as well?

In each timestep we could have number of incoming connection per neuron. This would implicitly contain the information about formations and removals. 

This approach would have the benefit of only requiring core-local information which I believe we already compute somewhere. 

The downside would be you need as much information to record this as for something like recording v or gsyn (info per neuron per timestep). 

Thoughts?

rowleya:
If the data is useful, it is still optional so having the ability to do it sounds fine to me.

pabogdan:
+ being able to record fan-ins for a subset of neurons would probably help here.

andrewgait:
> Plus, being able to record fan-ins for a subset of neurons would probably help here.

If we set it up to do e.g. incoming connections per neuron, then the already-existing selective recording mechanisms will probably work here too.

alan-stokes:
sounds good to me. assuming its uint32_ts and not floats. so its the basic stuff.

andrewgait:
We do compute this sort of information locally on cores, but only during spike processing.  So if a neuron processes no spikes in a timestep, then it's not immediately obvious how to obtain the information about its incoming "live" connections... unless I'm missing something when looking at the current code.

rowleya:
[...]

PR #1042, opened 2021-03-11
Structural plasticity recording

andrewgait:
Work related to https://github.com/SpiNNakerManchester/sPyNNaker/issues/487

In this PR the recording is done on the post-synaptic population of a projection, recorded at the moment in the same way that v / gsyn are recorded, i.e. per-neuron and per-timestep.  So it is accessed by doing

post_pop.record("rewiring")

and data is returned by

post_pop.get_data("rewiring")

In the C code this is recorded as an int32:

-1 if no rewiring happens at that post-neuron ID; 
otherwise
a 32-bit number with the final bit 0 (elimination) or 1 (formation) and the remaining 31 bits containing the (global) pre-neuron ID

This int32 is then intercepted during the call to get_data and converted into a neo Event object, where the times of formation and elimination events are separately recorded, and each time has an associated label which details the pre- and post-neuron ID, and 'formation' / 'elimination' as appropriate.  So a user can simply plot the event times (from segment.events[n].times, n=0 is formation, n=1 is elimination), or if they wish to get the labels for the relevant neuron IDs (from segment.events[n].labels) for further plotting / whatever then that is also possible.  (I have example code for plotting an animation for simple-ish "formation to full" or "elimination to empty" examples if there is interest in these being added to PyNN8Examples, though I know that matplotlib animation can be very OS-specific and therefore problematic, so I've left it out for now).

One more note: if the projection is not structural and the user asks for rewiring data from the post-synaptic population, it will currently (from the above rules) record arrays with -1 in every position, and this will end up with neo Event objects with nothing in them; I suggest perhaps warning the user and cancelling the recording request if this is asked for by mistake?

**Edited:** another note: I think (though having not tested I can't be certain) that this may also be limited to recording one event per timestep per core, though I'm not totally sure.  The examples I have run with so far have only had parameters such that there is at most one event per timestep.

**Further edited (April 28th 2021):** This has now been refactored to only record when an event takes place, and to treat the "rewiring" variable as different from a neuron_recordable_variable or a per_timestep_variable; I have called this new type of variable an "event_per_core_variable".  This necessitated a change to go back to a previous incarnation where the layout of the bits in the recorded value is now:

first bit: 0 (elimination) or 1 (formation)
next 8 bits: the (local) post-neuron ID
remaining 32 bits: the (global) pre-neuron ID

These are then decoded when the data has been collected (and converted to global IDs where necessary).  The neo Event objects that come back to the user are similar to before, though the newest versions of neo no longer allow an array in the labels object, so the labels are now strings that can be decoded by the user should they wish to do so.

Merge with: https://github.com/SpiNNakerManchester/sPyNNaker8NewModelTemplate/pull/72 and https://github.com/SpiNNakerManchester/SpiNNGym/pull/25 (Edit: note that after refactoring these PRs are no longer needed and can be closed)
Tested by: https://github.com/SpiNNakerManchester/IntegrationTests/pull/34


rowleya (changes requested):
Looks OK for a start, but there are a couple of things that might need to be looked at

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c
```diff
@@ -287,6 +297,18 @@ bool synapse_dynamics_initialise(
...
+        if (structural_rec_array == NULL) {
+            log_error("Unable to allocate structural recording array"
+                    "- Out of DTCM");
+            return false;
+        }
+        for (uint32_t i = 0; i < n_neurons; i++) {
+        	structural_rec_array[i].recorded_val = -1;
+        }
+    }
+
```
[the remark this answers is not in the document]
andrewgait: Yes - this is the main issue with this.  I will likely have to make this into an array in order to deal with multiple rewirings on the same timestep; it's more frequent than you might think because it's rewirings into the same post-neuron, rather than necessarily the same pair of pre- and post-neurons.

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c
```diff
@@ -467,6 +490,11 @@ bool synapse_dynamics_remove_neuron(uint32_t offset, synaptic_row_t row) {
 
     // Decrement FP
     fixed_region->num_plastic--;
+
+    // Create recorded value
+    // (bottom bit: added/removed, remainder: pre-neuron global id)
+    structural_rec_array[post_id].recorded_val = (0 << 0) | (pre_id << 1);
```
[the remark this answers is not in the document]
andrewgait: At the time it seemed like a good place, but I agree that inside the actual structural plasticity code is probably better.

spynnaker/pyNN/models/recorder.py
```diff
@@ -669,6 +692,93 @@ def __read_in_signal(
...
+                    bin_val = bin(int(events[n_atom]))[2:]  # take the "0b" off
+                    # front-pad with zeros to size 32
+                    len_bin = len(bin_val)
+                    for n in range(32-len_bin):
+                        bin_val = "0"+bin_val
+
+                    # final bit contains rewire status; remaining is preid
+                    rewire_status = int(bin_val[-1], 2)
+                    preid = int("0b"+(bin_val[:31]), 2)
+                    postid = n_atom
```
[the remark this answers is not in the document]
andrewgait: I will look into doing it with masking/shifting to make it clearer (and probably shorter).

pabogdan (approved):
I'm happy with these additions and think they would be very useful for someone working with structural plasticity in the future. My only concern is that a user may not know this feature exists so it would need to show up in documentation nicely.

neural_modelling/src/neuron/c_main.c
```diff
@@ -325,6 +325,21 @@ void background_callback(uint timer_count, uint local_time) {
         count_rewire_attempts++;
     }
 
+    // Does this happen here or can it be combined with the above?
```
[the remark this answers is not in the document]
andrewgait: Done.

neural_modelling/src/neuron/c_main.c
```diff
@@ -325,6 +325,21 @@ void background_callback(uint timer_count, uint local_time) {
...
+//      synapse_dynamics_changes(n_neurons, neurons_changed);
+        synapse_dynamics_recording_values(n_neurons, rec_values);
+        for (uint32_t i = 0; i < n_neurons; i++) {
+//          io_printf(IO_BUF, "time %u ", time);
+//          if (neurons_changed[i] == 1) {
+            neuron_record_structural(i, rec_values[i]);
+//          }
+        }
+    }
+
```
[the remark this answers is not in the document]
andrewgait: Done.

neural_modelling/src/neuron/neuron_recording.h
```diff
@@ -138,6 +138,13 @@ static inline void neuron_recording_record_int32(
     data[index] = value;
 }
 
+//static inline void neuron_recording_record_uint32(
+//        uint32_t var_index, uint32_t neuron_index, uint32_t value) {
+//    uint8_t index = neuron_recording_indexes[var_index][neuron_index];
+//    uint32_t *data = (uint32_t *) recording_values[var_index];
+//    data[index] = value;
+//}
+
```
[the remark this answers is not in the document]
andrewgait: Done.

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c
```diff
@@ -63,6 +63,16 @@ uint32_t num_plastic_pre_synaptic_events = 0;
...
+//! Structural rewiring recording values
+typedef struct structural_rec_t {
+    int32_t recorded_val;
+} structural_rec_t;
+
+typedef struct structural_rec_t* structural_rec_pointer_t;
+
+//! Array of things for structural recording
+static structural_rec_pointer_t structural_rec_array;
+
```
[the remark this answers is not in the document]
andrewgait: I've moved this to a different location now but not changed it as of yet.  Not sure it's totally necessary for it to remain visible if it's indicated by the name though...

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c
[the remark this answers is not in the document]
andrewgait: This has moved into topographic_map_impl.c but I have not updated it yet to deal with multiple rewirings on the same timestep - that's the next step.

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c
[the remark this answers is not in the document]
andrewgait: Moved to topographic_map_impl.c

neural_modelling/src/neuron/plasticity/synapse_dynamics_static_impl.c
```diff
@@ -37,9 +37,33 @@ static uint32_t synapse_type_bits;
 //! Mask to extract the synapse type (has ::synapse_type_bits bits set)
 static uint32_t synapse_type_mask;
 
+//! Structural rewiring recording values
+typedef struct structural_rec_t {
+    int32_t recorded_val;
+//    uint32_t changed;
+} structural_rec_t;
```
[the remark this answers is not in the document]
andrewgait: Moved to topographic_map_impl.c

neural_modelling/src/neuron/structural_plasticity/synaptogenesis/sp_structs.h
```diff
@@ -190,7 +190,9 @@ static inline bool sp_structs_get_sub_pop_info(
 //! \return True if the synapse was removed
 static inline bool sp_structs_remove_synapse(
         current_state_t *restrict current_state, synaptic_row_t restrict row) {
-    if (!synapse_dynamics_remove_neuron(current_state->offset, row)) {
+    if (!synapse_dynamics_remove_neuron(
+    		current_state->key_atom_info->lo_atom + current_state->pre_syn_id,
+    		current_state->post_syn_id, current_state->offset, row)) {
```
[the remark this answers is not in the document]
andrewgait: Moved to topographic_map_impl.c

neural_modelling/src/neuron/structural_plasticity/synaptogenesis/sp_structs.h
```diff
@@ -212,8 +214,9 @@ static inline bool sp_structs_add_synapse(
             current_state->pre_population_info->delay_lo;
 
     if (!synapse_dynamics_add_neuron(
+            current_state->key_atom_info->lo_atom + current_state->pre_syn_id,
             current_state->post_syn_id, row, appr_scaled_weight, actual_delay,
-            current_state->pre_population_info->connection_type)) {
+			current_state->pre_population_info->connection_type)) {
```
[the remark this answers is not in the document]
andrewgait: Moved to topographic_map_impl.c

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c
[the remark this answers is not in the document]
andrewgait: I think I now have this working for multiple rewirings on a single timestep.

spynnaker/pyNN/models/recorder.py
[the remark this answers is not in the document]
andrewgait: Improved the code to use shifting and masking (it happens in a slightly different location now).

andrewgait:
This is ready to review again now (current test error is related to a rat copyight issue...).

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c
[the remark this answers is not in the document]
andrewgait: Recording has moved into the structural code, so this is outdated.

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c
```diff
@@ -503,10 +531,23 @@ bool synapse_dynamics_add_neuron(uint32_t id, synaptic_row_t row,
 
     // Increment FP
     fixed_region->num_plastic++;
+
+    // Create recorded value
+    // (bottom bit: added/removed, remainder: pre-neuron global id)
+    structural_rec_array[id].recorded_val = (1 << 0) | (pre_id << 1);
```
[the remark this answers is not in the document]
andrewgait: Recording has moved into the structural code, so this is outdated.

neural_modelling/src/neuron/plasticity/stdp/synapse_dynamics_stdp_mad_impl.c
```diff
@@ -503,10 +531,23 @@ bool synapse_dynamics_add_neuron(uint32_t id, synaptic_row_t row,
...
         synapse_row_fixed_part_t *fixed) {
     return synapse_row_num_plastic_controls(fixed);
 }
+
+void synapse_dynamics_recording_values(uint32_t n_neurons, int32_t* rec_values) {
+	for (uint32_t i=0; i < n_neurons; i++) {
+		rec_values[i] = structural_rec_array[i].recorded_val;
+		// reset value
+		structural_rec_array[i].recorded_val = -1;
+	}
```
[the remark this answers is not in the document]
andrewgait: Recording has moved into the structural code, so this is outdated.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -733,49 +830,56 @@ def set_recording(self, variable, new_state, sampling_interval=None,
             raise ConfigurationException("Variable {} is not supported".format(
                 variable))
 
-    def _get_buffered_sdram(self, vertex_slice, n_machine_time_steps):
+    def _get_buffered_sdram(
+            self, vertex_slice, n_machine_time_steps, max_rewires_per_ts):
```
rowleya, line 834: I would make max_rewires_per_ts an object variable passed in during ```__init__```.  Then additionally __don't__ include it in the list of variables, instead making it active only if the value is > 0.  This makes it similar to other things.  Because this is the first example of a variable of this type (where the number of things of this type that get recorded per timestep is variable), it doesn't make sense to generalise beyond this at this point...

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -733,49 +830,56 @@ def set_recording(self, variable, new_state, sampling_interval=None,
...
+        :param int max_rewires_per_ts: the max possible rewires per timestep
         :rtype: list(int)
         """
         values = list()
         for variable in itertools.chain(
                 self.__sampling_rates, self.__per_timestep_variables):
             values.append(self.get_buffered_sdram(
-                variable, vertex_slice, n_machine_time_steps))
+                variable, vertex_slice, n_machine_time_steps,
+                max_rewires_per_ts))
```
rowleya, line 846: If max_rewires_per_ts is passed in during ```__init__```, you can then factor it out of this loop and just do a direct calculation for max_rewires_per_ts.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -733,49 +830,56 @@ def set_recording(self, variable, new_state, sampling_interval=None,
...
         return values
 
     def write_neuron_recording_region(
             self, spec, neuron_recording_region, vertex_slice,
-            data_n_time_steps):
+            data_n_time_steps, max_rewires_per_ts):
```
rowleya, line 851: Again, if this is passed in to the constructor it saves odd passing around here...

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -733,49 +830,56 @@ def set_recording(self, variable, new_state, sampling_interval=None,
...
             the vertex slice
         :param int data_n_time_steps: how many time steps to run this time
+        :param int max_rewires_per_ts: the max possible rewires per timestep
         :rtype: None
         """
         spec.switch_write_focus(neuron_recording_region)
         spec.write_array(get_recording_header_array(
-            self._get_buffered_sdram(vertex_slice, data_n_time_steps)))
+            self._get_buffered_sdram(
+                vertex_slice, data_n_time_steps, max_rewires_per_ts)))
```
rowleya, line 865: Avoid passing this here by making instance variable (see elsewhere).

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -733,49 +830,56 @@ def set_recording(self, variable, new_state, sampling_interval=None,
...
 
         # Write the number of variables and bitfields (ignore per-timestep)
         n_vars = len(self.__sampling_rates) - len(self.__bitfield_variables)
         spec.write_value(data=n_vars)
         spec.write_value(data=len(self.__bitfield_variables))
 
         # Write the recording data
-        recording_data = self._get_data(vertex_slice)
+        recording_data = self._get_data(vertex_slice, max_rewires_per_ts)
```
rowleya, line 873: Avoid passing this here by making instance variable (see elsewhere).

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -733,49 +830,56 @@ def set_recording(self, variable, new_state, sampling_interval=None,
...
         spec.write_array(recording_data)
 
-    def get_buffered_sdram_per_record(self, variable, vertex_slice):
+    def get_buffered_sdram_per_record(
+            self, variable, vertex_slice, max_rewires_per_ts):
```
rowleya, line 877: I think this will end up not being called here if it isn't treated like a normal "variable".

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -135,16 +138,21 @@ def __init__(
                 per_timestep_variables, start=region_id + 1):
             self.__region_ids[variable] = ts_region_id
 
-    def _count_recording_per_slice(self, variable, vertex_slice):
+    def _count_recording_per_slice(
+            self, variable, vertex_slice, max_rewires_per_ts):
```
rowleya, line 142: I think it makes more sense to do the calculation for this outside of this function.  If this is not passed in as a "variable" in the constructor but instead stored as max_rewires_per_ts in an instance variable, it will then not end up in a loop that calls this function, and can be dealt with separately.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -135,16 +138,21 @@ def __init__(
...
         :rtype: int
         """
         if self.__sampling_rates[variable] == 0:
             return 0
         if self.__indexes[variable] is None:
-            return vertex_slice.n_atoms
+            if variable == self.REWIRING:
+                return max_rewires_per_ts
+            else:
+                return vertex_slice.n_atoms
```
rowleya, line 155: Deal with separately - see above.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -784,7 +888,8 @@ def get_buffered_sdram_per_record(self, variable, vertex_slice):
                 return 0
             size = self.__per_timestep_datatypes[variable].size
             return self._N_BYTES_FOR_TIMESTAMP + size
-        n_neurons = self._count_recording_per_slice(variable, vertex_slice)
+        n_neurons = self._count_recording_per_slice(
+            variable, vertex_slice, max_rewires_per_ts)
```
rowleya, line 892: I think this will not end up being needed here if not being treated as a normal variable

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -816,13 +923,14 @@ def get_buffered_sdram_per_timestep(self, variable, vertex_slice):
             if rate == 0:
                 return 0
 
-        data_size = self.get_buffered_sdram_per_record(variable, vertex_slice)
+        data_size = self.get_buffered_sdram_per_record(
+            variable, vertex_slice, max_rewires_per_ts)
```
rowleya, line 927: I think this will not end up being needed here.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -816,13 +923,14 @@ def get_buffered_sdram_per_timestep(self, variable, vertex_slice):
...
         if rate == 1:
             return data_size
         else:
             return data_size // rate
 
-    def get_sampling_overflow_sdram(self, vertex_slice):
+    def get_sampling_overflow_sdram(self, vertex_slice, max_rewires_per_ts):
```
rowleya, line 933: I think this will not end up being needed here.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -845,11 +954,13 @@ def get_sampling_overflow_sdram(self, vertex_slice):
...
                 data_size = self.get_buffered_sdram_per_record(
-                    variable,  vertex_slice)
+                    variable, vertex_slice, max_rewires_per_ts)
                 overflow += data_size // rate * (rate - 1)
         return overflow
 
-    def get_buffered_sdram(self, variable, vertex_slice, n_machine_time_steps):
+    def get_buffered_sdram(
+            self, variable, vertex_slice, n_machine_time_steps,
+            max_rewires_per_ts):
```
rowleya, line 963: If you can make this an instance variable (see elsewhere) it won't need passing in here.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -858,18 +969,21 @@ def get_buffered_sdram(self, variable, vertex_slice, n_machine_time_steps):
...
             how many machine time steps to run for
+        :param int max_rewires_per_ts: the max possible rewires per timestep
         :return: data size
         :rtype: int
         """
         # Per timestep variables can't be done at a specific rate
         if variable in self.__per_timestep_variables:
-            item = self.get_buffered_sdram_per_record(variable, vertex_slice)
+            item = self.get_buffered_sdram_per_record(
+                variable, vertex_slice, max_rewires_per_ts)
```
rowleya, line 979: If max_rewires_per_ts is not part of the "variables" here, it won't end up being called here, and can instead be treated separately below this loop.  This will open up for future usage should another variable like this end up being needed.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -858,18 +969,21 @@ def get_buffered_sdram(self, variable, vertex_slice, n_machine_time_steps):
...
             return item * n_machine_time_steps
 
         rate = self.__sampling_rates[variable]
         if rate == 0:
             return 0
-        data_size = self.get_buffered_sdram_per_record(variable, vertex_slice)
+        data_size = self.get_buffered_sdram_per_record(
+            variable, vertex_slice, max_rewires_per_ts)
```
rowleya, line 986: Will be dropped

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -946,9 +1060,10 @@ def get_static_sdram_usage(self, vertex_slice):
             self.get_sdram_usage_in_bytes(vertex_slice))
         return int(sdram)
 
-    def get_variable_sdram_usage(self, vertex_slice):
+    def get_variable_sdram_usage(self, vertex_slice, max_rewires_per_ts):
```
rowleya, line 1063: Avoided by passing into constructor

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -959,7 +1074,7 @@ def get_variable_sdram_usage(self, vertex_slice):
             if rate > 0:
                 fixed_sdram += self._SARK_BLOCK_SIZE
                 per_record = self.get_buffered_sdram_per_record(
-                    variable, vertex_slice)
+                    variable, vertex_slice, max_rewires_per_ts)
```
rowleya, line 1077: Avoided by not having the variable in this loop

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -971,7 +1086,7 @@ def get_variable_sdram_usage(self, vertex_slice):
                     fixed_sdram += (per_record - average_per_timestep)
         for variable in self.__per_timestep_recording:
             per_timestep_sdram += self.get_buffered_sdram_per_record(
-                variable, vertex_slice)
+                variable, vertex_slice, max_rewires_per_ts)
```
rowleya, line 1089: Avoided by not having the variable in this loop - instead can be calculated separately below here

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -1045,9 +1160,10 @@ def __add_indices(self, data, variable, rate, n_recording, vertex_slice):
             data.append(
                 numpy.array(local_indexes, dtype="uint8").view("uint32"))
 
-    def _get_data(self, vertex_slice):
+    def _get_data(self, vertex_slice, max_rewires_per_ts):
```
rowleya, line 1163: Avoid by having this as in instance variable

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -1058,7 +1174,7 @@ def _get_data(self, vertex_slice):
                 continue
             rate = self.__sampling_rates[variable]
             n_recording = self._count_recording_per_slice(
-                variable, vertex_slice)
+                variable, vertex_slice, max_rewires_per_ts)
```
rowleya, line 1177: Will disappear...

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -1067,7 +1183,7 @@ def _get_data(self, vertex_slice):
         for variable in self.__bitfield_variables:
             rate = self.__sampling_rates[variable]
             n_recording = self._count_recording_per_slice(
-                variable, vertex_slice)
+                variable, vertex_slice, max_rewires_per_ts)
```
rowleya, line 1186: Avoided by removing from this loop

spynnaker/pyNN/models/neuron/implementations/neuron_impl_standard.py
```diff
@@ -45,18 +45,20 @@ class NeuronImplStandard(AbstractNeuronImpl):
         "__n_steps_per_timestep"
     ]
 
-    _RECORDABLES = ["v", "gsyn_exc", "gsyn_inh"]
+    _RECORDABLES = ["v", "gsyn_exc", "gsyn_inh", "rewiring"]
```
rowleya, line 48: This can be removed from here and instead treated more like packets per ts or spikes

spynnaker/pyNN/models/neuron/implementations/neuron_impl_standard.py
```diff
@@ -45,18 +45,20 @@ class NeuronImplStandard(AbstractNeuronImpl):
...
 
     _RECORDABLE_DATA_TYPES = {
         "v": DataType.S1615,
         "gsyn_exc": DataType.S1615,
-        "gsyn_inh": DataType.S1615
+        "gsyn_inh": DataType.S1615,
+        "rewiring": DataType.INT32
```
rowleya, line 54: Remove from here

spynnaker/pyNN/models/neuron/implementations/neuron_impl_standard.py
```diff
@@ -45,18 +45,20 @@ class NeuronImplStandard(AbstractNeuronImpl):
...
     }
 
     _RECORDABLE_UNITS = {
         'v': 'mV',
         'gsyn_exc': "uS",
-        'gsyn_inh': "uS"}
+        'gsyn_inh': "uS",
+        'rewiring': ""}
```
rowleya, line 61: Remove from here

spynnaker/pyNN/models/neuron/population_machine_vertex.py
```diff
@@ -467,10 +467,15 @@ def generate_data_specification(
         spec.write_array(simulation_utilities.get_simulation_header_array(
             self.__binary_file_name, machine_time_step, time_scale_factor))
 
+        max_rewires_per_ts = 1
```
rowleya, line 470: This may be 0 if no structural dynamics I think...

spynnaker/pyNN/models/neuron/population_machine_vertex.py
```diff
@@ -467,10 +467,15 @@ def generate_data_specification(
...
+        s_dynamics = self._app_vertex.synapse_manager.synapse_dynamics
+        if isinstance(s_dynamics, AbstractSynapseDynamicsStructural):
+            max_rewires_per_ts = s_dynamics.get_max_rewires_per_ts(
+                machine_time_step)
```
rowleya, line 474: This might end up in the constructor where the neuron recorder gets made where you can then pass it in to the constructor.

spynnaker/pyNN/models/common/neuron_recorder.py
andrewgait: When the NeuronRecorder is initially made (in the init of AbstractPopulationVertex), you don't yet know what max_rewires_per_ts will be - a Population always has to be built before a Projection to/from it, so the NeuronRecorder is built before the SynapseDynamics - you only know that these are structural when the Projection is built.  So this will have to be done slightly differently from passing it into the NeuronRecorder init... but it ought to be possible to effectively remove it from all the functions it's currently in and set it I guess.

rowleya: Ah OK, yes that is true!  But yes, a setter might be better to avoid this being passed around so much.  I guess this is the first time we have had something recorded on the synapses rather than the neurons.

andrewgait: I have now reworked this so that the value isn't passed around, and also changed to not include the "rewiring" parameter in the list of (neuron) recordable variables; instead this is now treated as an "event-per-core" variable, which gives us the option to add others in the future if we come up with any.

spynnaker/pyNN/models/common/abstract_neuron_recordable.py
```diff
@@ -107,3 +107,20 @@ def get_expected_n_rows(
         :param str variable: the variable being recorded
         :return: int the number of rows expected.
         """
+
+    @abstractmethod
+    def get_rewires(
```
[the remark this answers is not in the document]
rowleya: I think this got missed in the updates.  I wondered here if instead of get_rewires, you could call get_data("rewires".....  This would then avoid downstream changes, and is more consistent with the rest of the variables.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -154,6 +184,8 @@ def _neurons_recording(self, variable, vertex_slice):
         :param ~pacman.model.graphs.common.Slice vertex_slice:
         :rtype: iterable(int)
         """
+        if variable in self.__events_per_core_variables:
+            return range(vertex_slice.lo_atom, vertex_slice.hi_atom + 1)
```
rowleya, line 188: Is this needed for these types of variable?  I think they won't need the data that this function generates (like per-timestep variables)

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -355,6 +387,9 @@ def get_matrix_data(
         if variable in self.__per_timestep_variables:
             sampling_rate = 1
             data_type = self.__per_timestep_datatypes[variable]
+        elif variable in self.__events_per_core_variables:
+            sampling_rate = 1
+            data_type = self.__events_per_core_datatypes[variable]
```
rowleya, line 392: I think this call won't be made for events-per-core variables as they are handled via the get_rewires function?  Maybe a gate against this, like the above for bitfield variables, is needed.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -451,11 +486,104 @@ def get_spikes(
...
+        :param ~pacman.model.placements.Placements placements:
+            the placements object
+        :param application_vertex:
+        :type application_vertex:
+            ~pacman.model.graphs.application.ApplicationVertex
+        :param str variable:
+        :param int machine_time_step: microseconds
+        :return:
+        :rtype: ~numpy.ndarray(tuple(int,int,int,int))
+        """
```
rowleya, line 508: If this is the way to get event variables out, it should probably take the variable name, and the protect against getting other types of variables through this function.

spynnaker/pyNN/models/common/neuron_recorder.py
rowleya: Additionally note (because it won't let me do it directly), there should also be a gate in the get_spikes method in the same way.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -121,21 +138,34 @@ def __init__(
...
         self.__bitfield_variables = bitfield_variables
+
         self.__per_timestep_variables = per_timestep_variables
         self.__per_timestep_datatypes = per_timestep_datatypes
         self.__per_timestep_recording = set()
+
+        self.__events_per_core_variables = events_per_core_variables
+        self.__events_per_core_datatypes = events_per_core_datatypes
+        self.__events_per_core_recording = set()
+        self.__max_rewires_per_ts = 0
```
rowleya, line 149: To support events properly, this should probably become __max_events_per_ts and be a dictionary for each event variable.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -784,6 +969,12 @@ def get_buffered_sdram_per_record(self, variable, vertex_slice):
                 return 0
             size = self.__per_timestep_datatypes[variable].size
             return self._N_BYTES_FOR_TIMESTAMP + size
+        if variable in self.__events_per_core_variables:
+            if variable not in self.__events_per_core_recording:
+                return 0
+            size = self.__events_per_core_datatypes[variable].size
+            return self.__max_rewires_per_ts * (
```
rowleya, line 976: If this become max_events_per_ts dict, this would need to change to use the variable indicated.

spynnaker/pyNN/extra_algorithms/splitter_components/splitter_abstract_pop_vertex_slice.py
```diff
@@ -118,6 +121,12 @@ def get_variable_sdram(self, vertex_slice):
         :return: the variable sdram used by the neuron recorder
         :rtype: VariableSDRAM
         """
+        s_dynamics = self._governed_app_vertex.synapse_manager.synapse_dynamics
+        if isinstance(s_dynamics, AbstractSynapseDynamicsStructural):
+            machine_ts = get_simulator().machine_time_step
+            max_rewires_per_ts = s_dynamics.get_max_rewires_per_ts(machine_ts)
+            self._governed_app_vertex.neuron_recorder.set_max_rewires_per_ts(
+                max_rewires_per_ts)
```
rowleya, line 129: See later comments that will affect this call.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -1073,6 +1281,12 @@ def _get_data(self, vertex_slice):
 
         return numpy.concatenate(data)
 
+    def set_max_rewires_per_ts(self, max_rewires_per_ts):
+        """
+        :param int max_rewires_per_ts: the maximum rewires per timestep
+        """
+        self.__max_rewires_per_ts = max_rewires_per_ts
+
```
rowleya, line 1289: If this is going to be max_events_per_ts, this would need to take a variable name too, and should be gated to ensure it isn't called against other variable types.

spynnaker/pyNN/models/neuron/population_machine_vertex.py
```diff
@@ -467,6 +467,15 @@ def generate_data_specification(
...
             self.__binary_file_name, machine_time_step, time_scale_factor))
 
+        # If the dynamics are structural the neuron recorder needs to know
+        # the maximum rewires that could happend per timestep
+        s_dynamics = self._app_vertex.synapse_manager.synapse_dynamics
+        if isinstance(s_dynamics, AbstractSynapseDynamicsStructural):
+            max_rewires_per_ts = s_dynamics.get_max_rewires_per_ts(
+                machine_time_step)
+            self._app_vertex.neuron_recorder.set_max_rewires_per_ts(
+                max_rewires_per_ts)
```
rowleya, line 477: This will change if other comments are carried out.

rowleya (changes requested):
This is getting better, but a couple of things need to be carried forward to make the event recording complete

spynnaker/pyNN/models/common/abstract_neuron_recordable.py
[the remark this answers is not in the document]
andrewgait: I am not sure about this - spikes don't go through get_data("spikes"), for example, as far as I can tell?  The reason I did it this way was to avoid going through get_matrix_data(...), which is not what we want to do for rewires/events.  But I suppose we could go into the vertex's get_data(..) function and then go through to the neuron_recorder's get_rewires(..) function, which removes the need for a get_rewires() function in the vertex - is that what you mean?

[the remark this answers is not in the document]
rowleya: At this level spikes do go through get_data.  It is only the level below that factors this out into the spikes etc.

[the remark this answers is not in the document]
andrewgait: There's a get_spikes(...) function in the vertex - that's the level I'm talking about.  This comes from AbstractSpikeRecordable, rather than AbstractNeuronRecordable, though.  The only level where get_data("spikes") happens is at the Population level, I think - once you then get into the recorder's append_current_segment(...), it splits into get_spikes() and get_recorded_matrix(...).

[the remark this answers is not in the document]
rowleya: I don't think AbstractNeuronRecordable has get_spikes.  It might be that this is a mistake if things are already calling that!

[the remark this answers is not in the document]
andrewgait: This is true, get_spikes is in AbstractSpikeRecordable.  Perhaps we need AbstractEventRecordable as well?  Then this can be added to vertices as and when necessary (thus avoiding the downstream changes in e.g. SpiNNGym).

neural_modelling/src/neuron/structural_plasticity/synaptogenesis/topographic_map_impl.c
```diff
@@ -72,6 +79,17 @@ static circular_buffer current_state_queue;
...
+//! The recording region for the structural events
+uint32_t rewiring_recording_index;
+
+//! Struct for structural recording data
+typedef struct structural_recording_values_t {
+    uint32_t time;
+    uint32_t value;
+} structural_recording_values_t;
+
+structural_recording_values_t *structural_recording_values;
```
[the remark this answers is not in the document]
andrewgait: Done, also similar for later comments in this file.

spynnaker/pyNN/models/common/neuron_recorder.py
```diff
@@ -121,21 +138,34 @@ def __init__(
...
+        self.__max_rewires_per_ts = 0
+
         self.__region_ids = dict()
         region_id = 0
         for region_id, variable in enumerate(itertools.chain(
                 allowed_variables, bitfield_variables)):
             self.__sampling_rates[variable] = 0
             self.__indexes[variable] = None
             self.__region_ids[variable] = region_id
+        event_region_id = region_id + 1
```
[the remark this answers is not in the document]
andrewgait: Thanks - I think it's right now (or at least, the tests I expect to pass are passing).

spynnaker/pyNN/models/common/neuron_recorder.py
andrewgait: This isn't needed - a check has been added instead.

spynnaker/pyNN/models/common/neuron_recorder.py
andrewgait: I think I've guarded against this now by having _get_rewires(..) only called when get_events(variable, ..) with variable=REWIRING.

spynnaker/pyNN/models/common/neuron_recorder.py
andrewgait: Not needed any longer - moved into the _get_rewires function.

rowleya (approved):
I am happy with this now.  I have made a couple of "optional" things that could be changed to make this more "future compatible" if desired.  If not, please merge!

PR #1048, opened 2021-03-22
strip out max_delay as it was based on an INCONSTENT PyNN 

Christian-B:
strip out max_delay as it was based on an INCONSTENT PyNN value and only used for an incorrect warning

tested by https://github.com/SpiNNakerManchester/IntegrationTests/pull/36

### DIFF
--- a/spynnaker/pyNN/abstract_spinnaker_common.py
+++ b/spynnaker/pyNN/abstract_spinnaker_common.py
@@ -47,7 +47,6 @@ class AbstractSpiNNakerCommon(
         "__id_counter",
         "__live_spike_recorder",
         "__min_delay",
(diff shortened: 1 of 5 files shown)


Christian-B:
min_delay is used by SynapseDynamicsStatic and SynapseDynamicsSTDP if the delay is None

Christian-B:
Even without this value and regardless of what it was DelaySupportAdder throws an DelayExtensionException raise DelayExtensionException if it can not handle the delay based on what the post vertex says it can handle and what the max delay the DelayVertex can handle

Christian-B:
@rowleya Correctly points out "PyNN defines the min and max delays so we have to accept them even if to throw away".

PR #1049, opened 2021-03-23
Clear cache

Christian-B:
doing projection get weight both before and after run cause the before run to be returned both times.

This was due to SynapticMatrix caching the read values.

Fix 1. Clear the cache after reading pre data
Fix 2. Clear all caches after run so get reads in fresh data

Implemented both to be safe.


rowleya (approved):
Looks like it should work

PR #1051, opened 2021-03-24
Recorder database

Christian-B:
Depends on https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/774

Data is now cached in an SQL database in a user readable format.

Currently only used to cache data from previous segments or if specifically requested.

Medium term I could add methods to convert the data to Neo if anyone asks

Long term plans are to fill this database in the Java during data extraction.

The "Single" methods are designed for the long term where each core provides only 1 column (for example spike_counts)

### DIFF
--- a/spynnaker/pyNN/models/populations/population.py
+++ b/spynnaker/pyNN/models/populations/population.py
@@ -1040,10 +1040,10 @@ class Population(PopulationBase):
         raise ConfigurationException(
             "This population does not support describing its units")
 
-    def _cache_data(self):
+    def _cache_data(self, recorder_database):
         """ Store data for later extraction
         """
-        self._recorder.cache_data()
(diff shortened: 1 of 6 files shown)


rowleya:
Some steps that could come after this (i.e. in other PRs) are:
1. Add reader that makes the data end up in Neo format (as you suggested)
2. Make the code actually use this database when reading results normally
3. Create a "Neo IO" object that lets neo read from the database directly rather than converting to Neo
4. Make Java write the database directly (again as you suggested)

Christian-B:
"Make the code actually use this database when reading results normally"

Yes this is the long term plan but only makes sense if done one a per core basis otherwise it is just add extra step with no benefit unless we also have the Neo IO

Issue #1036, opened 2021-03-05
Untested binaries


dkfellows:
The following sPyNNaker binaries should be at least minimally tested (do they run?), at least in an overnight test:
- [x] `IF_cond_exp_stdp_mad_nearest_pair_additive.aplx`
- [ ] `IF_cond_exp_stdp_mad_nearest_pair_additive_structural_last_neuron_distance_weight.aplx`
- [ ] `IF_cond_exp_stdp_mad_pair_additive_structural_last_neuron_distance_weight.aplx`
- [ ] `IF_cond_exp_stdp_mad_pair_additive_structural_random_distance_weight.aplx`
- [ ] `IF_cond_exp_structural_last_neuron_distance_weight.aplx`
- [x] `IF_curr_alpha_stdp_mad_pair_additive.aplx`
- [ ] `IF_curr_exp_ca2_adaptive_stdp_mad_pair_additive.aplx`
- [x] `IF_curr_exp_stdp_mad_nearest_pair_multiplicative.aplx`
- [x] `IF_curr_exp_stdp_mad_pair_multiplicative.aplx`
- [ ] `IF_curr_exp_stdp_mad_recurrent_pre_stochastic_multiplicative.aplx`
- [x] `IZK_cond_exp.aplx`
- [x] `IZK_cond_exp_stdp_mad_pair_additive.aplx`
- [x] `IZK_curr_exp_stdp_mad_pair_additive.aplx`

The following SpiNNFrontEndCommon binaries should be tested, at least in an overnight test:
- [x] `spinn_front_end_common/common_model_binaries/bit_field_compressor.aplx`

<sup>This information was obtained from the Report stage of the any Jenkins job.</sup>

Christian-B:
@andrewgait commented "I got some scripts from Peter a while back that deal with the four "structural" binaries"

@andrewgait commented "The other two binaries in there I would suggest we should probably turn off the builds of them for now"

andrewgait:
The list in here is outdated now since the split synapse/neuron aplx files were introduced.  It may be worth keeping this issue open with an updated list of untested binaries.

Christian-B:
Better reporting done by https://github.com/SpiNNakerManchester/SpiNNUtils/pull/344

PR #1052, opened 2021-03-30
turn of build for two unused models

Christian-B:
turn of two binaries not being tested 
part of  https://github.com/SpiNNakerManchester/sPyNNaker/issues/1036

### DIFF
--- a/neural_modelling/makefiles/neuron/Makefile
+++ b/neural_modelling/makefiles/neuron/Makefile
@@ -23,12 +23,10 @@ MODELS = IF_curr_exp \
          IF_cond_exp_stdp_mad_pair_additive \
          external_device_lif_control \
          IF_cond_exp_stoc \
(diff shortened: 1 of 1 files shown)


dkfellows:
FWIW, I don't mind if we end up with a stupid integration test that instantiates these two and which passes if we can get _any_ spikes out of the far end. It wouldn't check that the numerics are correct, but it would at least ensure that there's no absolute howlers in the code.

rowleya:
For future reference as well, if we aren't sure what these are supposed to do mathematically, we can fix the functionality around what they currently do mathematically, on the basis that at least then if they change what they are doing, we can track it and decide if the change is correct.  If a user ever then tells us they are actually not doing the right thing, we can update them along with the test.

Christian-B:
In the end decided to keep these as plans are to add tests.

PR #1054, opened 2021-04-08
Move Development Status to setup.py

Christian-B:
This was the only place where Development Status was not hard coded into setup.py

Doing it the same everywhere make it easier to automate version bumping

If this goes in before the release will require the relase to be fixed buit that is not a bad thing.


Christian-B:
For slightly longer term we should unify the way spalloc and the other repositories do it.

This way is one less file that needs bumping

PR #1063, opened 2021-04-22
Add flake8 back for spynnaker_integration_tests, spynnaker8 directories

andrewgait:
flake8 was no longer running on spynnaker_integration_tests or spynnaker8 - also tidied up based on local run of flake8

Tested by https://github.com/SpiNNakerManchester/IntegrationTests/pull/45


rowleya:
This looks fine - just a thought that we should maybe also pylint spynnaker8?

andrewgait:
> This looks fine - just a thought that we should maybe also pylint spynnaker8?

Don't see why not - will add it.

PR #1066, opened 2021-04-28
Add test for output of record("packets-per-timestep")

andrewgait:
Add a test for the output of pop.record("packets-per-timestep").


rowleya (approved):
Looks like the right idea

PR #1069, opened 2021-05-04
removed dead cfg option

Christian-B:


dkfellows (approved):

dkfellows:
I'm forcing the commit in; pylint is failing, but it's due to changes in FEC that I wouldn't have merged if I'd known this bug was lurking.

PR #1071, opened 2021-05-06
Fix issue with master pop in some cases

rowleya:
Found an issue where it was possible to trigger the error check on the master pop table n_neurons.  The case is triggered by a pre-vertex with a splitting of atoms something like the following (a, a, a, b) where a > 2047 but b < 2047.  The check was being done on b but should have been done on a.  This fixes that, and adds a test to check it works, as well as adding a test to check for the maximum number of cores case.


andrewgait:
This fix looks fine to me, and it passes tests, so merging.

PR #1080, opened 2021-05-25
Test with graphviz installed

### LINKED ISSUE #1078: SpYNNakerNeuronGraphNetworkSpecificationReport untested requires graphviz
The report SpYNNakerNeuronGraphNetworkSpecificationReport  which is triggered by the cfg setting
[Reports]   
draw_network_graph=True

requires graphviz

Currently
import graphviz
is a delayed import hidden in a method
Suggesting that at some point someone had trouble with graphviz and a testing platform.
It also requires actual graphviz to be installed to run

So the questions is.
Do we want to to test this method?

If so how do we install graphviz in github action and on Jenkins.

Once Graphviz is installed be can trigger this by changing "draw_network_graph" to "write_network_graph"
or by adding 
draw_network_graph=True to one of the tests.

---
I also see nothing Spynnaker specifc or (needing a board) so would suggest if this is worth testing to
move it to FEC and test it with a unittest

dkfellows:
This fixes #1078 (or will do once the PR is finished)

### DIFF
--- a/.github/workflows/python_actions.yml
+++ b/.github/workflows/python_actions.yml
@@ -39,6 +39,10 @@ jobs:
         repository: SpiNNakerManchester/SupportScripts
         path: support
 
+    - name: Install Ubuntu dependencies
+      uses: ./support/actions/apt-get-install
+      with:
+        packages: graphviz
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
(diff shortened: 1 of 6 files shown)


Christian-B:
requirements.txt needs graphviz.

dkfellows:
OK, now we just need integration testing sorted out.

PR #1087, opened 2021-06-01
use svg for big files in SpYNNakerNeuronGraphNetworkSpecificationReport

Christian-B:
When using png on large files it reports
"graph is too large for cairo-renderer bitmaps"

The resulting svg file is USELESS. In the extreme cases just some gray shading.

Not sure about the cutoff


rowleya (approved):
No idea about the cutoff either, but this seems fine!

PR #1095, opened 2021-06-24
Speedup Big FromListConnector

Christian-B:
Depends on:
https://github.com/SpiNNakerManchester/PACMAN/pull/398
tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/64

BUG FIX: SynapseIORowBased
  Now asks splitters for the slices!

Method could_connect changed to use vertexes
    This allows __split_conn_list  to be used!

__split_conn_list now maps only pairs of slices that connect
  Instead of checking if not none and then length of item can now just check in

_split_connections
  always sets _split_connections to a dict even if empty
  Only maps if len(indices) > 0
  exit because slices seen is most common so moved first

create_synaptic_block
   exit fast if not in __split_conn_list which includes the not len(self.__sources) case

tests show partitioner  time cut by over 80 in large fromLists
for example:
import numpy
import spynnaker8 as sim

n_sources = 100000
n_targets = 100000
n_connections = 10000000
sources = numpy.random.randint(0, n_sources, n_connections)
targets = numpy.random.randint(0, n_targets, n_connections)
connection_list = numpy.dstack((sources, targets))[0]

sim.setup(timestep=1.0)
sim.set_number_of_neurons_per_core(sim.IF_curr_exp, 100)

pop_1 = sim.Population(n_sources, sim.IF_curr_exp(), label="pop_1")
input = sim.Population(n_targets, sim.SpikeSourceArray(spike_times=[0]), label="input")
input_proj = sim.Projection(input, pop_1, sim.FromListConnector(connection_list),
                            synapse_type=sim.StaticSynapse(weight=5, delay=1))
simtime = 10
sim.run(simtime)

In Master:
0:04:43.446246 taken by SpYNNakerSplitterPartitioner 
This PR:
0:00:38.930407 taken by SpYNNakerSplitterPartitioner


unittests/connector_tests/test_from_list_connector.py
```diff
@@ -151,6 +151,31 @@ def test_connector_split():
...
+        return (self.slices, True)
+
+
+class MockAppVertex(object):
+
+    def __init__(self, slices):
+        self.splitter = MockSplitter(slices)
+
+
+class MockMachieVertex(object):
```
andrewgait, line 172: MockMachieVertex -> MockMachineVertex

andrewgait (approved):
Looks fine to me and worth having in, just one minor spelling mistake

unittests/connector_tests/test_from_list_connector.py
Christian-B: fixed

PR #1110, opened 2021-09-22
Make return value from projection.get(.., 'list') usable in FromListConnector

andrewgait:
Not sure it's particularly a bug, but a user recently pointed out that the return value from `projection.get(.., 'list')` couldn't then be used as directly the argument in a `FromListConnector`; it turns out something goes strange with the formatting in the `ConnectionHolder`.  This is an attempt to fix that (and provide a test for it).

### DIFF
--- a/spynnaker/pyNN/models/neuron/connection_holder.py
+++ b/spynnaker/pyNN/models/neuron/connection_holder.py
@@ -171,15 +171,22 @@ class ConnectionHolder(object):
             # all the data
             if (self.__data_items_to_return is None or
                     not self.__data_items_to_return):
-                self.__data_items = connections[order]
+                data_items = connections[order]
             # There is more than one item to return, so let numpy do its magic
             elif len(self.__data_items_to_return) > 1:
-                self.__data_items = \
+                data_items = \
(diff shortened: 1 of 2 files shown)


spynnaker/pyNN/models/neuron/connection_holder.py
```diff
@@ -171,15 +171,25 @@ def _get_data_items(self):
...
+                data_items = \
                     connections[order][self.__data_items_to_return[0]]
+
+            # Return in a format which can be understood by a FromListConnector
+            self.__data_items = []
+            for data_item in data_items:
+                data_item_list = data_item
+                if hasattr(data_item_list, "__len__"):
+                    data_item_list = [
+                        data_item[n] for n in range(len(data_item))]
```
rowleya, line 190:
```suggestion
                    data_item_list = list(data_item)
```

rowleya: Not sure if that will work in these cases though

andrewgait: That does seem to work on the tests that were initially failing - running full tests again now...

PR #1117, opened 2021-10-05
Trying to fix a crash...

dkfellows:
Let's see if this works…


dkfellows:
[This](https://github.com/SpiNNakerManchester/sPyNNaker/runs/3805234082?check_suite_focus=true) says this (updating the version of the doxygen action we're using) worked.

Christian-B:
If it works no reason not to merge

PR #1120, opened 2021-10-21
Neuromodulation agdr

rowleya:
Note: This isn't a merge to master!

This changes quite a bit about neuromodulation, but is based on the changes already there.  The main changes are that neuromodulation is now done in plasticity, so it avoids any overhead when using static synapses, and the code now uses the (modified) standarad STDP rules, so it should work with any of them.


spynnaker_integration_tests/test_master_pop/spynnaker.cfg
```diff
@@ -1,8 +1,2 @@
-[Reports]
-extract_iobuf = True
-extract_iobuf_from_cores = 0,0,2
-extract_iobuf_from_binary_types = IF_curr_exp.aplx,reverse_iptag_multicast_source.aplx
-clear_iobuf_during_run = True
-
 [Mapping]
```
andrewgait, line 1: This change doesn't appear to be related to neuromodulation but I agree it should probably be done!

andrewgait (approved):
This looks fine in general to me, so approved to merge into the neuromodulation branch.  Just a few minor comments to consider (which can be changed here or after merging, whichever is easiest).  I also wondered whether, now that it's possible, you've tried out any examples of using the more complicated weight / timing rules combined with neuromodulation?  (Again this could wait until after merging into the neuromodulation branch if desired).

spynnaker_integration_tests/test_master_pop/spynnaker.cfg
rowleya: I think it was a failing test though I can't quite remember what caused it...  Possibly the fact that one of the binaries wasn't running any more.  The test seemed to be unrelated to the config though...

andrewgait:
Merging this into neuromodulation now, will look at LOG_LEVEL / other weight & timing rules with neuromodulation on in that branch.

PR #1122, opened 2021-10-27
No need for code in case there is no app graph

Christian-B:
part of  https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/884

Should not check for report no longer generated.


Christian-B:
done elsewhere

PR #1124, opened 2021-11-01
Fix dependency issue

rowleya:
There appears to be a dependency issue; this appears to fix it.


Christian-B:
Approved. merge as soon as you are happy with testing

PR #1125, opened 2021-11-01
Global Data model 

Christian-B:
Part of https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/886


Christian-B:
When compared to #887 @rowleya says "I think the new one makes sense".

As I agree with him I am closing this PR.

PR #1136, opened 2021-11-26
Meanfield model

garandel:
I would like put this on the meanfield_model branch.
That's working in the first order except that sigma_V is not yet to the sqrtk().
I couldn't add sqrtk() because otherwise I overflowed ITCM.


andrewgait (changes requested):
This looks in general fine to me; there are a few considerations you can make (see comments) as to where you can avoid some divides.  I don't think these will help hugely with making enough space to use sqrtk() - we may have to think of another way around that.  But I would be happy to merge this once you've addressed the comments and we can discuss what might be possible for sqrtk() after that.

garandel:
Hi Andrew,
yes I noticed before sending to the remote that I could do these
modifications, that I did on another branch on my local repo.
The problem is that it overflowed more with multiplication than with
division.
I think that the main reason is because the multiplication selected for
that is a too big memory needed.
So my question is, could I select one particular multiplication without
writing the name of the function and passing to " / "
 symbol?

On the division side I wonder also if I could use divergente series in
order to approximate the result efficiently?
Same idea for the square root but I didn't know any divergente series for
the square root, yet!! ^^

I was fight with this issue so sending this intermediate code was a step in
order to know the procedure.
But now that I know it, I will send it to you when the job is done.

Sincerely

Thomas


Le lun. 29 nov. 2021 à 10:04, Andrew Gait ***@***.***> a
écrit :

garandel:
rectification it's " * " symbol and not " / " one sorry.
Also the problem with Cm and Gl is that there exist also formulas with
multiplication, so the two possibilities are keep like this and do the
division or have two more constante with Gl and Cm so 1/Gl and 1/Cm.

Have a nice day

Thomas

Le lun. 29 nov. 2021 à 13:44, thomas garandel ***@***.***> a
écrit :

PR #1138, opened 2021-12-01
Key translation

rowleya:
Update to make use of Key Translation from SpiNNakerManchester/SpiNNFrontEndCommon#901 and add an example.


andrewgait (changes requested):
Looks fine to me, though as it's adding a new test I'd suggest running an integration test branch as well just to be safe.

rowleya:
You were right to request this - the test is indeed failing although I am struggling to work out why since it works for me... Further investigation is in progress

rowleya:
Tested by http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/key_translation/3/pipeline

andrewgait (approved):

PR #1139, opened 2021-12-01
Fix Live Voltage Test

rowleya:
Don't assume ordering in the live voltage test

### DIFF
--- a/spynnaker_integration_tests/test_external_devices/test_live_neuron_voltage.py
+++ b/spynnaker_integration_tests/test_external_devices/test_live_neuron_voltage.py
@@ -133,11 +133,10 @@ def live_neuron_voltage():
     assert(numpy.sum(translator_2.voltages[key_2_1]) == 0)
     assert(numpy.sum(translator_2.voltages[key_2_2]) == 0)
 
-    i = 0
+    # We can get packet loss so don't assume all voltages exist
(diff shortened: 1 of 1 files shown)


spynnaker_integration_tests/test_external_devices/test_live_neuron_voltage.py
```diff
@@ -133,11 +133,10 @@ def live_neuron_voltage():
     assert(numpy.sum(translator_2.voltages[key_2_1]) == 0)
     assert(numpy.sum(translator_2.voltages[key_2_2]) == 0)
 
-    i = 0
+    # We can get packet loss so don't assume all voltages exist
```
Christian-B, line 136: This comments say you can loose packets but code expects all there just reordered.

Christian-B (changes requested):

spynnaker_integration_tests/test_external_devices/test_live_neuron_voltage.py
rowleya: It expects all the ones that are received to actually be recorded; those can't go missing.

rowleya:
Latest test passes: http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/fix_live_voltage_test/2/pipeline/

PR #1147, opened 2022-01-21
Multirun packets-per-timestep fix

andrewgait:
Discovered while running other code that packets-per-timestep recording was broken for multi-run; this fixes it and adds a test for multi-run recording with packets-per-timestep.

Tested by https://github.com/SpiNNakerManchester/IntegrationTests/pull/95


Christian-B:
Thanks. Green to merge when Integration tests start ran

Christian-B:
updated to master to pickup setuptools cap

PR #1151, opened 2022-02-22
Whole board test

Christian-B:
Adds a test to check a whole board.

It creates synfires chains that make sure to use all links in a particular direction.
Doing all 6 directions.
The repeat until all cores used.

Check each synfire makes it to the end.

Tested by: (in progess)
http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/whole/1/pipeline

TODO:
What is the best way to run this on all or at least all available boards.


rowleya:
Changes are now dependent on SpiNNakerManchester/SpiNNFrontEndCommon#923

rowleya:
Changes tested by SpiNNakerManchester/IntegrationTests#101.

Note previous runs tested the actual machine test, but this will only now get run on the cron job.

test_whole_board/test_whole_board.py
```diff
@@ -0,0 +1,210 @@
...
+    ur = [(0, 0), (0, 1), (0, 2), (0, 3),
+          (1, 0), (1, 1), (1, 2), (1, 3), (1, 4),
+          (2, 0), (2, 1), (2, 2), (2, 3), (2, 4), (2, 5),
+          (3, 0), (3, 1), (3, 2), (3, 3), (3, 4), (3, 5), (3, 6),
+          (4, 0), (4, 1), (4, 2), (4, 3), (4, 4), (4, 5), (4, 6), (4, 7),
+          (5, 1), (5, 2), (5, 3), (5, 4), (5, 5), (5, 6), (5, 7),
+          (6, 2), (6, 3), (6, 4), (6, 5), (6, 6), (6, 7),
+          (7, 3), (7, 4), (7, 5), (7, 6), (7, 7)]
+    dl = list(ur)
+    dl.reverse()
```
andrewgait, line 56: I'm a bit confused as to what the difference is between "up" and "ur" (I assume short for "up right")?  I don't see how the code leads to different chains when using "ur" compared to "up" - is it just that once you've gone up, down, left, right, then you can only next go "up right" and then "down left", and so it picks the correct next chip from the list?

andrewgait: (Comment made before "ur" was changed to its current state...)

rowleya: I changed this at almost exactly the time you commented!

rowleya: Testing done manually and the example still works with the given changes

Christian-B: The Chips are added to the synfire chain in the order in the list. So up goes from (0, 0) to (0,1) which is in the up position. ect.  UR goes from (0,0) to (1, 4) which is over the up right link.

note: I dont check if the link exists just the chip. If the link os down the router will find an alternative route and not break the chain.

rowleya: The issue was that in the original version, up == ur (see 56fc4f3fba349177cadb4695b91d2a063cfd553b)

rowleya: In reality, the network run here is unimportant.  The only important bits are that it runs on boards that work and produces the expected results, and that it actually runs neural network code on all cores to test that they work when used with that code.

andrewgait (approved):
Happy with this as it stands too, though I do wonder whether the idea of "max_gen_data" might need extending to more than the synaptic data.  But this can go in as it stands now and we can think about that more if the need arises.

rowleya:
> Happy with this as it stands too, though I do wonder whether the idea of "max_gen_data" might need extending to more than the synaptic data. But this can go in as it stands now and we can think about that more if the need arises.

I think that was just to handle the case that was commonly failing; bitfield expansion appeared to be notorious for it when testing.  It could indeed be expanded if other cases arise.

PR #1153, opened 2022-03-09
Fix recording for rewiring and packets to work after reset

andrewgait:
@Christian-B is currently working with what happens with the code after calls to reset() and noticed that recording may be broken for the "rewiring" variable; tested and sure enough he was correct, and in addition "packets-per-timestep" was broken too.  This is because the region offset was mistakenly being added again to the recording region IDs for these variables.

Whilst I was looking at this I also fixed the structural plasticity cases after reset so that if a user specifies the same seed pre- and post-reset then they get the same answer.


rowleya (approved):
Looks like it works!

PR #1157, opened 2022-03-17
Unsynchronize Poisson sources

rowleya:
For some reason all of the Poisson source cores were using the same seed.  This should ensure that this is not the case.


spynnaker/pyNN/utilities/utility_calls.py
```diff
@@ -308,22 +308,20 @@ def _validate_mars_kiss_64_seed(seed):
...
+    :type rng: ~numpy.random.RandomState
     :param seed:
         the seed to create a random number generator if not handed.
     :type seed: int or None
     :return: a list of 4 ints which are used by the mars64 kiss random number
         generator for seeds.
     :rtype: list(int)
     """
-    if rng is None:
-        rng = numpy.random.RandomState(seed)
```
andrewgait, line 326: I was a bit worried about taking this out because I was sure that rng=None was being passed in here, however I've just noticed it never is - the confusion is because in the init of SpikeSourcePoissonVertex self.__rng is set to None at line 132, but then later on line 255 (still in the init) it is actually set to a value.  I'd suggest that line 132 should be removed from that file to avoid any further confusion in the future.

andrewgait (changes requested):
Looks fine to me, just one change related to this that I'd suggest.

spynnaker/pyNN/utilities/utility_calls.py
rowleya: Agreed - fixed!

andrewgait (approved):

Issue #956, opened 2021-01-04
Neo version upgrade needed


Christian-B:
Integration testing detected
Running PyNN-0.9.6/setup.py 
....
error: neo 0.6.1 is installed but neo>=0.8.0 is required by {'pynn'}

the latest release of Neo is 0.9
However
https://neo.readthedocs.io/en/stable/releases/0.9.0.html
reports 2 changes that will need work.
- Python 3 only
- Group and ChannelView replace Unit and ChannelIndex

As to support one of our uses that suffers with python 2,.7 we (at least for 2.7) are limited to 0.8

Examination is required to see how hard it would be to support both 0.8 and 0.9 at the with the same code base.

The suggestion is therefor to 
- move to 0.8 ASAP

Then leave this issue OPEN (but remove blocker and dump milestone)  to move to 0.9 if possible while still supporting 0.8 or if not possible after the dropping of 2,7 by us

Christian-B:
Code now upgrade to allow up to neo 0.8

Moving to Neo 0.9 is a bigger project so left for later.

Christian-B:
Code as it is still works with neo 0.9
Channel+index is deprecated but still exists.

Unable to upgrade to the new way as suggeted in ;

dkfellows:
Appears to have been solved.

PR #1165, opened 2022-04-13
Allow neo and pyNN versions 0.10

Christian-B:
Fixes: https://github.com/SpiNNakerManchester/sPyNNaker/issues/956

We had capped pyNN and Neo due to a bug in the plotting
https://github.com/NeuralEnsemble/PyNN/issues/708  whuich has been fixed

there is another.
https://github.com/NeuralEnsemble/PyNN/issues/750
But there is a temporary work around (revert back) to avoid that one.

This PR allow support for Both version 0.9 and 0.10 and the same time.

I suggest once pynn fixes https://github.com/NeuralEnsemble/PyNN/issues/708 or tells us what we are doing wrong
We support only from that version up.

tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/111

backward compatibility to pyNN/neo 0.9 tested by
https://github.com/SpiNNakerManchester/sPyNNaker/pull/1166
https://github.com/SpiNNakerManchester/IntegrationTests/pull/112

If this is merged address
https://github.com/SpiNNakerManchester/SpiNNakerManchester.github.io/issues/40


Christian-B:
Agreed with @rowleya to wait for the fix of https://github.com/NeuralEnsemble/PyNN/issues/708

PR #1168, opened 2022-04-14
Record verify

Christian-B:
checksome of the params going into record


Christian-B:
sampling interval check is wrong.

correct one done by neuron_recorder

to_file not use enough t be worth checking.

closing pr.

Issue #1170, opened 2022-04-22
Better call to clear_connection_cache


Christian-B:
Currently before and after a run 
there is a loop through projections to call vertex.clear_connection_cache

Would it not be better to just loop over the Vertices?
If yes that will be much easier after GlobalData

Also does it have to happen every run or only after a reset?
If After a reset could it somehow be called during reset

rowleya:
If it is easier to loop over vertices that should be fine to do.  I think it is necessary to check that the vertex supports the right interface for this of course (presumably something like accepts incoming synapses); those in projections definitely do as that is checked, but the check can be done in a loop too.

The clearing has to happen before a *new* run i.e. before the second and subsequent runs.  It shouldn't happen immediately after a run either as that is too early; the user will only request the data after control is returned back to them after a run.  The data is cached in case they ask for weights and then delays separately for example as the data is extracted all at once even if the user asks for it in parts.

Note that if the weights don't change during a run (for which I believe there is an interface) it isn't technically necessary to clear the cache since subsequent requests for the weights and delays should then result in the same data being returned...

Christian-B:
currently it is if isinstance(post_vertex, AbstractAcceptsIncomingSynapses):

Christian-B:
Looking deeper.
the only AbstractAcceptsIncomingSynapses is AbstractPopulationVertex
which iterates over the machine vertices if they isinstance HasSynapses

rowleya:
> The clearing has to happen before a _new_ run i.e. before the second and subsequent runs. It shouldn't happen immediately after a run either as that is too early; the user will only request the data after control is returned back to them after a run. The data is cached in case they ask for weights and then delays separately for example as the data is extracted all at once even if the user asks for it in parts.

See #1171 for why this isn't true.

PR #1171, opened 2022-04-22
Put projection._clear_cache back

Christian-B:
The calling of projection._clear_cache() was lost during the merge of spinnaker and asc

This puts it back.

Note in https://github.com/SpiNNakerManchester/sPyNNaker/issues/1170 @rowleya states
The clearing ... It shouldn't happen immediately after a run either as that is too early; the user will only request the data after control is returned back to them after a run. The data is cached in case they ask for weights and then delays separately for example as the data is extracted all at once even if the user asks for it in parts.

This PR put back as was before so does call it after run
see https://github.com/SpiNNakerManchester/sPyNNaker/blob/a45486217280d9d8ecf45f0e39210478b884774a/spynnaker/pyNN/abstract_spinnaker_common.py#L273

tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/118


rowleya:
I guess clearing before or after is actually OK thinking about it (and both does no harm if a little bit over the top).  The thing I missed when saying "calling immediately after run is too soon" is that the user can't request the connections until after the run is finished anyway, and the clear doesn't happen until they next call run.  So in between, they can get connections as many times as they like, and second and subsequent calls will then get cached versions.  So this is fine as it is.

PR #1179, opened 2022-05-16
spinnaker_get_data

Christian-B:
spinnaker_get_data now has an extra optional bool param as_matrix
If set will return V as a 2d matric ids * timestamps, (Without headers)

spinnaker_get_data enabled on views

better error with spinnaker_get_data("rewirings")

remove message that spinnaker_get_data may not be supported.


andrewgait (approved):
This looks fine to me - is it tested?

PR #1167, opened 2022-04-14
Strickter Pylint

Christian-B:

PR #1181, opened 2022-05-18
removed private property

Christian-B:
removed the private property  __get_max_delay
added __update_max_delay method
for get use using already existing and required max_support_delay() method instead.

Note: There where two different ways to calculate self.__allow_delay_extension

from init
 if (self.__max_delay is not None and
                self.__allow_delay_extension is None):
            self.__allow_delay_extension = True

from __get_max_delay/__update_max_delay
if self.__allow_delay_extension is None:
            self.__allow_delay_extension = max_delay_bits > final_n_delay_bits

Issue was that after a reset the first was not called so the second would apply.

Which is correct if the USER supplies a max_delay and no allow_delay_extension

### DIFF
--- a/spynnaker/pyNN/extra_algorithms/splitter_components/splitter_abstract_pop_vertex_neurons_synapses.py
+++ b/spynnaker/pyNN/extra_algorithms/splitter_components/splitter_abstract_pop_vertex_neurons_synapses.py
@@ -92,8 +92,8 @@ class SplitterAbstractPopulationVertexNeuronsSynapses(
         "__max_delay",
         # The user-set maximum delay, for reset
         "__user_max_delay",
-        # Whether to allow delay extensions to be created
-        "__allow_delay_extension",
+        # Whether you expect delay extensions to be asked to be created
+        "__expect_delay_extension",
         # The user-set allowing of delay extensions
         "__user_allow_delay_extension",
         # The fixed slices the vertices are divided into
@@ -135,8 +135,15 @@ class SplitterAbstractPopulationVertexNeuronsSynapses(
         self.__n_synapse_vertices = n_synapse_vertices
         self.__max_delay = max_delay
         self.__user_max_delay = max_delay
-        self.__allow_delay_extension = allow_delay_extension
         self.__user_allow_delay_extension = allow_delay_extension
+        if max_delay is None:
+            # to be calcutaed by __update_max_delay
+            self.__expect_delay_extension = None
+        else:
+            # The user may ask for the delay even if then told no
+            self.__expect_delay_extension = True
+            if allow_delay_extension is None:
+                self.__user_allow_delay_extension = True
         self.__slices = None
         self.__next_synapse_index = 0
 
@@ -595,7 +602,11 @@ class SplitterAbstractPopulationVertexNeuronsSynapses(
         self.__synapse_vertices = None
         self.__synapse_verts_by_neuron = None
         self.__max_delay = self.__user_max_delay
-        self.__allow_delay_extension = self.__user_allow_delay_extension
+        if self.__user_max_delay is None:
+            # to be calcutaed by __update_max_delay
+            self.__expect_delay_extension = None
+        else:
+            self.__expect_delay_extension = True
 
     @property
     def n_synapse_vertices(self):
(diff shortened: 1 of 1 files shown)


Christian-B:
Discovered by https://github.com/SpiNNakerManchester/sPyNNaker/pull/1167
But separate PR due to __allow_delay_extension  question

rowleya (approved):

rowleya:
1. If the user specifies max_delay and doesn't allow delay extensions, then delay extensions are not allowed.  This is a user preference.  If the max_delay can't work because of bit space, this should be an error.  If a delay comes in that is bigger than max_delay, this will result in an error.
2. If the user specifies max_delay and does allow delay extensions, then delay extensions can be used when delays are bigger than max_delay.  This should always work basically.
3. If the user doesn't specify max_delay then this is the same as cases 1 and 2 but with max_delay being calculated based on the bits available.

rowleya:
I should add:
4. If the user specifies max_delay and doesn't say whether to allow delay extensions, allowing delay extensions is based on whether there are enough bits to do it without them or not (if there are enough bits, don't allow delay extensions).
5. If the user doesn't specify either, just allow delay extensions.  Arguably that could also be the case for 4 to avoid confusion.

andrewgait:
> I wonder if the above means that allow_delay_extensions should be true by default, and no longer calculated...

It certainly seems that way from what you've said - then the only instance where it's turned off is when the user wants to do so (and understands the issues with having done so).  If the user sets max_delay too high for the number of bits available for delay extensions then error (as we already do).

rowleya:
Yes, I agree.  I think I over-thought the cases a bit.  Arguably the code uses the incoming projections to work out whether delay extensions are needed or not before turning them off or on, but then it doesn't really need to since if they are not needed, they won't be used!  I guess it was a belt-and-braces approach to make sure the two sides agreed...

Christian-B:
Having looked deeper I notice that values have double meaning

1. The Spliiter has the max_delay the user has specified

2. The splitter calculates the biggest delay that any projection asks for. (ignoring DelayExtensionVertex)

3. The splitter calculates the biggest delay the vertex could support

self.__max_delay and therefor max_support_delay is the users value or if None LOWER of the two.

Currently NO error is thrown if the user specifies a delay bigger than the splitter would calculate is possible

---
allow_delay_extension also has several meanings

1. user specified so believe the user

2. system determined that no delay would ever be needed 

---

My proposal is.
If the user supplies max_delay trust the user (or GIGO)
If the user supplies max_delay require the user to also supply  allow_delay_extension
     - They dont like the splitters way of calculating delays
    
If the user does not supply max_delay then
accepts_edges_from_delay_vertex (Only usage of allow_delay_extension) either returns 
True: Yes I was expecting you to add Delay vertex
Exception: I was not expecting you to add Delay Vertex

If the user does not supply max_delay but does supply allow_delay_extension 
ignore True and act like None. If false error when we calc that a delay will be needed.

Christian-B:
allow_delay_extension is specifically documented as "max_delay  is provided, this will default to True"
So will keep the behaviour as that

Christian-B:
@rowleya  said "The max_delay specifies the maximum number of delay slots on the core"

But which number of delau slots? The max possible to the max calcuated to be requested (if less)

rowleya:
My suggestion is that the number of delay slots be removed as a parameter, and always calculated to be the max possible since using that value should have no major effect (hopefully).  I believe the current calculation allows half the DTCM to be used for ring buffers.

Christian-B:
@rowleya said "current calculation allows half the DTCM to be used for ring buffers"

In this splitter MAX_RING_BUFFER_BITS = 14

This is then split into three part
bits for neuron id
bits of synapse_type
remainder for the delay

method create_machine_vertices checks this and raises an exception if the three together are too big.

Christian-B:
regarding:
splitter.max_support_delay

I see that AbstractSpynnakerSplitterDelay has this hard coded as 16
Completely ignoring the number of delays actually needed.

So why should this splitter the need to lower the value from max possible to what was actually needed is not understood.

My suggestion is to just return the max possible which then saves the work of asking all the incoming_projections

rowleya:
The delay splitter probably shouldn't do that.  There is an issue when the max delay is mixed though because of how the delay system works.  Arguably, there needs to be a different delay vertex for each post-vertex max delay encountered.  This is because each delay stage should be the same as the post-vertex max-delay.  I would guess this needs looking at again...

Christian-B:
This is a BLOCKER as I want to see if this is the correct approach for pylint_fix which is then a BLOCKER due to the changes in pylint!

Christian-B:
Alternatively I can do a smaller change to make pylint happy and look at the bigger change later

rowleya (approved):

Christian-B:
Merged into pylint_fix after approval (subject to spelling fix) from @rowleya and @andrewgait

Issue #1182, opened 2022-05-19
set_number_of_neurons_per_core vs already existing Populations


Christian-B:
In all our examples we do
set_number_of_neurons_per_core before creating the Populations

So
sim.set_number_of_neurons_per_core(sim.IF_curr_exp, 2)
pop_1 = sim.Population(10, sim.IF_curr_exp(), label="pop_1")

Will result in 5 Machine Vertexes

But
pop_1 = sim.Population(10, sim.IF_curr_exp(), label="pop_1")
sim.set_number_of_neurons_per_core(sim.IF_curr_exp, 2)

results in 1 Machine Vertex

So the questions is this the desired behaviour

If yes I suggest a warning if set_number_of_neurons_per_core is called after Population

If no on a set_number_of_neurons_per_cor iterate through all Populations to find any existing ones and limit them too.

andrewgait:
I think I'd say no, because the set_number_of_neurons_per_core does specify which model (not Population) that you wish to do this for.  So yes, I would iterate through all current Populations to find any matching the specified model and limit them too.

PR #1184, opened 2022-05-26
app vertex max_atoms_per_core

Christian-B:
Part of https://github.com/SpiNNakerManchester/PACMAN/pull/433

Fixes:
https://github.com/SpiNNakerManchester/sPyNNaker/issues/1182
by warning that it will not affect already created Populations

We currently dont support the raising of 
sim.set_number_of_neurons_per_core
Mainly because I was too lazy to work out how to get and check the model hard cap
And there was no known need for it.

beyond the model cap give a clear error that is not allowed
higher than a previously lower value give an error that we dont allow that at the moment

pop.set_max_atoms_per_core
can now also raise a previously lowered value but only up to the cap for that model.
That is both the model hard cap and the sim.set_number_of_neurons_per_core soft cap.

### DIFF
--- a/spynnaker/pyNN/spinnaker.py
+++ b/spynnaker/pyNN/spinnaker.py
@@ -514,7 +514,7 @@ class SpiNNaker(AbstractSpinnakerBase, pynn_control.BaseState):
             if previous < max_permitted:
                 raise SpynnakerException(
                     f"Attempt to increase number_of_neurons_per_core "
-                    f"from {previous} to {previous} not supported")
+                    f"from {previous} to {max_permitted} not supported")
         neuron_type.set_model_max_atoms_per_core(max_permitted)
         self.__neurons_per_core_set.add(neuron_type)
         if self._populations:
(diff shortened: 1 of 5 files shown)


spynnaker/pyNN/spinnaker.py
```diff
@@ -511,12 +512,14 @@ def set_number_of_neurons_per_core(self, neuron_type, max_permitted):
         if hasattr(neuron_type, "get_max_atoms_per_core"):
             previous = neuron_type.get_max_atoms_per_core()
             if previous < max_permitted:
-                logger.warning(
-                    "Attempt to increase number_of_neurons_per_core "
-                    "from {} to {} ignored", previous, max_permitted)
-                return
+                raise SpynnakerException(
+                    f"Attempt to increase number_of_neurons_per_core "
+                    f"from {previous} to {previous} not supported")
```
rowleya, line 517: Same argument there - I assume you meant one to be different!

rowleya (approved):
Fine once passing and error is fixed!

PR #1189, opened 2022-06-06
Current source fix

andrewgait:
Fix a bug in current sources which meant that certain examples didn't work (particularly with multiple neurons connected to the same current source).

Tested by https://github.com/SpiNNakerManchester/IntegrationTests/pull/121


rowleya (approved):
Looks good.  Is it worth adding the case that caused the error to the integration tests?

andrewgait:
> Looks good. Is it worth adding the case that caused the error to the integration tests?

Yes - done.

Issue #1186, opened 2022-05-31
Splitters not working with self-connections and neuromodulation


andrewgait:
Example below.  With splitter, weights returned on STDP self-connection on post-pop as [[0, 1, 0.0], [1, 0, 0.0]].  Without splitter, weights returned on STDP self-connection on post-pop as [[0, 1, 13.53515625], [1, 0, 0.0]] (which I think is correct...).

```
import pyNN.spiNNaker as sim

from spynnaker.pyNN.extra_algorithms.splitter_components import (
    SplitterAbstractPopulationVertexNeuronsSynapses)

timestep = 1.0
duration = 3000

# Main parameters from Izhikevich 2007 STDP paper
t_pre = [1500, 2400]    # Pre-synaptic neuron times
t_post = [1502]         # Post-synaptic neuron stimuli time
t_dopamine = [1600]     # Dopaminergic neuron spike times
tau_c = 1000            # Eligibility trace decay time constant.
tau_d = 200             # Dopamine trace decay time constant.
DA_concentration = 0.05  # Dopamine trace step increase size

# Initial weight
rewarded_syn_weight = 0.0

cell_params = {
    'cm': 0.3,
    'i_offset': 0.0,
    'tau_m': 10.0,
    'tau_refrac': 4.0,
    'tau_syn_E': 1.0,
    'tau_syn_I': 1.0,
    'v_reset': -70.0,
    'v_rest': -65.0,
    'v_thresh': -55.4}

sim.setup(timestep=timestep)

pre_pop = sim.Population(2, sim.SpikeSourceArray, {
    'spike_times': t_pre})

# Create a population of dopaminergic neurons for reward
reward_pop = sim.Population(2, sim.SpikeSourceArray, {
    'spike_times': t_dopamine}, label='reward')

# Stimulus for post synaptic population
post_stim = sim.Population(2, sim.SpikeSourceArray, {
    'spike_times': t_post})

# Create post synaptic population which will be modulated by DA concentration.
post_pop = sim.Population(
    2, sim.IF_curr_exp, cell_params,
    label='post1', additional_parameters={
        "splitter": SplitterAbstractPopulationVertexNeuronsSynapses(1)})

# Stimulate post-synaptic neuron
sim.Projection(
    post_stim, post_pop,
    sim.AllToAllConnector(),
    synapse_type=sim.StaticSynapse(weight=6),
    receptor_type='excitatory')

# Stimulate from pre pop
sim.Projection(
[...]

PR #1191, opened 2022-06-07
Allow STDP on self-projections

andrewgait:
Fixes https://github.com/SpiNNakerManchester/sPyNNaker/issues/1186 for both neuromodulated and non-neuromodulated STDP.

[Previous comment: Discussions of https://github.com/SpiNNakerManchester/sPyNNaker/issues/1186 via Skype highlighted that there were also problems with non-neuromodulated STDP on self-projections.  This fixes those cases but there still remain issues when the STDP is neuromodulated, for which discussions and investigations are ongoing.]

Tested by https://github.com/SpiNNakerManchester/IntegrationTests/pull/122

### DIFF
--- a/spynnaker/pyNN/extra_algorithms/splitter_components/splitter_abstract_pop_vertex_neurons_synapses.py
+++ b/spynnaker/pyNN/extra_algorithms/splitter_components/splitter_abstract_pop_vertex_neurons_synapses.py
@@ -488,8 +488,8 @@ class SplitterAbstractPopulationVertexNeuronsSynapses(
                     neuron_vertex, synapse_vertex)
                 machine_graph.add_edge(
                     neuron_to_synapse_edge, SPIKE_PARTITION_ID)
-                synapse_vertex.set_neuron_to_synapse_edge(
-                    neuron_to_synapse_edge)
+                synapse_vertex.set_neuron_vertex_and_partition_id(
+                    neuron_vertex, SPIKE_PARTITION_ID)
             else:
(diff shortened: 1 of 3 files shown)


spynnaker/pyNN/extra_algorithms/splitter_components/splitter_abstract_pop_vertex_neurons_synapses.py
```diff
@@ -483,9 +483,16 @@ def __add_plastic_feedback(
...
+            if (app_vertex.self_projection is None):
+                neuron_to_synapse_edge = MachineEdge(
+                    neuron_vertex, synapse_vertex)
+                machine_graph.add_edge(
+                    neuron_to_synapse_edge, SPIKE_PARTITION_ID)
+                synapse_vertex.set_neuron_to_synapse_edge(
+                    neuron_to_synapse_edge)
+            else:
+                synapse_vertex.set_neuron_vertex_and_partition_id(
+                    neuron_vertex, SPIKE_PARTITION_ID)
```
rowleya, line 495: This is fine, but note that in both cases you could use this call since you always know the neuron vertex and SPIKE_PARTITION_ID.  That would then save an "if" in the destination.

rowleya (approved):

andrewgait:
> This change looks fine - happy for this to be merged if ready

I think I'd prefer to leave this on a branch for now while I investigate the neuromodulation aspect a bit further - that's still not behaving the same when a splitter is used compared to when it isn't.

PR #1198, opened 2022-06-29
Change the PyNN URL to new home

rowleya:


Christian-B (approved):
ok to merge if you feel tested enough.

or easy to revert if is is not better

PR #1209, opened 2022-08-04
Changes necessary for the new spalloc

dkfellows:
See SpiNNakerManchester/SpiNNFrontEndCommon#842


dkfellows:
I'm not convinced that the test that this tests really ought to be in this repo at all. FEC ought to be what tests ASB...

dkfellows:
If #1210 goes in, this PR becomes valueless.

PR #1210, opened 2022-08-04
Delete a duplicate test

dkfellows:
Pretty much the same test file (Same name! Almost identical content, including in all useful senses as far as I can see.) exists in FEC, and there it appears to be actually in the same repository as the code it is testing. In sPyNNaker it's just weird.

If it used the sPyNNaker-specific subclass of `AbstractSpinnakerBase`, the test might be doing something useful. But it isn't.


andrewgait (approved):
Agreed, this should be (and already is) a test in FEC.

PR #1223, opened 2022-09-28
Single key_and_mask

Christian-B:
Part of https://github.com/SpiNNakerManchester/PACMAN/pull/472


spynnaker/pyNN/utilities/bit_field_utilities.py
```diff
@@ -227,7 +227,7 @@ def write_bitfield_init_data(
     for source_vertex, n_atoms, n_atoms_per_core in sources:
         r_info = routing_info.get_routing_info_from_pre_vertex(
             source_vertex, SPIKE_PARTITION_ID)
-        spec.write_value(r_info.first_key)
+        spec.write_value(r_info.first)
```
rowleya, line 230: Should that one have been key?

rowleya (changes requested):
I think there is an oops

spynnaker/pyNN/utilities/bit_field_utilities.py
```diff
@@ -229,7 +229,7 @@ def write_bitfield_init_data(
...
```
rowleya, line 232:
```suggestion
        spec.write_value(r_info.key)
```

PR #1226, opened 2022-09-29
Iterate projections

rowleya:
This avoids reusing incoming_projections to make sure that iterations over projections actually happen in all cases.

Tested by SpiNNakerManchester/IntegrationTests#148


Christian-B (approved):
Approved.

PR #1196, opened 2022-06-27
Reduce overhead data spec

rowleya:
Once #1195 is merged, this can follow.

Updates the way data specification is done to make things faster, using repeated data specification from various parts and caching it.

PR #1233, opened 2022-10-20
Count machine entries as well as application entries

andrewgait:
In certain scenarios with large enough source populations and multiple projections between this source and a single target, the master pop table size is being underestimated (this was found when investigating particular parts of the cerebellum network).  This PR is a temporary sticking plaster and whether this calculation is correct or not should be considered again when e.g. https://github.com/SpiNNakerManchester/sPyNNaker/pull/1196 is reviewed or merged.

Tested by https://github.com/SpiNNakerManchester/IntegrationTests/pull/156


andrewgait:
The problem this PR was solving was also fixed by https://github.com/SpiNNakerManchester/sPyNNaker/pull/1196 so this fix is no longer needed.

PR #1241, opened 2022-10-27
Fix SDRAM calculation on neurons-synapses splitter, add test

andrewgait:
SDRAM bug found when trying to use splitters with the cerebellum network as it currently stands.

The test included here is a simplified version of the problem found which (on current master) results in the exception:
`
Exception: Too much SDRAM has been allocated on chip 0, 0:  129711198 of 123469792`

(Note: the cerebellum network currently uses FromList connectors; I've used an AllToAll here to simplify the script so it runs faster than having to load in a cerebellum-like FromList).

Tested by https://github.com/SpiNNakerManchester/sPyNNaker/pull/1241


Christian-B (approved):
ok to merge

PR #1258, opened 2022-11-28
Fixes pool dense connector in many ways

rowleya:
This fixes various issues with PoolDenseConnector / PoolDense synapse type.

Also sneaks in ability to use SpikeInjector in 2D...


andrewgait (approved):
Looks fine to me - are there any integration tests we can add here based on the user code that needed these changes?

rowleya:
I was considering how to test this, but then realising that it needs an external device to test the main change that I think it fixed, which was to work correctly with external devices.  There is another change that lets it handle multiple incoming connections with the same kernel better, but this is more of an efficiency change than anything else...

PR #1259, opened 2022-11-29
test with python 3.11

Christian-B:
Adds python 3.11 to the tested versions.

once approved this should be added to the protection rules.

Also done in other repositories.


andrewgait (approved):

Christian-B:
different pr with tested branch name

PR #1264, opened 2022-12-01
Specify to use accum

andrewgait:
There were missing k's to specify accums in the sEMD synapse type.  It also seems that in places it's better to use ZERO rather than 0, so for consistency I've edited all the places I could find where this was needed.


rowleya (approved):
Yes, accum = int is bad enough, accum comparison with float is awful!

PR #1268, opened 2022-12-22
Tidy up code with warnings in gcc v12.2

andrewgait:
Building code with -Werror just to catch some (recent) warnings that had crept in, some are only found by gcc v12 but others existed in gcc v9.  One warning found further down the toolchain which I suggest we don't (yet) fix is described at https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/1013.

[I suggest we probably leave -Werror out of the default (user) build in automatic_make as we don't necessarily want to force people to use a particular version of gcc, though perhaps it might be a nice idea to have it somewhere in a build process that gets used in CI?]

Will test this branch once the big machine is back up and running again; for now scripts related to where changes were made seem to be working locally at least.

### DIFF
--- a/neural_modelling/src/neuron/c_main_neurons.c
+++ b/neural_modelling/src/neuron/c_main_neurons.c
@@ -122,7 +122,7 @@ static uint32_t recording_flags = 0;
 static struct sdram_config sdram_inputs;
 
 //! The inputs from the various synapse cores
-static weight_t *synaptic_contributions[N_SYNAPTIC_BUFFERS];
+static weight_t **synaptic_contributions;
 
 //! The timer overruns
 static uint32_t timer_overruns = 0;
@@ -271,6 +271,7 @@ static bool initialise(void) {
     spin1_memcpy(&sdram_inputs, sdram_config, sizeof(struct sdram_config));
 
     uint32_t n_words = sdram_inputs.size_in_bytes >> 2;
+    synaptic_contributions = spin1_malloc(N_SYNAPTIC_BUFFERS * sizeof(uint32_t*));
     for (uint32_t i = 0; i < N_SYNAPTIC_BUFFERS; i++) {
         synaptic_contributions[i] = spin1_malloc(sdram_inputs.size_in_bytes);
         if (synaptic_contributions == NULL) {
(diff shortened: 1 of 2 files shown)


neural_modelling/src/neuron/c_main_neurons.c
```diff
@@ -122,7 +122,7 @@ static uint32_t recording_flags = 0;
 static struct sdram_config sdram_inputs;
 
 //! The inputs from the various synapse cores
-static weight_t *synaptic_contributions[N_SYNAPTIC_BUFFERS];
+static weight_t **synaptic_contributions;
```
rowleya, line 125: I would be interested to know what the warning is on that one...

rowleya:
If we want a "warnings as errors" version in the testing, we can always add an environment variable.  It might be that adding -DCFLAGS=-werror would already work but I am not certain...

neural_modelling/src/neuron/c_main_neurons.c
```diff
@@ -271,6 +271,7 @@ static bool initialise(void) {
     spin1_memcpy(&sdram_inputs, sdram_config, sizeof(struct sdram_config));
 
     uint32_t n_words = sdram_inputs.size_in_bytes >> 2;
+    synaptic_contributions = spin1_malloc(N_SYNAPTIC_BUFFERS * sizeof(uint32_t*));
     for (uint32_t i = 0; i < N_SYNAPTIC_BUFFERS; i++) {
         synaptic_contributions[i] = spin1_malloc(sdram_inputs.size_in_bytes);
         if (synaptic_contributions == NULL) {
```
rowleya, line 277: The real error the warning is finding is here; it should be checking:
if (synaptic_contributions[i] == NULL) {

andrewgait:
I think we'd have to add to the CFLAGS variable to make warnings-as-errors work in testing; the other issue then would be that eventually we'd run into https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/issues/1013.

andrewgait:
Although, we could possibly just add to CFLAGS when building the sPyNNaker C code and that would avoid that issue.

rowleya:
Yes, until the bugs are removed we might have to wait (or at least disable some warnings I guess)

rowleya:
Note: it is possible to do -Wall -Wno-array-bounds so we don't do the check that is in error I believe...

PR #1284, opened 2023-02-02
Use more accurate calculation of core_mask

andrewgait:
Fix the core_mask calculation to work when the machine_mask is too large for the master pop table.


rowleya (approved):
Approved once tests pass!

andrewgait:
> Approved once tests pass!

Still some failing tests so don't merge yet...

PR #1295, opened 2023-02-23
Auto excludes

Christian-B:
part of https://github.com/SpiNNakerManchester/sphinx8/issues/7


Christian-B:
Now works after a work around for strange readthedoc behaviour where the number of spaces in rst files is different here than in the global version

PR #1296, opened 2023-02-28
Fix pushbot

rowleya:
Various fixes to make the pushbot work again (and a bit more reliable)


andrewgait (approved):
Looks good to me.  Difficult to add tests without a real pushbot being present to run them on...

PR #1297, opened 2023-02-28
Fix pool dense odd

rowleya:
Fixing another oddity in this connector, along with a test


andrewgait (approved):
Looks fine to me too, good to have a test as well.

PR #1300, opened 2023-03-07
Local only delays

rowleya:
This allows delays to be used with local-only connectors.

Some side affects include:
 - All cores can now calculate how much delay they can manage and adjust the number of ring buffer entries appropriately
 - Delay extensions are added based on what each core can do
 - This together means that e.g. a core with a single neuron with two synaptic types might be able to do 512 delay slots without a delay extension.

Testing with: http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/local_only_delays/2/pipeline


unittests/model_tests/neuron/test_synaptic_manager.py
```diff
@@ -406,23 +406,21 @@ def test_set_synapse_dynamics():
         # Only undelayed, all edges exist
         (range(10), [], 1000, 100, None),
         # Only delayed, all edges exist
-        ([], range(10), 1000, 100, 20),
+        ([], range(10), 1000, 100, 200),
```
andrewgait, line 409: Increasing this so it's large enough to now allow a delay extension is fine; I wonder if there are any integration tests with large enough delays so that delay extensions are also triggered there.

spynnaker/pyNN/extra_algorithms/splitter_components/abstract_spynnaker_splitter_delay.py
```diff
@@ -26,17 +26,14 @@ class AbstractSpynnakerSplitterDelay(object, metaclass=AbstractBase):
 
     __slots__ = []
 
-    # max delays supported by a slice split machine vertex
-    MAX_SUPPORTED_DELAY_TICS = 16
-
```
andrewgait, line 31: This simplifies the issues I've been having on other branches somewhat so I'm happy to see it go in.  I wonder what the effect of this is in general on networks which previously used delay extensions, and whether it means that we might need some further integration testing such that delay extensions are still tested properly?

andrewgait (approved):
I'm happy for this to go in providing tests pass but have made a few comments about how this makes delay extensions (much) less prevalent than currently.

unittests/model_tests/neuron/test_synaptic_manager.py
rowleya: I know there is one that I added, but I will check the others!

spynnaker/pyNN/extra_algorithms/splitter_components/abstract_spynnaker_splitter_delay.py
rowleya: Yes worth a check in general

andrewgait: Looks like a few tests are failing because of some assumptions about delay extensions so it's worth looking into them.

unittests/model_tests/neuron/test_synaptic_manager.py
rowleya: There are definitely some delay extension tests now, including one that tests various stages, so I am reasonably happy with this.

spynnaker/pyNN/extra_algorithms/splitter_components/abstract_spynnaker_splitter_delay.py
rowleya: These are passing now.  The issue was more about clearing the delay status in the splitter.

Issue #1249, opened 2022-11-09
get_data clear does not clear previous segments


Christian-B:
Population.get_data(...Clear=True) removes data from the current sqlite3 database 

It however does not affect either
- previous sqlite3 databases no longer used after a reset (in current master only hard ones)
- data held in memory from previous segments

Christian-B:
To replicate:
import pyNN.spiNNaker as sim

sim.setup(timestep=1.0)
sim.set_number_of_neurons_per_core(sim.IF_curr_exp, 100)

pop_1 = sim.Population(1, sim.IF_curr_exp(), label="pop_1")
input = sim.Population(1, sim.SpikeSourceArray(spike_times=[0]), label="input")
input_proj = sim.Projection(input, pop_1, sim.OneToOneConnector(),
                            synapse_type=sim.StaticSynapse(weight=5, delay=1))
pop_1.record(["spikes", "v"])
simtime = 10
sim.run(simtime)
sim.reset()
sim.run(simtime)
neo = pop_1.get_data(variables=["spikes", "v"], clear=True)
spikes = neo.segments[0].spiketrains
print(spikes)
spikes1 = neo.segments[1].spiketrains
print(spikes1)
v = neo.segments[0].filter(name='v')[0]
v1 = neo.segments[1].filter(name='v')[0]
print(v)
print(v1)
sim.run(simtime)

neo = pop_1.get_data(variables=["spikes", "v"])
spikes = neo.segments[0].spiketrains
print(spikes)
spikes1 = neo.segments[1].spiketrains
print(spikes1)
v = neo.segments[0].filter(name='v')[0]
v1 = neo.segments[1].filter(name='v')[0]
print(v)
print(v1)
sim.end()

PR #1305, opened 2023-03-10
fix getdata clear on previous segment Clear

Christian-B:
For the record I hate get_data clear!
I would rather log warn that we ignore the setting!
I would rather support the ability to get a time subset of data out of the database

Well if we must this now support get_data clear =True on a previous segment
Fixes https://github.com/SpiNNakerManchester/sPyNNaker/issues/1249

also a quick change to support
pop.record["all"])


rowleya (approved):

Christian-B:
tested by http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/clear/2/pipeline

PR #1309, opened 2023-03-17
test small world split no random

Christian-B:
This shows the small world connector is BROKEN if the vertex is split.

The tests remove all randomness/ rewiring.


Christian-B:
Intentionally not fixing the pylint issues so it fails actions and can not be merged as the many core test fails!

Christian-B:
Still showing connectors broken

Christian-B:
Assume this is no longer needed

rowleya:
Believed now fixed

PR #1314, opened 2023-03-28
Neaten the output by providing the actual ports to be used

rowleya:


andrewgait (approved):
This was the way I'd fixed this test as well.

PR #1322, opened 2023-04-13
Enable spell checking of docstrings

dkfellows:
See:
* https://github.com/SpiNNakerManchester/SupportScripts/pull/54
* https://github.com/SpiNNakerManchester/SpiNNUtils/pull/225


dkfellows:
I think I'll leave these last four things unhandled. I don't know if we want to rename `stdev` to `stddev` and the other three are (probably) correct-ish but I'm not sure what to do about them.

PR #1330, opened 2023-04-24
Avoid list searches

rowleya:
Speed up runs by avoiding the searching of lists.

NOTE: This removes some safety checks


rowleya:
Done elsewhere

Issue #1323, opened 2023-04-17
Allow python lists/arrays in KernelConnector


andrewgait:
Currently the arrays passed in to a KernelConnector need to be converted into a numpy array by the user first (see https://github.com/SpiNNakerManchester/sPyNNaker/blob/e2c3d66ca2831e2e4ef239eca07d00cc88a17c35/spynnaker/pyNN/models/neural_projections/connectors/kernel_connector.py#L229).  This seems unnecessary so I suggest it gets fixed so that an array (of arrays) can be passed in as well.

rowleya:
Yes, I think this was what was done some time ago; normal arrays should work I would guess (or they can be converted by the constructor if that is easier)

PR #1336, opened 2023-05-02
Allow lists to be sent into kernel connector, update test

andrewgait:
Fixes https://github.com/SpiNNakerManchester/sPyNNaker/issues/1323

### DIFF
--- a/spynnaker/pyNN/models/neural_projections/connectors/kernel_connector.py
+++ b/spynnaker/pyNN/models/neural_projections/connectors/kernel_connector.py
@@ -243,9 +243,9 @@ class KernelConnector(AbstractGenerateConnectorOnMachine,
                 if correct_size:
                     return numpy.asarray(vals).view(ConvolutionKernel)
         raise SpynnakerException(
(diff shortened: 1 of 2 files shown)


spynnaker/pyNN/models/neural_projections/connectors/kernel_connector.py
```diff
@@ -232,11 +232,21 @@ def __get_kernel_vals(self, vals):
...
+        elif ((isinstance(vals, list))):
+            height = len(vals)
+            if height == self._kernel_h:
+                correct_size = True
+                for h in range(height):
+                    width = len(vals[h])
+                    if width != self._kernel_w:
+                        correct_size = False
+                if correct_size:
+                    return numpy.asarray(vals).view(ConvolutionKernel)
```
rowleya, line 244: Would it be easier to add a check for a list in advance of this and then do vals = numpy.asarray(vals) there?  Then the other checks would happen automatically?

andrewgait: Agreed - done.

Issue #1337, opened 2023-05-03
STDP not working after a (soft reset)


Christian-B:
To replicate:
edit IntroLab/learning/struct_pl_stdp.py
replace
sim.run(simtime)
with
sim.run(simtime)
sim.reset()
sim.run(simtime)

Error:
Spinnman.exceptions.SpiNNManCoresNotInStateException: waiting for cores ['0, 0, 4 (ph: 3)', '0, 0, 7 (ph: 6)'] to reach one of [<CPUState.PAUSED: 10>, <CPUState.READY: 5>]
On chip_runtime_updater
004 and 007 are the SpikeSourcePoisson cores.

Christian-B:
Forcing the reset to be hard by preceding with it with a 
m= sim.get_machine()
resolves this.

Christian-B:
Suggestion is to remove the requires Data generation and convert its use to requires remapping

Christian-B:
All the scripts in https://github.com/SpiNNakerManchester/sPyNNaker/tree/master/spynnaker_integration_tests/test_stdp****

work wiitha reset

Christian-B:
Fixed by https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1059 and https://github.com/SpiNNakerManchester/sPyNNaker/pull/1339

PR #1338, opened 2023-05-03
removes requires_data_generation

Christian-B:
Fixes: https://github.com/SpiNNakerManchester/sPyNNaker/issues/1337

The only know place that requires_data_generation instead of requires_mapping was used was 
self.__synapse_dynamics.changes_during_run

As test show that this use case actually fails the simplest solution is just to make that a requires_mapping.

This change does come at a time cost the runs after reset.
As it will have to rerun the mapping and get machine steps.

Based on timing done on the microcircuit this extra time would add about 50% to the preparation time of the run after reset.
Ref: http://apollo.cs.man.ac.uk:8080/blue/organizations/jenkins/Integration%20Tests/detail/time_report2/27/pipeline/99
Reports stage

If this PR is refused another solution for https://github.com/SpiNNakerManchester/sPyNNaker/issues/1337 will need to be found. And even if fixed this time there is a danger another error could slip in this rare use case.

This allows the complete removal of the requires_data_generation:
https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1058
https://github.com/SpiNNakerManchester/SpiNNUtils/pull/230

tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/198

### DIFF
--- a/spynnaker/pyNN/models/neuron/abstract_population_vertex.py
+++ b/spynnaker/pyNN/models/neuron/abstract_population_vertex.py
@@ -927,7 +927,7 @@ class AbstractPopulationVertex(
         # If synapses change during the run also regenerate these to get
         # back to the initial state
         if self.__synapse_dynamics.changes_during_run:
-            SpynnakerDataView.set_requires_data_generation()
+            SpynnakerDataView.set_requires_mapping()
         else:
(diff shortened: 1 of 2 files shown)


Christian-B:
Did not work as expected.

PR #1340, opened 2023-05-10
retouter reports for virtual vertices

Christian-B:
Test for https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1060 which fail without it.


Christian-B:
just tests so merging

PR #1345, opened 2023-05-18
Interdimensional compatibility

rowleya:
Makes it possible for vertices with different dimensions to communicate correctly.  This means that any connectors can be used between vertices and they should still work.

Depends on:
SpiNNakerManchester/PACMAN#516
SpiNNakerManchester/SpiNNFrontEndCommon#1064
SpiNNakerManchester/PyNN8Examples#104


unittests/mocks.py
```diff
@@ -34,3 +37,13 @@ def label(self):
 
     def __repr__(self):
         return "Population {}".format(self._label)
+
+    @property
+    def _vertex(self):
+        return self._mock_vertex
+
+
+class MockVertex(object):
```
Christian-B, line 46: Why not use pacman_test_objects/simple_test_vertex.py and extend that with the extra method

optional!

unittests/model_tests/neuron/test_synaptic_manager.py
```diff
@@ -113,13 +113,14 @@ def test_write_data_spec():
         additional_parameters={
             "splitter": SplitterAbstractPopulationVertexFixed()})
     proj_one_to_one_1 = p.Projection(
-        pre_pop, post_pop, p.OneToOneConnector(),
+        pre_pop, post_pop, p.FromListConnector([(i, i) for i in range(10)]),
```
Christian-B, line 116: Why remove the check of a OneToOne and an AllToAll?

If wanted ADD a fromList or convert one of the OneToOne

unittests/test_using_virtual_board/test_from_file_connector.py
```diff
@@ -90,16 +90,6 @@ def test_simple(self):
         ]
         self.check_other_connect(as_list)
 
-    def test_list_too_big(self):
```
Christian-B, line 93: Why remove this test?

If it no longer works check the pop

unittests/test_using_virtual_board/test_from_list_connector.py
```diff
@@ -78,16 +78,6 @@ def test_simple(self):
         ]
         self.check_other_connect(as_list)
 
-    def test_list_too_big(self):
-        as_list = [
```
Christian-B, line 82: Either it work and the test stays or it does not and we check for a good error

spynnaker/pyNN/models/neuron/population_machine_neurons.py
```diff
@@ -335,10 +319,16 @@ def _write_current_source_parameters(
         spec.switch_write_focus(self._neuron_regions.current_source_params)
 
         # Get the current sources from the app vertex
+        app_current_sources = self._pop_vertex.current_sources
         current_source_id_list = self._pop_vertex.current_source_id_list
 
         # Work out which current sources are on this core
-        current_sources = self.__get_current_sources_sorted()
+        current_sources = set()
```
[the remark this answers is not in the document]
rowleya: Changed to OrderedSet

spynnaker/pyNN/models/projection.py
```diff
@@ -525,3 +528,7 @@ def size(self, gather=True):  # @UnusedVariable
         """
         # TODO
         _we_dont_do_this_now()
+
```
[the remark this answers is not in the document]
rowleya: Removed and fixed the test

spynnaker_integration_tests/test_various/test_synaptic_expander.py
```diff
@@ -48,18 +48,15 @@ def run_script():
     connectors = [
         (p.OneToOneConnector, functools.partial(check_one_to_one, 10)),
         (p.AllToAllConnector,
-         functools.partial(check_all_to_all, 10, True)),
-        (functools.partial(p.AllToAllConnector,
```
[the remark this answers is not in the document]
rowleya: This test is done elsewhere now as the current test doesn't do self-connection (so the test fails its own check)

unittests/connector_tests/test_csa_connector.py
```diff
@@ -1,111 +0,0 @@
-﻿# Copyright (c) 2017 The University of Manchester
```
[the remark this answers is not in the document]
rowleya: This is done as an integration test now, as it wasn't able to pass as a unit test.

unittests/mocks.py
rowleya: Could have... but then I was changing sPyNNaker here!

unittests/model_tests/neuron/test_synaptic_manager.py
rowleya: Undone changes - must have been testing something here!

unittests/test_using_virtual_board/test_from_file_connector.py
rowleya: Restored - must have removed thinking it couldn't work in unit test, but actually found issues!

unittests/test_using_virtual_board/test_from_list_connector.py
rowleya: Restored as above!

spynnaker/pyNN/models/neuron/population_machine_neurons.py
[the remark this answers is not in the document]
Christian-B: Actually changed to list (by sorted method) which is the correct fix

Christian-B (approved):
Please link all related branches so we don't miss one

PR #1351, opened 2023-05-30
Uncap numpy

Christian-B:


spynnaker/pyNN/models/neural_projections/connectors/from_list_connector.py
```diff
@@ -147,7 +147,7 @@ def _split_connections(self, post_slices):
 
         # Split the sort order in to groups of connection indices
         split_indices = numpy.array(numpy.split(
-            sort_indices, numpy.cumsum(index_count)))[:-1]
+            sort_indices, numpy.cumsum(index_count)), dtype=object)[:-1]
```
dkfellows, line 150: What does this do?

dkfellows (approved):

spynnaker/pyNN/models/neural_projections/connectors/from_list_connector.py
Christian-B: the split is a [array([0, 1, 2, 3, 4]), array([], dtype=int64)] and in numpy 1.23 that was ok for array and in 1.24 numpy changed the rules

Now it gives:
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.

the fix was suggested by @rowleya

PR #1354, opened 2023-06-14
Izk cond exp dual

andrewgait:
Build Izhikevich dual-excitatory synapse neuron models by default.

Tested by https://github.com/SpiNNakerManchester/IntegrationTests/pull/208


andrewgait:
Interesting - the C actions here lead to a successful build whereas on Jenkins (and probably on Jupyter as well) there's an ITCM overflow, despite apparently the same gcc version in each case.  I guess we ought to update the ubuntu/gcc version on Jenkins and Jupyter... ?

PR #1356, opened 2023-06-15
Rng warnings

rowleya:
Ensures that if RNG is used in the wrong place, the user knows about it.  This comes as exceptions by default, with an option to turn it into a warning if wanted.

Note that the exception/warning is circumstantial: if the exact parameters mean that things will be generated on machine, but the user has supplied a seeded RNG to places it will be ignored, they are told about it.

All locations in the current code have been fixed in tests and examples.  The following have also been changed to match:
SpiNNakerManchester/PyNN8Examples#105
SpiNNakerManchester/IntroLab#47
SpiNNakerManchester/microcircuit_model#33

Tested with:
SpiNNakerManchester/IntegrationTests#210


spynnaker/pyNN/models/neuron/abstract_population_vertex.py
```diff
@@ -104,6 +108,40 @@ def _prod(iterable):
     return reduce(operator.mul, iterable, 1)
 
 
+def _all_gen(rd):
```
andrewgait, line 111: Both this and the following function seem like they should be somewhere more general (utility_calls perhaps?) than in the population vertex.  I don't mind them being here for now but perhaps consider moving them somewhere else in the future?

andrewgait (approved):
Approved with a comment that you can take or leave...

spynnaker/pyNN/models/neuron/abstract_population_vertex.py
rowleya: Yes, could be moved if needed in future; currently they are only ever used here...

PR #1367, opened 2023-07-05
Fix poisson multirun error

andrewgait:
The included test fails on the master branch because the region size calculation doesn't take into consideration that the set rate values may change from being the same to being different on a subsequent call to run().  This branch fixes the issue by assuming all the rates are different, though there may be a better way of doing it... ?

Tested by https://github.com/SpiNNakerManchester/IntegrationTests/pull/216

### DIFF
--- a/spynnaker/pyNN/models/spike_source/spike_source_poisson_machine_vertex.py
+++ b/spynnaker/pyNN/models/spike_source/spike_source_poisson_machine_vertex.py
@@ -411,6 +411,9 @@ class SpikeSourcePoissonMachineVertex(
         data_items = list()
         data_items.append([int(self.__rate_changed)])
         data_items.append([0])
+        # In order to allow changes in rate between runs track the data size
+        # for the case where all rates are different and use that
+        data_size = 2
         n_items = 0
         data = self._app_vertex.data
         ids = self.vertex_slice.get_raster_ids()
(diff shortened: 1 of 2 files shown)


Christian-B (commented):
I dont understand how you are changing the size of the region to reserve on the python side with no change on the c side which uses this region.

rowleya:
The issue being fixed here is that the Poisson code contains code to generate Poisson parameters when it is loaded or reloaded.  If the region it reads to do this is sized for only the initial parameters (which might include a repeat statement if all the rates are the same), then if they are changed later to something less uniform (so more repeat statements with less repeats per statement), this won't fit in the initial sized regeon.  The solution here is to size the region bigger in the first place so it is big enough for one repeat statement per neuron, in case that happens later.

Downside is that this reserves more SDRAM than needed in most cases.  That isn't critical since SDRAM isn't that limited and the amount here isn't huge.  Upside is that it works in all situations!

Christian-B:
> The issue being fixed here is that the Poisson code contains code to generate Poisson parameters when it is loaded or reloaded. If the region it reads to do this is sized for only the initial parameters (which might include a repeat statement if all the rates are the same), then if they are changed later to something less uniform (so more repeat statements with less repeats per statement), this won't fit in the initial sized regeon. The solution here is to size the region bigger in the first place so it is big enough for one repeat statement per neuron, in case that happens later.
> 
> Downside is that this reserves more SDRAM than needed in most cases. That isn't critical since SDRAM isn't that limited and the amount here isn't huge. Upside is that it works in all situations!

While it is true a bigger region is reserved I see no change to the reported amount of sdram that the Vertex is setup to use which is then reported by sdram_required method.
So I hope the Vertex already claimed the larger sdram even if it did not actively use it, 

In which case there is no downside. 
IE this change will not reduce the number of steps in an auto_pause_resume loop which is what using more sdram would do.

somewhere there was a calculation of how much SDRAM this reserve could take.
Why not reuse that number here rather than recalculate it.

spynnaker/pyNN/models/spike_source/spike_source_poisson_machine_vertex.py
```diff
@@ -421,14 +424,16 @@ def _write_poisson_rates(self, spec):
                  _u3232_to_uint64(item['starts']),
                  _u3232_to_uint64(item['durations']))
                 )[0]
-            data_items.extend([[count], [len(items)], [0],
-                               numpy.ravel(items).view("uint32")])
+            data_to_add = [[count], [len(items)], [0],
+                           numpy.ravel(items).view("uint32")]
+            data_items.extend(data_to_add)
             n_items += 1
+            data_size += count * (3 + len(data_to_add[-1]))
```
rowleya, line 431: Note this could simply use get_expander_rates_bytes, which would then match up with the pre-computed value

rowleya (changes requested):

rowleya:
> somewhere there was a calculation of how much SDRAM this reserve could take. Why not reuse that number here rather than recalculate it.

I have asked for this change - as you say, using the calculated maximum size used for placement is a good idea.

spynnaker/pyNN/models/spike_source/spike_source_poisson_machine_vertex.py
andrewgait: Done and fixed the calculation of get_expander_rates_bytes at the same time.

rowleya (approved):

PR #1390, opened 2023-09-29
Temp fix for Pynn 0.12.0

Christian-B:
Do not merge but copy to test other branches.

Revert with  PyNN 0.12.1 which is being prepared


Christian-B:
better fix is https://github.com/SpiNNakerManchester/sPyNNaker/pull/1391.

Do not delete as this branch is testing PyNN 0.12 assuming the next version fixes the dependency

PR #1393, opened 2023-10-02
A idea to keep databases in view

Christian-B:
part of https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1122


Christian-B:
Did not work as hoped due to Various objects and threads using the same sqlite3 file

PR #1400, opened 2023-10-10
Fix issue with neuron generation global parameters

rowleya:
This only shows up when you have a single global parameter (or maybe more global that not).


Christian-B (approved):
agreed

PR #1458, opened 2024-06-20
Manual weight scaling

rowleya:
Add option to allow manual scaling of weights by users.  This then disables the automatic scaling of weights for that Population.  This can result in more weight overflows but that is given as a warning if it happens already.

Example use:
```
pop_fixed_1 = p.Population(256, p.IF_curr_exp(), max_expected_summed_weight=[2.0, 0.0])
```

Values are in units of the synapses (nA for current-based synapese, uS for conductance), and represent the maximum representable sum of the weights incoming to a synapse of a neuron in any single time step.  One value must be specified for each synapse type of the neuron (e.g. for IF_curr_exp, one value each for excitatory and inhibitory).


Christian-B (changes requested):
A big fat no to adding a param middle list.

Use a "*," to force callers to use named params

PR #1463, opened 2024-07-11
Remove bitfeilds

Christian-B:
stopped creating the ds bit_field_region

disabled the code that uses that data


Christian-B:
did not work

PR #1474, opened 2024-08-01
Region downloading

rowleya:
Add changes to allow fast download of synapses, and test.

Depends on:
 - https://github.com/SpiNNakerManchester/PACMAN/pull/570
 - SpiNNakerManchester/SpiNNFrontEndCommon#1202
 - SpiNNakerManchester/JavaSpiNNaker#1165


Christian-B (changes requested):
In principle I approve these changes.

What worries me is that to trigger recording there is a script level addition to trigger the "download_synapses".  This is less than ideal but I can not think of a clean solution to trigger this for only a few projections.

However I can not find a similar level call to get the data,

I see it in the data.sqlite3 file but there is not way to get it back to script level.

Which means the test only ask for it but never check it was recorded read back let alone correct

Christian-B:
Fails if you download_synapses on two (or more) projections into the same target vertex

see:
https://github.com/SpiNNakerManchester/sPyNNaker/blob/region_downloading/spynnaker_integration_tests/test_various/test_double_download_synapses.py

spynnaker/pyNN/models/neuron/population_machine_synapses.py
```diff
@@ -266,3 +268,12 @@ def max_gen_data(self) -> int:
...
     def bit_field_size(self) -> int:
         return self._synaptic_matrices.bit_field_size
+
+    @overrides(AbstractReceiveRegionsToHost.get_download_regions)
+    def get_download_regions(
+            self, placement: Placement) -> Sequence[Tuple[int, int, int]]:
+
+        # Make sure not to overwrite the recording regions
+        start_index = len(self._pop_vertex.get_recording_variables())
+        return self._synaptic_matrices.get_download_regions(
```
Christian-B, line 278: The local_recording_index for recording variables is fixed regardless which ones are actually recordings.
For example V is always 0 and Spikes always 3

So you can not use len of recording variables!

Christian-B: fixed by using get_recordable_variables

Christian-B:
I see that  SynapticMatrixApp. get_connections (called indirectly from Projection.get) tries to read back the data saved.

This works fine if the simulation is just run once.
But then I am not sure what the benefit is other than the raw data is available in the database.

However this fails if the simulation is run multiple times.

    sim.setup(timestep=1.0, n_boards_required=1)
    pop_1 = sim.Population(1, sim.IF_curr_exp(), label="pop_1")
    input_pop = sim.Population(
        1, sim.SpikeSourceArray(spike_times=[0]), label="input")
    input_proj = sim.Projection(input_pop, pop_1, sim.OneToOneConnector(),
                                synapse_type=sim.StaticSynapse(weight=5, delay=1)
                                , download_synapses=True
                                )
    pop_1.record(["spikes", "v"])
    simtime = 10
    sim.run(simtime)
    sim.run(simtime)

con = input_proj.get(["weight","delay"], "list")
print(con)

---
print  is
([0, 0, 5.0, 1.0], [1, 0, 5.0, 1.0])

--
The reason is the bufferExtractor is run at the end of each run (or auto pause resume loop)
The data for the region is added each time.
(In Database terms it is addred to the RegionExtra) table

So in the above example the block of data passed to convert_to_connections is twice as log as expected and this confused the function.

rowleya:
To answer some questions:
 - You are right that you have to specify the "download_synapses" option to make this work.  This is because I didn't want to pre-download all synapses just in case they were wanted; most simulations don't want to read the synapses back, so this would be wasteful.  Instead, this flag is used to indicate synapses that are wanted, and so these are downloaded using the "fast data read" protocol.  Note that reading back synapses without the flag should still work, it is just that they will be "slow" downloaded.  
 - The downloading could be altered so metadata is used once and the downloading actually happens when the call is made rather than pre-emptively.  This could then be used for recording downloading too.
 - You are right that mutliple runs will fail because things get extended.  This will either need changing as specified above (one off metadata and clear) or else the region needs to be cleaned immediately once read or once simulation restarts (probably easier but not sure).

PR #1475, opened 2024-08-06
Fix poisson types

rowleya:
This fixes a few things mainly with the types of Poisson variables but also with a circular buffer "out-by-one" when the number of sources was a power of 2.


rowleya:
Turns out this has been done elsewhere, but better (except for one change here)

PR #1476, opened 2024-08-06
Fix poisson

rowleya:
Fixing types of the Poisson spike source


Christian-B (approved):
So of the test changes appear unnecessary but are not wrong.

PR #1480, opened 2024-08-28
Dns hack

Christian-B:
hack/ work around for https://github.com/SpiNNakerManchester/SpiNNMan/issues/414

dns 
linux 
- set by https://github.com/SpiNNakerManchester/SupportScripts/pull/78

windows 
- not yet failed so no change

macos 
 -fails 
- no dns setting hack found yet
- seperated macos test to separate checks

----
Also simplifies the test run.
One run of simple in debug mode vs several.

Later prs could add extra tests if needed


Christian-B:
Note: This PR is Green by fluke!

Will need https://github.com/SpiNNakerManchester/SupportScripts/pull/78 to work more often in linux

PR #1487, opened 2024-09-20
Wta connector

rowleya:
A new connector for sPyNNaker that can to "Winner takes all" connections.  That is, it can generate an all-to-all-except-same-index connection set, but also over multiple "groups", where a group is a subset of the neurons.  For example between 10 neurons with 2 groups, there will be a WTA connection set between neurons 0-4 and another WTA connection set between neurons 5-9, each of which are independent.  This is useful for the multi-TSP networks.

The WTA can also have weights for each group, where there have to be n_values * (n_values - 1) weights.  These are used in all groups if specified.


Christian-B (approved):

Christian-B (changes requested):
I would like to see an example/tests where the weights at least move towards a winner

PR #1489, opened 2024-09-25
a warning for default/None timestep

Christian-B:
Mainly to explain the message:
 WARNING: A timestep was entered that has forced SpiNNaker to automatically slow the simulation down from real time by a factor of 10.

### DIFF
--- a/spynnaker/pyNN/__init__.py
+++ b/spynnaker/pyNN/__init__.py
@@ -246,7 +246,7 @@ def distance(src_cell: IDMixin, tgt_cell: IDMixin,
         src_cell, tgt_cell, mask, scale_factor, offset, periodic_boundaries)
 
 
-def setup(timestep: Optional[float] = None,
+def setup(timestep: Optional[Union[float, Literal["auto"]]] = None,
           min_delay: Union[float, Literal["auto"]] = (
               _pynn_control.DEFAULT_MIN_DELAY),
           max_delay: Optional[Union[float, Literal["auto"]]] = None,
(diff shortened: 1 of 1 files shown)


spynnaker/pyNN/__init__.py
```diff
@@ -291,9 +291,15 @@ def setup(timestep: float = _pynn_control.DEFAULT_TIMESTEP,
...
     # pylint: disable=global-statement,too-many-arguments
+    # Check for "auto" values and None
     global __simulator
-    # Check for "auto" values
-    if timestep == "auto":
+    if timestep is None:
+        logger.warning(
+            f"The default PyNN timestep of {_pynn_control.DEFAULT_TIMESTEP} "
+            "is less than 1(ms) that Spynakker is designed for. "
+            "Consider including a timestep in you setup call.")
```
rowleya, line 300: "you" --> "your"

spynnaker/pyNN/__init__.py
```diff
@@ -246,7 +246,7 @@ def distance(src_cell: IDMixin, tgt_cell: IDMixin,
         src_cell, tgt_cell, mask, scale_factor, offset, periodic_boundaries)
 
 
-def setup(timestep: float = _pynn_control.DEFAULT_TIMESTEP,
+def setup(timestep: Optional[float] = None,
```
rowleya, line 249: Should this actually be Optional[Union[float, str]]?

rowleya: Or Optional[Union[float, Literal["auto"]]]?

spynnaker/pyNN/__init__.py
rowleya, line 300:
```suggestion
            f"The default PyNN timestep of {_pynn_control.DEFAULT_TIMESTEP} "
            "is less than 1(ms) that SpyNNaker is designed for. "
            "Consider including a timestep in your setup call.")
```

PR #1491, opened 2024-09-27
Add a shift connector

rowleya:
Yet another connector!  Hopefully better implemented and named this time... this one simply makes connections where post is shifted from pre by a fixed amount (with optional wrap and groups).  This will again be useful in the TSP, though I'm sure other people might also find a use!

### DIFF
--- a/spynnaker/pyNN/models/neural_projections/connectors/one_to_one_offset_connector.py
+++ b/spynnaker/pyNN/models/neural_projections/connectors/one_to_one_offset_connector.py
@@ -36,30 +36,31 @@ if TYPE_CHECKING:
         ProjectionApplicationEdge)
 
 
-class ShiftConnector(AbstractGenerateConnectorOnMachine,
-                     AbstractGenerateConnectorOnHost):
+class OneToOneOffsetConnector(
+        AbstractGenerateConnectorOnMachine,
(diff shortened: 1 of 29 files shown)


spynnaker/pyNN/models/neural_projections/connectors/one_to_one_offset_connector.py
```diff
@@ -0,0 +1,241 @@
...
+                # If there is a wrap, there will always be a next connection
+                return synapse_info.n_pre_neurons
+            # If there isn't a wrap, there are always offset less connections
+            return synapse_info.n_pre_neurons - abs(self.__offset)
+
+        if self.__wrap:
+            # If there is a wrap, there will always be a next connection
+            return self.__n_neurons_per_group
+        # If there isn't a wrap, there are always offset less
+        return self.__n_neurons_per_group - abs(self.__offset)
```
Christian-B, line 100: n_neurons = synapse_info.n_pre_neurons
        n_groups = n_neurons // self.__n_neurons_per_group 
        return n_neurons - n_groups * abs(self.__offset)

Christian-B (changes requested):
__n_connections incorrect in one case

otherwise approved.

PR #1496, opened 2024-10-04
Allow different partitions

rowleya:
Allow different partitions from the default SPIKE_PARTITION_ID to be used.  By default this on is used if not specified, and some models will only *send* with that partition id still, but this allows the reception of different partition ids at least, which makes it easier to integrate with other sending models.

Tested with SpiNNakerManchester/IntegrationTests#294 (failing only for unrelated reasons)

### DIFF
--- a/neural_modelling/src/neuron/plasticity/weight_change/synapse_dynamics_external_weight_change.c
+++ b/neural_modelling/src/neuron/plasticity/weight_change/synapse_dynamics_external_weight_change.c
@@ -160,6 +160,8 @@ static inline updatable_synapse_t process_plastic_synapse(
             } else {
                 synapse.weight = (weight_t) new_weight;
             }
+			log_info("Weight changed to %d for pre-neuron %u, post-neuron %u",
+					synapse.weight, pre_spike, s.index);
             log_debug("        Weight now %d", synapse.weight);
             *changed = 1;
 
@@ -198,7 +200,7 @@ static inline void process_weight_update(
     const uint32_t *words = (uint32_t *) synapse_row_plastic_controls(fixed_region);
     uint32_t pre_spike = plastic_region_address->pre_spike;
 
-	log_debug("Weight change update for pre-neuron %u", pre_spike);
+	log_info("Weight change update for pre-neuron %u", pre_spike);
 
     // Loop through synapses
     for (; n_synapses > 0; n_synapses--) {
@@ -210,7 +212,7 @@ static inline void process_weight_update(
         uint32_t neuron_index = synapse_row_sparse_index(word,
         		synapse_index_mask);
 
-        log_debug("    Adding weight change %d to post-neuron %u",
+        log_info("    Adding weight change %d to post-neuron %u",
 						weight_change, neuron_index);
 
         // Get post event history of this neuron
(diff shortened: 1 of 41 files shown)


spynnaker/pyNN/extra_algorithms/delay_support_adder.py
```diff
@@ -28,14 +28,14 @@
...
-        Sequence[DelayExtensionVertex], Sequence[ApplicationEdge]]:
+        Sequence[DelayExtensionVertex], Sequence[Tuple[ApplicationEdge, str]]]:
     """
     Adds the delay extensions to the application graph, now that all the
     splitter objects have been set.
 
     :return: The delay vertices and delay edges that were added
     :rtype: tuple(list(DelayExtensionVertex), list(DelayedApplicationEdge or
-        DelayAfferentApplicationEdge))
+        DelayAfferentApplicationEdge, str))
```
Christian-B, line 38: this type is wrong. I suggest just removing it as then the annotation is sued.

spynnaker_integration_tests/test_various/test_send_multiple_partitions.py
```diff
@@ -0,0 +1,80 @@
...
+
+        sim.end()
+
+        print(spikes)
+        print(weights_1)
+        print(weights_2)
+        print(weights_3)
+
+        # There should be a spike to each neuron
+        assert len(spikes[0]) == 1
```
Christian-B, line 71: use self.assertEquals   which gives a better error
self because BaseTestCase extends unittest

spynnaker_integration_tests/test_various/test_send_multiple_partitions.py
```diff
@@ -0,0 +1,80 @@
...
+        assert len(spikes[1]) == 1
+        assert len(spikes[2]) == 1
+
+        assert numpy.array_equal(weights_1, ([0, 0, 5.0], ))
```
Christian-B, line 75: Use self.ListEqual

spynnaker_integration_tests/test_various/test_send_multiple_partitions.py
```diff
@@ -0,0 +1,80 @@
...
+    def send_spike(self, label, conn):
+        sleep(0.1)
+        conn.send_spike(label, 0)
+
+    # Added to check that the delay expander runs; a previous fix
+    # for a related issue inadvertently turned it off for this type of case
+    def do_run(self):
+        conn = sim.external_devices.SpynnakerLiveSpikesConnection(
+            send_labels=["Inject"], local_port=None)
+        conn.add_start_callback("Inject", self.send_spike)
```
Christian-B, line 34: he method 'add_start_callback(label, start_callback)' is in deprecation, and will be replaced with the method 'add_start_resume_callback(label, start_resume_callback)' in a future release.

neural_modelling/src/neuron/plasticity/weight_change/synapse_dynamics_external_weight_change.c
```diff
@@ -198,7 +200,7 @@ static inline void process_weight_update(
     const uint32_t *words = (uint32_t *) synapse_row_plastic_controls(fixed_region);
     uint32_t pre_spike = plastic_region_address->pre_spike;
 
-	log_debug("Weight change update for pre-neuron %u", pre_spike);
+	log_info("Weight change update for pre-neuron %u", pre_spike);
```
Christian-B, line 203: Do we need this (and others) to stay at info level

PR #1497, opened 2024-10-11
cap morphio

Christian-B:
work around due to https://github.com/BlueBrain/MorphIO/issues/509


rowleya:
May want to wait until later depending on what happens with this:
https://github.com/BlueBrain/MorphIO/issues/509

PR #1512, opened 2024-12-03
Reset cache

Christian-B:
Builds on https://github.com/SpiNNakerManchester/sPyNNaker/pull/1511
use reset_number instead of segment counter

Completely removes the segment_counter and its test

Move the database to segment/reset to the one in https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1248
But do store which files have the extended DDL run.


Christian-B:
better fix is https://github.com/SpiNNakerManchester/SpiNNFrontEndCommon/pull/1252

PR #1526, opened 2025-01-20
Actually detect window closure

rowleya:


Christian-B (approved):
I see no error here.

ok to merge

PR #1528, opened 2025-01-22
Abstract provides defaults

Christian-B:
In master there are two ways that the decorators @default_parameters and @default_initial_values are converted to a mapping of the values in the init.  Making these properties

@defaults
This calculated the mappings when the class is imported (even if not used)
The resulting mapping replace the calculate functions so the calculation is only done once!
This style also did not have properties for mypy to find 

AbstractPyNNModel
This calculated the mapping every time the property was called
Only worked for classes that extended AbstractPyNNModel

This PR merges the best of both solutions into AbstractProvidesDefaults
AbstractPyNNModel and any class which used to be decorated by @defaults now extends AbstractPyNNModel

The calculation of the mappings is delayed until either property is used.
Then BOTH properties are filled in (as the calculation uses a lot of sharable work)

Tested by:
https://github.com/SpiNNakerManchester/IntegrationTests/pull/307

question.
Which should we do:
1. Keep @defaults but deprecate warning as now
2. Keep defaults but only to raise an error how to change

### DIFF
--- a/spynnaker/pyNN/models/defaults.py
+++ b/spynnaker/pyNN/models/defaults.py
@@ -209,9 +209,9 @@ class AbstractProvidesDefaults(object):
         default_args = ([] if init_args.args is None else
                         init_args.args[n_args - n_defaults:])
         if init_args.defaults is None:
-           default_values = []
+            default_values = []
         else:
-           default_values = init_args.defaults
+            default_values = init_args.defaults
 
         # get the keys based on the decorators
         if hasattr(init, "_parameters"):
--- a/spynnaker/pyNN/models/defaults.py
+++ b/spynnaker/pyNN/models/defaults.py
@@ -189,6 +189,12 @@ def defaults(cls: type) -> type:
 
 
 class AbstractProvidesDefaults(object):
+    """
+    Provides the default_parameters and default_initial_values properties
+
+    These will be filled in based on the @default_parameters and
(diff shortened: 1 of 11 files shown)


spynnaker/pyNN/models/defaults.py
```diff
@@ -182,3 +186,103 @@ def defaults(cls: type) -> type:
...
+    @classmethod
+    def __fill_in_defaults(cls):
+        """
+        Fills in default_parameters and default_initial_values attributes
+        """
+        # get the init method
+        init = getattr(cls, "__init__")
+        # Find the real method as there may be decorators
+        while hasattr(init, "_method"):
+            init = getattr(init, "_method")
```
rowleya, line 208: Does this terminate?  How does the attribute get removed?

rowleya: I assume it does terminate as it is the same code as before... sometimes my own code confuses me!

rowleya: Oh I see now, it is essentially recusing the init until it finds the real init

rowleya (approved):
Looks good to me if it works!

rowleya:
> Which should we do:
> 1. Keep @defaults but deprecate warning as now
> 2. Keep defaults but only to raise an error how to change

I would probably say deprecate as otherwise people might have things broken at the last minute by accidentally updating...

PR #1530, opened 2025-01-23
Coloured application vertex

Christian-B:
PopulationApplicationVertex and  have a common method DelayExtensionVertex have common method n_colour_bits

The local classes mix the two when caching pre vertex

This creates a class between these two classes and ApplicationVertex which adds the method

A bit of extra typing was added to make sure it all worked..


Christian-B:
The alternative of just using Union[PopulationApplicationVertex, DelayExtensionVertex] got messy as  importing AbstractPopulationVertex outside of Type_CHECKING was not possible in spynnaker/pyNN/models/common/local_only_2d_common.py
circular import monster!

rowleya (approved):

PR #1546, opened 2025-03-05
typing of tests

Christian-B:
requires https://github.com/SpiNNakerManchester/PACMAN/pull/600

ConnectionHolder types updated to included list types.

---

"Values" TypeAlias used for setting PyNn values such as State Values and parameter Values updated to include a callable to go from int to a float.  

---

The input types allowed for Weight and delays
spynnaker/pyNN/types.py
changed to allow NDArray of float.
removed "int" as float implies int

---

added a writer _get_id_counter test method
as hacking the data and adding an ignore in 80 characters was too tricky

---

Population / PopulationView all method type fixed.

---

weight_scales can be numpy array or Sequence of float

---

Type alias for mask used to create PopulationViews relaxed  to accept any Sequence (for example a tuple)

---

spynnaker_integration_tests/test_additional_params_on_pop_not_recogonised/test_additional_params_mix.py
Changed to include the checking of logs

---

/home/brenninc/spinnaker/sPyNNaker/unittests/test_from_file_connector.py
replaced:
       elif len(header):
            f.write("# {}\n".format(header))
with:
        else:
            assert len(header) == 0
reason is the write of a string fails! but was never actually called!

---

numpy.array_equal can compare None to None even though mypy thinks it can not


rowleya (approved):
Haven't looked in detail at the changes to the tests themselves, but the changes to the non-test code look fine

PR #1558, opened 2025-05-02
Split cores

rowleya:
This integrates the execution of neuron code and synapse code on separate cores more completely, making it nicer for the user to use.  This includes:
 - The user can choose to set a number of synapse cores for a model (like the number of neurons per core)
 - The user decide whether they want to allow delay extensions or not for a model (as above)
 - The user can control the number of neurons per core, synapse cores and allowing of delay extensions when creating populations.
 - The user can use 0.1ms timesteps in realtime without having to add an option to the configs.
 - The code will try to work out whether to use synapse cores and how many synapse cores to use if so, and whether to allow delay extensions.

What it doesn't do:
 - Theoretically, the use of synapse cores shouldn't need to depend on the timestep.  In this code, it won't automatically use synapse cores unless the timestep is less than 1.0.  See #1559.

Requires:
 - SpiNNakerManchester/PACMAN#599
 - SpiNNakerManchester/SpiNNFrontEndCommon#1284
 - SpiNNakerManchester/PyNNExamples#127
 - SpiNNakerManchester/microcircuit_model#44
 - SpiNNakerManchester/TSPOnSpiNNaker#61

Tested by:
 - SpiNNakerManchester/IntegrationTests#317


spynnaker/pyNN/data/spynnaker_data_view.py
```diff
@@ -263,6 +269,62 @@ def set_number_of_neurons_per_dimension_per_core(
...
+        from spynnaker.pyNN.models.neuron import AbstractPyNNNeuronModel
+        if not issubclass(neuron_type, AbstractPyNNNeuronModel):
+            raise TypeError(f"{neuron_type} is not an AbstractPyNNNeuronModel")
+
+        if SpynnakerDataView.get_n_populations() > 0:
+            warn_once(logger,
+                      "set_allow_delay_extensions "
+                      "only affects Populations not yet made. "
+                      "Either move the set to the beginning of the script "
+                      "or consider using Population.set_n_synapse_cores")
```
Christian-B, line 324: why Population.set_n_synapse_cores?

spynnaker/pyNN/models/neuron/abstract_pynn_neuron_model.py
```diff
@@ -40,9 +42,51 @@ class AbstractPyNNNeuronModel(AbstractPyNNModel):
     """
     __slots__ = ("__model", )
 
+    # The number of synapse cores for PyNN models that use PopulationVertex
```
Christian-B, line 45: Why here and in  spynnaker_data_view.py

Christian-B (changes requested):
remove one _n_synapse_cores and one _allow_delay_extensions

spynnaker/pyNN/data/spynnaker_data_view.py
rowleya: No longer used; this is now an error

spynnaker/pyNN/models/neuron/population_vertex.py
```diff
@@ -474,35 +498,195 @@ def set_structure(self, structure: BaseStructure) -> None:
...
         split synapse-neuron core model.
 
-        .. note::
-            This is currently hard-coded but could be updated to work
-            this out based on the number of incoming synapses.
-
         :rtype: int
         """
-        return 1
+        self.__update_n_synapse_cores()
```
Christian-B, line 540: if you are always going to call self.__update...

Why not merge the two methods and just have it return the value.

for example:
 def n_synapse_cores_required(self) -> int:
   if self.__n_synapse_cores is not None:
       return self.__n_synapse_cores
  ....
  return self.__n_synapse_cores

spynnaker/pyNN/models/neuron/population_vertex.py
```diff
@@ -474,35 +498,195 @@ def set_structure(self, structure: BaseStructure) -> None:
...
+            n_synapse_cores = math.ceil(
+                (synapses_per_second + poisson_synapses_per_second) /
+                synapses_per_core_per_sim_second)
+
+            # If the number of cores needed is more than the maximum, use the
+            # maximum
+            self.__n_synapse_cores = min(max_n_cores, n_synapse_cores)
+
+    @property
+    def max_delay_steps(self) -> int:
```
Christian-B, line 617: also merge

spynnaker/pyNN/models/neuron/population_vertex.py
```diff
@@ -474,35 +498,195 @@ def set_structure(self, structure: BaseStructure) -> None:
...
+        return self.__max_delay_slots_available
+
+    @property
+    def max_delay_steps_incoming(self) -> int:
+        """
+        The maximum delay steps needed to handle incoming synapses,
+        accounting for delay extensions.
+        """
+        self.__update_max_delay()
+        assert self.__max_delay_ms is not None
```
Christian-B, line 632: max_delay = self.def max_delay_steps

It that method (merged or not) do the update and the assert it is an int

spynnaker/pyNN/models/neuron/population_vertex.py
```diff
@@ -474,35 +498,195 @@ def set_structure(self, structure: BaseStructure) -> None:
...
+
+    @property
+    def allow_delay_extension(self) -> bool:
+        """
+        Whether delay extension should be allowed or not.
+        """
+        if not self.__allow_delay_extensions:
+            return False
+        # Determine if we *expect* a delay extension; if not disallow
+        self.__update_max_delay()
```
Christian-B, line 647: use self.def max_delay_steps

spynnaker/pyNN/models/neuron/population_vertex.py
```diff
@@ -543,13 +727,21 @@ def add_incoming_projection(self, projection: Projection) -> None:
...
         # pylint: disable=protected-access
         pre_vertex = projection._projection_edge.pre_vertex
         self.__incoming_projections[pre_vertex].append(projection)
         if pre_vertex == self:
             self.__self_projection = projection
         if isinstance(pre_vertex, SpikeSourcePoissonVertex):
-            self.__incoming_poisson_projections.append(projection)
+            # Only keep these when there is only one projection to this
+            # vertex, as otherwise not interesting
+            if pre_vertex not in self.__incoming_poisson_projections:
```
Christian-B, line 741: is looks dangerous!

the first time add the Projection
the second time del the Projection
the third time will add it again!

spynnaker/pyNN/models/neuron/synapse_dynamics/abstract_synapse_dynamics.py
```diff
@@ -398,3 +409,11 @@ def validate_connection(
         # By default, just ask the connector
         synapse_info.connector.validate_connection(
             application_edge, synapse_info)
+
+    @property
+    @abstractmethod
+    def synapses_per_second(self) -> int:
+        """
+        Approximate number of synapses that can be processed per second
```
Christian-B, line 417: Would be good to add a comment about which it does under estimate, over estimate or as close as possible.

spynnaker/pyNN/models/neuron/synapse_dynamics/abstract_synapse_dynamics.py
```diff
@@ -223,6 +223,17 @@ def is_combined_core_capable(self) -> bool:
         """
         raise NotImplementedError
 
+    @property
+    @abstractmethod
+    def is_split_core_capable(self) -> bool:
```
Christian-B, line 228: first thought is to combine this with is_combined_core_capable as otherwise both could return False

second though is if a model defines both as False the developers should be laughed at

spynnaker/pyNN/models/spike_source/spike_source_poisson_machine_vertex.py
```diff
@@ -246,20 +246,18 @@ def __init__(
...
-        self.__sdram_partition: Optional[AbstractSDRAMPartition] = None
+        self.__sdram_partition: Optional[
+            SourceSegmentedSDRAMMachinePartition] = None
         self.__rate_changed = True
 
     @property
     def _pop_vertex(self) -> SpikeSourcePoissonVertex:
         return cast('SpikeSourcePoissonVertex', self.app_vertex)
 
+    @overrides(SendsSynapticInputsOverSDRAM.set_sdram_partition)
```
Christian-B, line 257: docs in abstract says it should never be called twice. This allows that

Christian-B (changes requested):

spynnaker/pyNN/models/abstract_models/sends_synaptic_inputs_over_sdram.py
```diff
@@ -12,10 +12,28 @@
...
-class SendsSynapticInputsOverSDRAM(AbstractSupportsSDRAMEdges):
+class SendsSynapticInputsOverSDRAM(
+        AbstractSupportsSDRAMEdges, metaclass=AbstractBase):
     """
-    A marker interface for an object that sends synaptic inputs over SDRAM.
+    An object that sends synaptic inputs over SDRAM.
     """
+
+    @abstractmethod
+    def set_sdram_partition(
```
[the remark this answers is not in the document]
rowleya: This was more for the sake of typing; happy to leave for a later fix

spynnaker/pyNN/extra_algorithms/splitter_components/spynnaker_splitter_selector.py
```diff
@@ -56,8 +56,22 @@ def spynnaker_splitter_selector() -> None:
         string_describing_what_being_progressed=PROGRESS_BAR_NAME,
         total_number_of_things_to_do=SpynnakerDataView.get_n_vertices())
 
-    for app_vertex in progress_bar.over(SpynnakerDataView.iterate_vertices()):
+    remaining: List[ApplicationVertex] = []
+    for app_vertex in SpynnakerDataView.iterate_vertices():
```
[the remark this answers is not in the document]
rowleya: To determine if split cores are required, we potentially need to know how many synapse cores are required (certainly if the time step is less than 1ms in this PR, and in the future this might be more important if #1559 is implemented).  To know how many synapse cores we need, we need to know if the Poisson sources are direct, and to know that, we need to know what splitter they have.

spynnaker/pyNN/models/neuron/population_vertex.py
rowleya: Could do for one variable, but need both; each of these methods are subtly different e.g. max_delay_steps is how many steps are available for use given other things, max_delay_steps_incoming are how many we actually need for the incoming synapses, with delay extensions accounted for, and allow_delay_extensions is checking if the maximum delay of the incoming synapses without extensions is more than the maximum number available for use.

spynnaker/pyNN/models/neuron/population_vertex.py
```diff
@@ -474,35 +498,195 @@ def set_structure(self, structure: BaseStructure) -> None:
...
+                if rate > 0:
+                    spikes_per_second = rate
+
+            pre_synapses_per_second: int = 0
+            for proj in projs:
+                # pylint: disable=protected-access
+                s_info = proj._synapse_information
+                dynamics = s_info.synapse_dynamics
+                conn = s_info.connector
+                n_conns: Optional[int] = None
```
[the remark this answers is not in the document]
rowleya: The first call below can return None as it doesn't know, then you need it to be allowed to be None until after the second call, at which point it isn't None!  Alternative would be to use separate variables, but might be a lot of effort...

spynnaker/pyNN/spinnaker.py
```diff
@@ -126,6 +128,12 @@ def __init__(
             db.insert_version("neo_version", neo_version)
             db.insert_version("lazyarray_version", lazyarray_version)
 
+        # Clears all previously added ceiling on the number of neurons per core
+        AbstractPyNNModel.reset_all()
```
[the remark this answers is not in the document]
rowleya: Fixed now

spynnaker/pyNN/models/neuron/population_vertex.py
```diff
@@ -136,6 +136,10 @@
                            NeuronRecorder.PACKETS: "",
                            NeuronRecorder.REWIRING: ""}
 
+# The number of non-synapse cores that are likely needed (neuron core +
+# extra monitor in general)
+_N_OTHER_CORES = 2
```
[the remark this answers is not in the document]
rowleya: Fixed (must have missed that this was a thing)

spynnaker/pyNN/models/neuron/population_vertex.py
```diff
@@ -253,6 +264,12 @@ def __init__(
         :param str label: The label on the population
         :param int max_atoms_per_core:
             The maximum number of atoms (neurons) per SpiNNaker core.
+        :param n_synapse_cores:
+            The number of synapse cores to use: 0 for combined core,
+            or None for automatic determination
+        :type n_synapse_cores: int or None
```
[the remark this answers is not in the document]
rowleya: Fixed - removed all types (was matching the existing style but better to fix)

spynnaker/pyNN/models/neuron/population_vertex.py
```diff
@@ -474,35 +498,195 @@ def set_structure(self, structure: BaseStructure) -> None:
...
+        # If we can't use a combined core, use a split core
+        if not self.__synapse_dynamics.is_combined_core_capable:
+            return False
+
+        # If we can't use a split core, use a combined core
+        if not self.__synapse_dynamics.is_split_core_capable:
+            return True
+
+        # If the user has chosen to have a synapse core, add one
+        if self.__n_synapse_cores is not None and self.__n_synapse_cores > 0:
```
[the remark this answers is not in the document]
rowleya: Added some error checking for both cases

spynnaker/pyNN/models/neuron/population_vertex.py
rowleya: Updated

spynnaker/pyNN/models/neuron/population_vertex.py
rowleya: Updated I think now

spynnaker/pyNN/models/neuron/population_vertex.py
rowleya: I have used the return types from __update_max_delay now

spynnaker/pyNN/models/neuron/population_vertex.py
rowleya: Now updated to use the return values

spynnaker/pyNN/models/neuron/synapse_dynamics/abstract_synapse_dynamics.py
rowleya: This is a tricky one, as both can be True or one can be True, but both can't be False.  So the two are independent but also you need one to be True as otherwise nothing happens.  I have added a check when splitting happens so hopefully that will avoid the situation

spynnaker/pyNN/models/spike_source/spike_source_poisson_machine_vertex.py
rowleya: Fixed

Christian-B (approved):
can be merged when green

spynnaker/pyNN/models/neuron/synapse_dynamics/abstract_synapse_dynamics.py
Christian-B: Should that not be the other way around.  Underestimating will create a system that can not handle the received spikes

rowleya: No, if you underestimate how many spikes you can process, you can process more than you estimate - that is, your estimate is under what you can really handle.  If you overestimate you are guessing you can do more than you really can, so more spikes will be sent to you than you can handle.

PR #1560, opened 2025-05-06
no need to cache which types have set neurons per core

Christian-B:


Christian-B:
now part of https://github.com/SpiNNakerManchester/sPyNNaker/pull/1558

PR #1570, opened 2025-05-30
remove unneeded set to None

Christian-B:
removes an unneeded line which confuses mypy

found by new mypy 1.16


Christian-B:
ran tests on my machine.

self approving as blocker.

PR #1575, opened 2025-06-09
fixes as dtype param can not be integer or floating

Christian-B:
requires https://github.com/SpiNNakerManchester/SpiNNUtils/pull/312

this covers the part where I think we actually need to change code even if mypy catches up


Christian-B:
https://github.com/SpiNNakerManchester/sPyNNaker/actions/runs/15817402994 shows this is needed.

rowleya (approved):

PR #1582, opened 2025-06-30
Less doc types

Christian-B:
depends on https://github.com/SpiNNakerManchester/TestBase/pull/63


rowleya (approved):
Not looked at all files, but assuming it is all type removal, and is checked by tests anyway, all good!

PR #1646, opened 2026-02-12
Block connector reuse

### LINKED ISSUE #1406: Connectors should be reusable
The set_projection_information should be removed to allow connectors to be reused.  This requires a bit of thought where a calculation has been done and needs to be cached somewhere.  The obvious place for cached data is the projection, but this would have to be connector-specific information so some thought is required!

Christian-B:
Fixes https://github.com/SpiNNakerManchester/sPyNNaker/issues/1406 or better said marks that as will not do!

Reusing Connectors would take a lot of changes with risk of adding extra confusion and bugs.

Instead the solution here is to track if a connector has been used and then to raise an exception on reuse.

True many connectors could be safely reused but the clearest message to uses is user is 

We do not support Connector reuse!


rowleya (approved):

Christian-B:
no longer needed as alternative done

PR #1651, opened 2026-02-18
connector Parameters and clone on reuse.

### LINKED ISSUE #1406: Connectors should be reusable

Christian-B:
Alternative to https://github.com/SpiNNakerManchester/sPyNNaker/pull/1646

Builds on https://github.com/SpiNNakerManchester/sPyNNaker/pull/1649

Fixes: https://github.com/SpiNNakerManchester/sPyNNaker/issues/1406

Implements the methods described at https://neuralensemble.org/docs/PyNN/reference/connectors.html

removes Setters on connections.

--
Tracks when a conenctor is used

On reuse a Conectors creates a clone of itself.

It does so using the get_parameters() method
-These return a dict of params which will can replicate the original Connector as it was when created

Note: get_parameters is not Abstract to not break user created Connectors

Connectors that do not implement get_parameters will not be reusable


rowleya (approved):
I note that this means that people can no longer set the parameters of a connector; I guess this is by design since a copy is made that is somewhat hidden to the user.  I am unsure if this will affect anyone but if not, this is fine!

PR #1653, opened 2026-02-20
Population parameters

Christian-B:
Specifically adds a few create model parameters that can be added to Populations.

The ones I found 
- max_rate
- n_colour_bits
- n_synapse_cores
- neurons_per_core
- port
- reserve_reverse_ip_tag
- seed
- splitter
- virtual_key:

Other ones would continue to work as additional_parameters or kwargs

----
Docs for params based on what was found in the model or vertex.

---
Ugliness:

mypy got confused on extended classes which required converting kwargs to additional_parameters

### DIFF
--- a/spynnaker/pyNN/models/populations/population.py
+++ b/spynnaker/pyNN/models/populations/population.py
@@ -108,7 +108,8 @@ class Population(PopulationBase):
         :param label: A label for the population
         :param max_rate: Typed semantic sugar for an additional_parameter
         :param n_colour_bits: Typed semantic sugar for an additional_parameter
-        :param n_synapse_cores: Typed semantic sugar for an additional_parameter
+        :param n_synapse_cores:
+            Typed semantic sugar for an additional_parameter
         :param neurons_per_core:
             Typed semantic sugar for an additional_parameter
         :param port: Typed semantic sugar for an additional_parameter
--- a/spynnaker/pyNN/models/populations/population.py
+++ b/spynnaker/pyNN/models/populations/population.py
@@ -86,7 +86,6 @@ class Population(PopulationBase):
             structure: Optional[BaseStructure] = None,
             initial_values: Optional[Dict[str, float]] = None,
             label: Optional[str] = None,
-            *,
             max_rate: Optional[float] = None,
             n_colour_bits: Optional[int] = None,
             neurons_per_core: Optional[Union[int, Tuple[int, ...]]] = None,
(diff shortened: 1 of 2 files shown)


rowleya:
I note that some (possibly many) of these additional parameters are not general to all populations, that is, they depend on the model being run.  We allow the passing of these parameters at the Population level because they will apply to the vertex that the model gets converted to, rather than to each atom of the Population, which the model itself represents.  This is a tricky distinction!  I wonder if it would be better to document these somehow in the model, but as I say, these are tricky things to represent...

spynnaker/pyNN/models/populations/population.py
```diff
@@ -95,17 +106,72 @@ def __init__(
...
+            Model.create_vertex parameter.
+            Likely: The Population seed,
+            used to ensure the same random generation on each run.
+        :param splitter:
+            Model.create_vertex parameter.
+            Likely: splitter from application vertices to machine vertices
+        :param virtual_key:
+            Model.create_vertex parameter.
+            Likely:
+            The virtual_key of the population to identify the population.
```
rowleya, line 140: I think this unfortunately adds to the confusion; adding these parameters implies all models accept them but in reality they don't.  I would prefer that this is left to the model itself to document therefore.

A note here is that PyNN itself doesn't actually accept additional parameters, so what we do with this is our own; we could, in fact, force that all parameters are something that the vertex accepts.

Additional note is that these are not the same as the cell_params (which are, in fact, deprecated).  The cell_params are arguments passed when creating the model, the additional_args are arguments passed when creating the vertex, and model is also one of those arguments.  In our case, we accept a vertex in place of a model, so this could be resolved that way instead, but it would then start also to look odd when working with standard models, as you would then have something like p.Population(n_neurons, p.PopulationVertex(..., model=p.IF_curr_exp(...))).

spynnaker/pyNN/models/populations/population.py
```diff
@@ -95,17 +106,72 @@ def __init__(
...
-            Additional parameters to pass to the vertex creation function.
+            Alternative way of entering Model.create_vertex parameter.
+            Ignored if also provided as a named parameter.
         :param additional_kwargs:
-            A nicer way of allowing additional things
+            Additional Model.create_vertex parameters.
+            See the Model's create_vertex method for more details.
+            create_vertex parameters will be ignored if
+            the Model does not accept this parameter,
+            or raise an Exception if a Vertex is passed in
```
rowleya, line 149: I actually think that we should raise an exception in both cases, especially if this is the only use of these additional args.

PR #1661, opened 2026-04-07
cache the index used to make call repeatable

Christian-B:
SplitterPopulationVertexNeuronsSynapses.get_source_specific_in_coming_vertices

Assumes the method is only called once per source_vertex, partition_id pair.

In practice this assumption holds.

BUT if something like 
pacman/operations/multi_cast_router_check_functionality/valid_routes_checker.py

may call it a second time with the same pair.

This pr caches the minimum data needed to make the call repeatable.

Faster alternative are available caching more data but as in normal operation the cache is not needed the choice was for size vs speed.


Christian-B:
Found by and tested by https://github.com/SpiNNakerManchester/PACMAN/pull/627 but independant

rowleya (approved):

PR #1677, opened 2026-07-06
Weight Scale Responsibility

rowleya:
Move responsibility for weight scaling to the synapse dynamics object.  Although this has little effect on the sPyNNaker code itself, this has been done in recognition that NestML wants to represent weights differently in the synapses and the ring buffers,  This is a valid option - the scaling can be done at the time of addition of weights to the ring buffers to allow weights to be represented differently in the synapses.  

Note that this also helps with STDP rules where the weight representation might have more fractional bits to allow small updates to occur, even if those updates can't immediately affect the inputs to the ring buffers.  Multiple updates however might do so!


Christian-B (approved):
assuming https://github.com/SpiNNakerManchester/IntegrationTests run and flag no related issues.

PR #1678, opened 2026-07-07
Clean up comments related to traces

clinssen:
Tiny cleanup of the naming and comments related to synaptic traces. (I believe "o1" and "o2" are related to triplet STDP rules, that's probably where this came from.)


rowleya:
Please add "pull_request" to the list here:
https://github.com/clinssen/sPyNNaker/blob/trace_comments_cleanup/.github/workflows/python_actions.yml#L23
and here:
https://github.com/clinssen/sPyNNaker/blob/trace_comments_cleanup/.github/workflows/c_actions.yml#L19

This will then make the tests run on our side and let me merge things directly.

PR #1679, opened 2026-07-07
More flexible STDP make

rowleya:
Refactor make process a little to allow other STDP systems which don't need the weight and timing rules to be separately implemented.  This is particularly relevant to NESTML where the STDP code is generated in a single file.


Christian-B (commented):
Looks ok assuming it works.

Needs a full https://github.com/SpiNNakerManchester/IntegrationTests run.

approved if that is green.

rowleya:
Not green, but the failures were unrelated

Christian-B (approved):

PR #1695, opened 2026-08-14
TEST PR for read the docs

Christian-B:


Christian-B:
Testing show that it was the line 
IoDest: TypeAlias = str | neo.baseio.BaseIO | None
Which broke read the docs.
Error: 
TypeError: unsupported operand type(s) for |: 'type' and 'BaseIO'-

Just adding 
from __future__ import annotations
was not enough

What was needed is 
- move declaration under if TYPE_CHECKING:
- move any import that uses it under if TYPE_CHECKING:
- Add from __future__ import annotations where if TYPE_CHECKING: used

Christian-B:
Testing complete will do fix in https://github.com/SpiNNakerManchester/sPyNNaker/pull/1694

PR #1722, opened 2026-09-11
replace getattr with a direct call

Christian-B:
see https://docs.astral.sh/ruff/rules/get-attr-with-constant/

part of https://github.com/SpiNNakerManchester/SupportScripts/issues/135


Christian-B:
After looking at pylint results and discussing on teams choice is to keep as is because this is pythonic decorators.

"this is helping PyNN users do things in a PyNN way which is already enforcing a lot of pythonicism!"

PR #1723, opened 2026-09-11
NOQA B009

Christian-B:
part of https://github.com/SpiNNakerManchester/SupportScripts/issues/135

replaces https://github.com/SpiNNakerManchester/sPyNNaker/pull/1722

NOQA as

"this is helping PyNN users do things in a PyNN way which is already enforcing a lot of pythonicism!"


Christian-B:
includes changes it should not

